PR #24, opened 2025-06-13
feat: use string union instead of enums

swapydapy:
Typescript reviewers suggested: TypeScript enums are an anti-pattern -- prefer string literal unions (e.g. "foo" | "bar"). all the same autocomplete niceties but doesn't create non-erasable syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ class MyAgentExecutor implements AgentExecutor {
         taskId: taskId,
         contextId: contextId,
         status: {
-          state: "canceled",
+          state: 'canceled',
           timestamp: new Date().toISOString(),
         },
         final: true,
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -160,7 +160,7 @@ class MyAgentExecutor implements AgentExecutor {
         taskId: taskId,
         contextId: contextId,
         status: {
-          state: TaskState.Canceled,
+          state: "canceled",
```
mbleigh: Should be consistent in single or double quotes (above are single).

src/samples/agents/movie-agent/index.ts
```diff
@@ -162,7 +162,7 @@ class MovieAgentExecutor implements AgentExecutor {
           taskId: taskId,
           contextId: contextId,
           status: {
-            state: TaskState.Canceled,
+            state: "canceled",
```
mbleigh: Quote consistency again.

PR #31, opened 2025-06-18
feat: Added the Eliza chatbot to demo support for user/business/gov scoped agent identities and universal authentication (DIDs, JWT, and the Agentic Profile)

mprynce:
This PR adds Decentralized ID (DID)/JWT based authentication and user, business, and gov scoped identity to A2A agents.  While DIDs can support DLT, this implementation is web based and all DID documents are accessible via HTTP(S).

# What is this solving?

While most agent orchestration frameworks authenticate with remote agents using API keys as bearer tokens or user driven OAuth, there is a big gap in the ability of two different entities agents being able to authenticate.  For example, if you and I are attending the same event and we should really meet to discuss the **next big thing**, then how does an agent representing you find an agent representing me, and then how to they authenticate each other, and start a dialog to determine your and my synergies?

Having a DID associated with you, and a DID for me is the first step.  A presence service can then see we are nearby.  We can resolve DID documents from our DIDs which in turn list the agents representing us.  Public key cryptography obviates authentication servers as the DID document contains the public keys.  Standard HTTP 401/WWW-Authenticate header with challenge provides a very well established and secure way for agents to establish secure communications.

# Design

The Agentic Profile is a very thin wrapper around DIDs and A2A that supports the above interaction.  I've modified the A2A server to accept an AgentSessionResolver for use when agents require authentication.  A working example is in the eliza-app.ts file.

I also modified the A2AClient to accept an authHandler option which intercepts HTTP 401 responses and can inject the WWW-Authenticate header.  I added support for two more parameters to the CLI to indicate which authentication profile to use, and which agent from that profile to use to generate the JWT.

Finally, I added a utility script at scripts/create-global-agentic-profile to publish a demo agentic profile to the public web for testing.  A local copy of the private keys for that profile are stored at ~/.agentic/iam

Much more information is available on my [Agentic Profile blog](https://agenticprofile.ai) and also in the NPM agentic profile packages:

- [@agentic-profile/auth](https://www.npmjs.com/package/@agentic-profile/auth)
- [@agentic-profile/a2a](https://www.npmjs.com/package/@agentic-profile/a2a)

Thank you!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "@a2a-js/sdk",
       "version": "0.2.1",
       "dependencies": {
+        "@agentic-profile/auth": "^0.6.0",
         "@types/cors": "^2.8.17",
         "@types/express": "^5.0.1",
         "body-parser": "^2.2.0",
@@ -16,6 +17,7 @@
         "uuid": "^11.1.0"
       },
       "devDependencies": {
+        "@agentic-profile/eliza": "^0.1.0",
         "@genkit-ai/googleai": "^1.8.0",
         "@genkit-ai/vertexai": "^1.8.0",
         "@types/chai": "^5.2.2",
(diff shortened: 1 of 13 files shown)


holtskinner:
Can you move this sample agent to https://github.com/google-a2a/a2a-samples ?

mprynce:
@holtskinner - happy to move the Eliza sample out.  It was really just to give a working demo of the DID/Agentic Profile auth code.  Can I do another PR with just the auth framework?

PR #33, opened 2025-06-19
feat: Added generic client hooks for HTTP based authentication, and improved agent.json resolution

mprynce:
This PR adds a generic client hook for HTTP based authentication.  This allows protocols, like the [Agentic Profile](https://agenticprofile.ai), to implement Decentralized ID (DID)/JWT based authentication which then enables user, business, and gov scoped identity for A2A agents.  When an authHandler is provided and an HTTP 401 is received, the client will try a second time (see the A2AClient._fetch() function) with revised auth headers if new ones are made available by the authHandler.

This PR also improves the client agent card resolution.  Previously the card was always found at <agentpath>/.well-known/agent.json which is only correct for agents at the URL path root.  For agents with paths, such as https://example.com/agents/coder the URL for the agent card should be https://example.com/agents/coder/agent.json. This PR fixes the error.

I've moved the previous Eliza sample agent to a separate PR into a2a-samples.  The Eliza sample provides A2A server side authentication, and a modified A2A CLI with authentication.

The Eliza sample code relies on this PR to function.


src/client/client.ts
```diff
@@ -51,9 +57,10 @@ export class A2AClient {
    * The `url` field from the Agent Card will be used as the RPC service endpoint.
    * @param agentBaseUrl The base URL of the A2A agent (e.g., https://agent.example.com).
    */
-  constructor(agentBaseUrl: string) {
+  constructor(agentBaseUrl: string, options?: A2AClientOptions) {
```
swapydapy, line 61: I suggest, in the options, we allow users to pass on their implementation of "fetch". This would give them maximum flexibility on how to handle API auth failures. Something like:

 customFetch = (input: RequestInfo | URL, init?: RequestInit): Promise<Response>

swapydapy (changes requested):

swapydapy:
Thank you for the PR. Suggested an alternative approach of allowing custom `fetch` to passed on. Please rebase with latest changes.

mprynce:
> Thank you for the PR. Suggested an alternative approach of allowing custom `fetch` to passed on. Please rebase with latest changes.

Good idea and happy to implement.  That was actually in the original implementation (before me) and wasn't added for the current one.  I'll get it back in :)

src/client/client.ts
mprynce: I used the pattern from the older code of fetchImpl and passed as an option to the constructor.  Am writing unit tests next...

mprynce:
Added  client unit tests and I think it just needs approval from a maintainer... Let me know of any other changes that are needed.  Thanks!

src/index.ts
```diff
@@ -3,5 +3,27 @@
  * Exports the common types.
  */
 
+export type { AgentExecutor } from "./server/agent_execution/agent_executor.js";
```
swapydapy: Instead of the main index.ts, could you export these server types in /server/index.ts, and the client one in /client/index.ts

We have 3 entrypoints now. One each for client & server, one for common API types

test/server/default_request_handler.spec.ts
```diff
@@ -282,7 +284,9 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
 
     it('sendMessage: should handle agent execution failure for non-blocking calls', async () => {
         const errorMessage = 'Agent failed!';
-        (mockAgentExecutor as MockAgentExecutor).execute.rejects(new Error(errorMessage));
+        (mockAgentExecutor as MockAgentExecutor).execute.callsFake(async () => {
```
swapydapy: is it any different from previous invocation of directly using `.execute.rejects(new Error(errorMessage));`

src/client/auth-handler.ts
```diff
@@ -0,0 +1,31 @@
...
+     * the revised headers.
+     * @param req The RequestInit object used to invoke fetch()
+     * @param res The fetch Response object
+     * @returns If the HTTP request should be retried then returns the HTTP headers to use,
+     * 	or returns undefined if no retry should be made.
+     */
+    shouldRetryWithHeaders: (req:RequestInit, res:Response) => Promise<HttpHeaders | undefined>;
+
+    /* If the last call using the headers was successful, report back using this function. */
+    onSuccess: (headers:HttpHeaders) => Promise<void>
```
swapydapy: We can make this optional. I assume for majority cases, the implementation of `shouldRetryWithHeaders` would also save the new headers.

swapydapy (changes requested):

src/client/auth-handler.ts
```diff
@@ -0,0 +1,31 @@
...
+}
```
swapydapy: Add a newline

swapydapy:
Thanks for the changes. Added some minor comments. Once addressed, we can merge this PR.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+          headers: { 'Content-Type': 'application/json' }
+        });
+      }
+      
+      // Simulate RPC endpoint calls
+      if (url.includes('/api')) {
+        const authHeader = options?.headers?.['Authorization'] as string;
+        
+        // First call: no auth header, return 401
+        if (fetchCallCount === 2 && !authHeader) {
```
swapydapy: Can we simplify this fake fetch setup. I thin it can simply check for auth header, if not return 401, otherwise a response.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+          error: {
+            code: -32001,
+            message: 'Authentication required'
+          },
+          id: fetchCallCount
+        }), {
+          status: 401,
+          headers: { 
+            'Content-Type': 'application/json',
+            'WWW-Authenticate': 'Agentic eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c'
```
swapydapy: Can we simply this WW-Authenticate header. Just keep it 'Agentic: <predefined-10-char value>'

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+            text: 'Second message'
+          } as TextPart]
+        }
+      };
+
+      // First request - should trigger auth flow
+      await client.sendMessage(messageParams);
+      
+      // Reset call count for second request
+      fetchCallCount = 0;
```
swapydapy: As stated, don;t think we need to track this fetchCallCount

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+      mockFetch.reset();
+      
+      // Create a new mock for the second request that expects auth header
+      mockFetch.callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: why is this new fake needed

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+          parts: [{
+            kind: 'text',
+            text: 'Concurrent message'
+          } as TextPart]
+        }
+      };
+
+      // Create a new mock that handles concurrent requests properly
+      mockFetch.reset();
+      mockFetch.callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: This new fake is not needed.

swapydapy (changes requested):
Please go through the tests once. There seem to be lot of redundant code, especially regarding the fake Fetch.

src/index.ts
mprynce: Refactored client and server exports to the appropriate client/index.ts and server/index.ts files

test/server/default_request_handler.spec.ts
mprynce: Previously it was failing in my local tests, but with the newest codebase the original version is working so I've reverted it

src/client/auth-handler.ts
mprynce: I've updated auth-handler.ts with a more complete description of the logic.  In short, shouldRetryWithHeaders usually provides an HTTP Authorization header that might or might not work.  This new header is not saved until the onSuccess() callback.  If the invalid header was saved it could lead to additional authorization requests with a bad header.  This logic is currently in production on Matchwise and working well.

src/client/auth-handler.ts
mprynce: Fixed :)

test/client/client_auth.spec.ts
mprynce: The first request is (intentionally) missing the authorization header, which causes the server to issues WWW-Authenticate with a challenge (in this case, which must be signed by the client).  This is the authentication flow used by the agentic profile.

test/client/client_auth.spec.ts
mprynce: This is used to sequence the first response missing the auth, which provides the challenge, and the second response which makes sure the auth header is provided.  Otherwise a test that immediately passes the auth header would pass which is not what we are testing here.

test/client/client_auth.spec.ts
mprynce: (Not me, this was generated, but it seems to make sense... :)

- Original mock: Simulates the full auth flow (no token → 401 → get token → retry with token)
- New mock: Simulates a scenario where the client already has a token and should use it immediately

The new fake ensures that the test properly validates that the client is reusing the cached authentication token instead of going through the full authentication flow again.

The new fake is needed because the test is transitioning from testing the authentication flow to testing token reuse. Here's what's happening:

## The Original Mock (lines 67-180)
The original mockFetch was designed to simulate the complete authentication flow:
1. First call: Agent card fetch (always succeeds)
2. Second call: RPC request without auth header → returns 401 to trigger auth
3. Third call: RPC request with auth header → returns success
The original mock uses fetchCallCount to track which call it is and behaves differently based on that count.

## The Problem with Reusing the Original Mock
When the second test (should reuse authentication token for subsequent requests) runs:
1. The first request completes the auth flow and stores the token
2. The test calls mockFetch.reset() to clear the call count
3. But the original mock logic still depends on fetchCallCount - it expects the second API call to be call #2 (without auth) and the third to be call #3 (with auth)

## Why the New Fake is Required
The new fake on line 270 creates a different behavior that's appropriate for the second test:
1. It doesn't use fetchCallCount - it doesn't need to track call order
2. It expects the auth header immediately - since the client should now have a cached token
3. It only makes one call - no retry should be needed since the token is already available
4. It validates the exact token - checks that the correct cached token is being used

test/client/client_auth.spec.ts
mprynce: The token68 after the "Agentic" is the challenge which in production is signed using the agents private key, and resent as a JWT.  These tests don't include creating the signature or verify signature tests.

test/client/client_auth.spec.ts
mprynce: The new fake on line 327 is needed because the test is transitioning from testing sequential authentication behavior to testing concurrent authentication behavior. It's basically the same as the answer to the earlier question about removing the second fake

mprynce:
I've made all the requested changes, and given reasons (I believe) why some of the tests were appropriate.  I've also updated the auth-handler.ts to explain why the onSuccess() call is necessary and the suggeted headers should not be saved until onSuccess is called.

Thank you so much for all the feedback and hoping this branch is ready to merge :)

src/client/client.ts
```diff
@@ -31,45 +31,53 @@ import {
   A2AError,
   SendMessageSuccessResponse
 } from '../types.js'; // Assuming schema.ts is in the same directory or appropriately pathed
+import { AuthenticationHandler, HttpHeaders } from './auth-handler.js';
 
 // Helper type for the data yielded by streaming methods
 type A2AStreamEventData = Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent;
 
+export interface A2AClientOptions {
+  authHandler?: AuthenticationHandler;
```
swapydapy: My proposal would be to keep the client more flexible, at varying levels of complexity.

Just like Axios, we can have requestConfig. A request config can used to specify:
1. headers (pre-determined): simplest.
2. requestInterceptors (loads & refreshes creds on-demand): Little more involved.
3. Supply your own fetch: Most advanced.

A2AClientOptions can have agentCardPath & requestConfig.

The current authHandler is basically intercepting, injecting headers & re-running the request in case of auth. This should be modeled outside the client, as custom fetch => fetchWithAuthRetry.

I suggest we use the `AuthenticationHandler` through a factory which generates a custom fetch.

const fetchWithAuth = (authHandler: AuthenticationHandler) = {
  return () => {
     ... handle request/response & retry
  }
}

test/client/client_auth.spec.ts
swapydapy: Since it is already missing an auth-header, the last line with "// Any other case without auth header should fail" would return the auth challenge

test/client/client_auth.spec.ts
swapydapy: I would suggest to use well known auth schemes in this authenticate header. Something like:
Bearer client_id="some-id", authorization_uri="https://agent-path/oauth2/authorize", error="insufficient_claims", claims="call_api"

test/client/client_auth.spec.ts
swapydapy: These tracking of fetchCounts makes the test brittle and susceptible to busg. The test are complex to reason about. Can we simplify the fake, by returning 401 when header is not present otherwise, some responses. Also, lets not use global fake, instead create one for each test.

swapydapy (changes requested):

test/client/client_auth.spec.ts
swapydapy: I meant, we can use the simplified fake across all tests.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,963 @@
...
+      const noRetryHandler = new MockAuthHandler();
+      const originalShouldRetry = noRetryHandler.shouldRetryWithHeaders.bind(noRetryHandler);
+      noRetryHandler.shouldRetryWithHeaders = sinon.stub().resolves(undefined);
+
+      const clientNoRetry = new A2AClient('https://test-agent.example.com', {
+        authHandler: noRetryHandler,
+        fetchImpl: mockFetch
+      });
+
+      const messageParams: MessageSendParams = {
```
swapydapy: can we create these as constants and re-use across tests. Same for response for for auth fake. That would substantially improve the readability.

swapydapy (changes requested):

test/client/client_auth.spec.ts
mprynce: I'm working on a big refactor to make the tests much easier :)

test/client/client_auth.spec.ts
mprynce: I reworked that whole section... much simpler now.  It still does the full test pass of missing auth -> send client challenge -> client signs challenge and reissues request -> server confirms signature.  I've moved all the auth code into a fetchImpl (as you suggested).  More on that later...

test/client/client_auth.spec.ts
mprynce: Huge refactor.  Separate client.spec.ts and client_auth.spec.ts

Hopefully this new code better follows the advice you gave above

src/client/client.ts
mprynce: Followed the spirit of the above advice and removed the AuthenticationHandler from the A2AClient options and instead wrapped it in a fetchImpl.  Much simpler API now and the fetchImpl can control the retries with dynamic authentication.

test/client/client_auth.spec.ts
mprynce: beforeEach() and afterEach() do cleanup including the mockFetch which is probably a good idea.  mockFetch.reset() is also used while tracking function invocation counts for testing.

And yes, I went through and reduced redundant code in many places including messageParams.

mprynce:
Did a big refactor, including moving the auth handler out of A2AClient and instead using a fetch wrapper.  Divided client testing into basic testing, and auth testing files.

src/client/index.ts
```diff
@@ -3,3 +3,4 @@
  */
 
 export { A2AClient } from "./client.js";
```
adamfweidman, line 5: export A2AClientOptions as well

mprynce: Added export - thank you for the catch!

src/client/auth-handler.ts
swapydapy: I understand that use-case. Though, this should be the developer's choice to need this onSuccessful(). They can very well detect that request failed because creds expired, and they can refresh the creds and save it for future requests.

In such cases, the onSuccess is not really needed. Or cases where they supply static headers. In such cases, a no-op implementation of `onSuccess` needs to be provided.

That's why I suggest to not enfore on the develope to always provide the `onSuccess`.

Also I would like to rename `onSuccess` to `onSuccessfulRetry` to clearly convey the behaviour.

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Handle HTTP 401 and 403 error codes from fetch results. If the shouldRetryWithHeaders
+ * function returns revised headers, then retry the request using revised headers and report
+ * success with onSuccess().
+ */
+export interface AuthenticationHandler {
+    /**
+     * Provides additional HTTP request headers.
+     * @returns HTTP headers which should include Authorization if available.
+     */
+    headers: () => HttpHeaders;
```
swapydapy: This needs to be return a Promise. As devs most probably would need to go through network or read localStorage to fetch creds.

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Called to check if the HTTP request should be retried with *new* headers.  This usually
```
swapydapy: The message is a little misleading. It seems as if this method is only called on request failures. So repharse with something like: "This method will be always called after each request is executed. Handler can check if the there auth related failures and if the request needs to be retried with revised headers."

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+     * occours when the HTTP response issues a 401 or 403.  If this
+     * function returns new HTTP headers, then the request should be retried with
+     * the revised headers.
+     *
+     * Note that the new headers returned by this request are transient, and will only be saved
```
swapydapy: My understanding is that the consumer of AuthenticationHandler is not responsible for saving the revised headers. If yes, then we need to fix the comment here.

"Note that the new headers returned by this method are only used for retrying the particular request. If it succeeds, then onSuccess() method will be called. AuthenticationHandler can choose to only update headers when onSuccess()"

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+      ...(init || {}),
+      headers: {
+        ...authHeaders,
+        ...(init?.headers || {}),
+      },
+    };
+
+    let response = await this.fetchImpl(url, mergedInit);
+
+    // Check for HTTP 401/403 and retry request if necessary
```
swapydapy: Rephrase this comment as this AuthHandlingFetch is not checking for 401/403. The authHandler can possibly check for auth errors.

src/client/client.ts
```diff
@@ -35,41 +35,45 @@ import {
 // Helper type for the data yielded by streaming methods
 type A2AStreamEventData = Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent;
 
+export interface A2AClientOptions {
+  agentCardPath?: string;
+  fetchImpl?: typeof fetch;
```
swapydapy, line 41: Can we put the fetchImpl in another interface called RequestConfig. This way other options can be added at requestConfig (like timeout). In future, that requestConfig can be overridden per API call as well.

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param authHandler Authentication handler for managing auth headers and retries
+   */
+  constructor(fetchImpl: typeof fetch, authHandler: AuthenticationHandler) {
+    super();
+    this.fetchImpl = fetchImpl;
+    this.authHandler = authHandler;
+    
+    // Make the instance callable
+    const boundFetch = this._executeFetch.bind(this);
+    Object.setPrototypeOf(boundFetch, AuthHandlingFetch.prototype);
```
swapydapy: This class structure is difficult to comprehend. As suggested in my previous comment, could you refactor this into higher-order-function. That would be much easier to go through and serve as sample for providing custom fetch.

const authenticatedFetch = (fetchImpl: typeof fetch, authHandler: AuthenticationHandler) => {

... return () => {}
}

test/client/util.ts
```diff
@@ -0,0 +1,221 @@
...
+    status = status !== 200 ? status : 500;
+  } else {
+    jsonRpcResponse.result = result;
+  }
+  
+  // Create a fresh body each time to avoid "Body is unusable" errors
+  const body = JSON.stringify(jsonRpcResponse);
+  
+  // Create a ReadableStream to ensure the body can be read multiple times
+  const stream = new ReadableStream({
```
swapydapy: Any reason to create these streams for generating hardcoded responses. This seems to be generated by LLM.

test/client/client.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+// Factory function to create fresh mock fetch functions
+function createMockFetch() {
+  return sinon.stub().callsFake(async (url: string, options?: RequestInit) => {
+    // Create a fresh mock fetch for each call to avoid Response body reuse issues
+    return createFreshMockFetch(url, options);
+  });
+}
+
+// Helper function to create fresh mock fetch responses
+function createFreshMockFetch(url: string, options?: RequestInit) {
```
swapydapy: What is the point of this method (naming seems to be incorrect as well). This code can be directly added to `createMockFetch()`

test/client/client.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+      expect(agentCard.capabilities).to.have.property('streaming', true);
+      expect(agentCard.capabilities).to.have.property('pushNotifications', true);
+    });
+
+    it('should cache agent card for subsequent requests', async () => {
+      // First call
+      await client.getAgentCard();
+      
+      // Reset fetch mock
+      mockFetch.reset();
```
swapydapy: No need to reset this. Just check the call count to be 2

test/client/client.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+        await missingUrlClient.getAgentCard();
+        expect.fail('Expected error to be thrown');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Error);
+        expect((error as Error).message).to.include("does not contain a valid 'url'");
+      }
+    });
+  });
+
+  describe('Request ID Management', () => {
```
swapydapy: Not a valid test. Any reason for adding this test.

test/server/default_request_handler.spec.ts
```diff
@@ -208,7 +208,9 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
 
     it('sendMessage: should handle agent execution failure for blocking calls', async () => {
         const errorMessage = 'Agent failed!';
-        (mockAgentExecutor as MockAgentExecutor).execute.rejects(new Error(errorMessage));
+        (mockAgentExecutor as MockAgentExecutor).execute.callsFake(async () => {
```
swapydapy: Not needed

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { A2AClient } from '../../src/client/client.js';
+import { AuthenticationHandler, HttpHeaders, AuthHandlingFetch } from '../../src/client/auth-handler.js';
+import { AgentCard, MessageSendParams, TextPart, Message, SendMessageResponse, SendMessageSuccessResponse } from '../../src/types.js';
+import { extractRequestId, createResponse, createAgentCardResponse, createMockAgentCard, createMessageParams, createMockMessage } from './util.js';
+
+
+// Challenge manager class for authentication testing
+class ChallengeManager {
```
swapydapy, line 15: I suggest we simplify the whole testing by just sending an 401 header code in response to auth header missing. Then SuppliedAuthHandler can add a bearer auth header.

This challenge manager is out of the scope of this library and significantly increases the comlixty of these tests. Doing so moves away the focus from testing the auth retry part to this auth handshake mechanism. This makes it hard to reason about the tests and increases the chances of bugs.

test/client/client_auth.spec.ts
swapydapy: Doesn't make sense to me. The mockFetch is simulating the server right. The server doesn't need to change with each request. Instead we need to assert the behaviour by looking at calls to your authHandler.

For second test, when cached token needs to be used. You should check that the call to mockFetch already has the token which was generated in first call. No need to change the implementation of the fetch itself.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+          
+          return createAgentCardResponse(mockAgentCard);
+        }
+        
+        if (url.includes('/api')) {
+          const authHeader = options?.headers?.['Authorization'] as string;
+          
+          // If no auth header, return 401 to trigger auth flow
+          if (!authHeader) {
+            const requestId = extractRequestId(options);
```
swapydapy: At this point we are just repeating the same code again and again.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+        await clientHeaderTest.sendMessage(messageParams);
+        expect.fail('Expected error to be thrown');
+      } catch (error) {
+        // Verify that the WWW-Authenticate header was returned
+        expect(capturedResponse).to.not.be.null;
+        expect(capturedResponse!.headers.get('WWW-Authenticate')).to.equal('Agentic challenge123');
+      }
+    });
+
+    it('should parse WWW-Authenticate header and generate correct Authorization header', async () => {
```
swapydapy: Out of the scope for this auth-handler-fetch lib code

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+      expect(result).to.have.property('error');
+      expect((result as any).error).to.have.property('code', -32001);
+      expect((result as any).error).to.have.property('message', 'Authentication required');
+
+      // Verify that fetch was called only once (no retry attempted)
+      expect(noAuthHandlerFetch.callCount).to.equal(2); // One for agent card, one for API call
+    });
+  });
+
+  describe('Error Handling', () => {
```
swapydapy: These are repeated tests from client.spec.tests

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+      try {
+        await clientWithMalformed.sendMessage(messageParams);
+        expect.fail('Expected error to be thrown');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Error);
+      }
+    });
+  });
+
+  describe('Agent Card Caching', () => {
```
swapydapy: Again duplicate

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,977 @@
...
+      expect(typeof authHandlingFetch).to.equal('function');
+      expect(authHandlingFetch).to.be.instanceOf(AuthHandlingFetch);
+    });
+
+    it('should support direct function calls', async () => {
+      const response = await authHandlingFetch('https://test.example.com/api');
+      expect(response).to.be.instanceOf(Response);
+    });
+
+    it('should support fetch method calls', async () => {
```
swapydapy: ??????

swapydapy (changes requested):
Thanks @mprynce for the changes. Added comments regarding the structure of AuthHadnlingFetch & tests.

I understand we are using LLMs for assisted coding, but please go through the generated code with fine-toothed comb.

Otherwise the responsibility of making sure there are no bugs lies entirely on the reviewers, which significantly makes adds to their burden.

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * A fetch wrapper that handles authentication logic including retries for 401/403 responses.
+ * This class can be used as a drop-in replacement for the native fetch function.
+ * 
+ * Usage examples:
+ * - const authFetch = new AuthHandlingFetch(fetch, authHandler);
+ * - const response = await authFetch(url, options);
+ * - const response = await authFetch(url); // Direct function call
+ */
+export class AuthHandlingFetch extends Function {
```
swapydapy: nit: rename to AuthenticatedFetchWithRetry

src/client/auth-handler.ts
mprynce: Good catch on the headers() promise - I've updated that code.  Also updated to onSuccessfulRetry().

Expired credentials is only one authentication failure mode.  Other modes are:
- Client signing a challenge with invalid private keys (outside it's knowledge)
- Client using a signing algorithm that is not supported by the server.
 
For these cases the client **should not assume success**, and instead wait for verification from the server that the Authorization header was accepted.

NOTE that above are outside the centralized authentication server scenarios, and instead use decentralized signing systems.

The onSuccessfulRetry() provides a separation of concerns where the AuthenticationHandler can simply handle (1) cached authorization headers, (2) generating new headers, and (3) caching the headers that worked.  We *could* move that logic into something similar to the AuthHandlingFetch() class, but that overloads that implementation and makes it unnecessarily complex.

mprynce: Also made onSuccessfulRetry() optional, for situations where the new authentication header is guaranteed to be valid.

src/client/auth-handler.ts
mprynce: It's been updated to be more broad: "Check if the auth handler wants to retry the request with new headers"

src/client/auth-handler.ts
mprynce: Updated to "This method will be always called after each request is executed.  Handler can check if there are auth related failures and if the request needs to be retried with revised headers."

src/client/auth-handler.ts
mprynce: New wording "Note that the new headers returned by this request may be transient, and might only be saved
     when the onSuccessfulRetry() function is called, or otherwise discarded.  This is an
     implementation detail of an AuthenticationHandler.  If the headers are transient, then
     the onSuccessfulRetry() function should be implemented to save the headers for subsequent
     requests."

src/client/client.ts
mprynce: I'm looking at client.ts:_fetchRpc(...,requestConfig) which would likely be the recipient of something like a sendMessage(...,requestConfig) and wondering exactly how they would integrate together.

Target is:

    return this.fetchImpl(url, requestInit)

To:

   const fetcher = requestConfig?.fetchImpl ?? this.fetchImpl;
   return fetcher(url, requestInit);


Since fetch doesn't support timeouts, the typical solution is to implement a wrapper such as:

  const fetchWithTimeout = (url, options = {}, timeout = 5000) => {
    const controller = new AbortController();
    const id = setTimeout(() => controller.abort(), timeout);

    return fetch(url, {
      ...options,
      signal: controller.signal
    })
    .finally(() => clearTimeout(id)); // Clear the timeout if the fetch completes
};

And this circles us back to more simply creating a fetch wrapper.

Are we getting into scope creep?

src/client/auth-handler.ts
mprynce: onSuccessfulRetry is now optional

src/client/auth-handler.ts
mprynce: Converted AuthHandlingFetch to HOC with createAuthenticatingFetchWithRetry()

test/client/util.ts
mprynce: I believe this code was added in response to errors from earlier code.  The defensive design is best practice, but the current tests do not cause this code to be needed, so I have removed it.

test/client/client.spec.ts
mprynce: This design avoids the issues from response body reuse (the other defensive code removed from a prior change request)

test/client/client.spec.ts
mprynce: Removed reset and fixed call count to 1

test/client/client.spec.ts
mprynce: Removed

test/server/default_request_handler.spec.ts
mprynce: Reverted.  I had added this to handle errors with previous versions - seems to be working now.

test/client/client_auth.spec.ts
mprynce: This is a real world test and reduces the chances of people adding breaking changes.  I can remove it if you'd like, but we will have greatly reduced coverage.

test/client/client_auth.spec.ts
mprynce: Soak tests are ok, but I've removed it.

test/client/client_auth.spec.ts
mprynce: Removed the WWW-Authenticate header test

test/client/client_auth.spec.ts
mprynce: Removed

test/client/client_auth.spec.ts
mprynce: Removed

test/client/client_auth.spec.ts
mprynce: Yep - LLM fail.  I cleaned it up.

src/client/auth-handler.ts
mprynce: Renamed to createAuthenticatingFetchWithRetry - hope that's ok?

test/client/client_auth.spec.ts
mprynce: Please confirm you'd like me to remove the ChallengeManager and the related coverage tests - thank you!

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
+export interface HttpHeaders { [key: string]: string };
+
+/**
+ * Generic interface for handling authentication for HTTP requests.
+ * 
+ * An example flow using Universal Authentication ([DID](https://w3c.github.io/did/) based signing
```
swapydapy: Since the scope of this auth lib is general and not specific to one auth scheme. Let's keep the instructions general as well.

Lets remove the current specific sample of Agentic JWT based flow.

```
An example auth flow:
1. First HTTP request adds the headers from the headers() method to the request. Typically, these headers
   are supplied by establishing an auth session and caching those credentials for subsequent requests.
2. For every HTTP response (even 200s) the shouldRetryWithHeaders() method is called. This method is supposed to check if the request needs to be retried and 	 if, yes, return a set of headers. An A2A server might indicate auth failures in its response by JSON-rpc codes, HTTP codes like 401, 403 or headers like WWW-Authenticate.
3. If the HTTP request is successful, then onSuccessfulRetry() method is called to signal the
   authentication was accepted by the server. AuthenticationHandler can choose to safely cache those headers for subsequent requests.
```

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+    headers: () => Promise<HttpHeaders>;
+
+    /**
+     * This method will be always called after each request is executed.  Handler can check if
+     * there are auth related failures and if the request needs to be retried with revised headers.
+     *
+     * New headers are usually needed when the HTTP response issues a 401 or 403.  If this function returns
+     * new HTTP headers, then the request should be retried with the revised headers.
+     *
+     * Note that the new headers returned by this request may be transient, and might only be saved
```
swapydapy: This comment about transient is not needed here. Instead can be specified for onSuccessfulRetry

src/client/auth-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+     * Provides additional HTTP request headers.
+     * @returns HTTP headers which may include Authorization if available.
+     */
+    headers: () => Promise<HttpHeaders>;
+
+    /**
+     * This method will be always called after each request is executed.  Handler can check if
+     * there are auth related failures and if the request needs to be retried with revised headers.
+     *
+     * New headers are usually needed when the HTTP response issues a 401 or 403.  If this function returns
```
swapydapy: Please see my comment for not being too specific to one way of auth.

"For every HTTP response (even 200s) the shouldRetryWithHeaders() method is called. This method is supposed to check if the request needs to be retried and 	 if, yes, return a set of headers. An A2A server might indicate auth failures in its response by JSON-rpc codes, HTTP codes like 401, 403 or headers like WWW-Authenticate.
"

test/client/util.ts
```diff
@@ -0,0 +1,210 @@
...
+  headers: Record<string, string> = {}
+): Response {
+  const defaultHeaders = { 'Content-Type': 'application/json' };
+  const responseHeaders = { ...defaultHeaders, ...headers };
+  
+  // Create a fresh body each time to avoid "Body is unusable" errors
+  const body = JSON.stringify(data);
+  
+  // Create a ReadableStream to ensure the body can be read multiple times
+  const stream = new ReadableStream({
```
swapydapy: no need of this readable stream. Just return string.

test/client/client_auth.spec.ts
swapydapy: Yes, please. We should only test for the behaviour expected from `createAuthenticatingFetchWithRetry`.
1. Supply some headers in the request from supplied AuthHandler.
2. Retries for a given response, and with new headers for it. `createAuthenticatingFetchWithRetry` doesn;t care how suplied authHAndler deemed a response to be retryable, also whether new headers are correct. We just verify that they should be the some, which are generated from calling shouldRetry() from supplied auth handler.
3. When retried request is successful, the OnSuccess of supplied AuthHandler is called.

swapydapy: I understand you want to provide an end-to-end example. But that should go into another repo as an sample.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+    const challenge = challengeManager.createChallenge();
+
+    return createResponse(requestId, undefined, {
+      code: -32001,
+      message: 'Authentication required'
+    }, 401, { 'WWW-Authenticate': `Agentic ${challenge}` });
+  }
+
+  // We have an auth header, so make sure the scheme is correct
+  const [ scheme, params ] = authHeader.split(/\s+/);
```
swapydapy: For the purposes of testing retryable auth, we can just test if there's a auth header. No need to check auth scheme or verify tokens.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+    // Parse WWW-Authenticate header to extract the token68/challenge value
+    const [scheme, challenge] = res.headers.get('WWW-Authenticate')?.split(/\s+/) || [];
+    if (scheme !== 'Agentic')
+      return undefined;  // Not handled, only Agentic is supported
+
+    // Use the ChallengeManager to sign the challenge
+    const token = ChallengeManager.signChallenge(challenge);
+      
+    // have the client try the token, BUT don't save it in case the client doesn't accept it
+    return { 'Authorization': `Agentic ${token}` };
```
swapydapy: Please be generic here. Like: "some-auth-scheme"

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      const messageParams = createMessageParams({
+        messageId: 'test-msg-2',
+        text: 'Second message'
+      });
+
+      // First request - should trigger auth flow
+      await client.sendMessage(messageParams);
+      
+      // Reset calls
+      mockFetch.reset();
```
swapydapy: Not needed right, since we are creating a new fetch for each test run in beforeEach().

test/client/client_auth.spec.ts
swapydapy: This comment is not resolved.

test/client/client_auth.spec.ts
swapydapy: This test should be called "Retry with new auth headers". The whole authenticate header is internal detail to AuthHandler. This lib doesn't care about that. For all it matters, the server just needs header named API-KEY and specifies that as error code with no WWW-Authenticate header.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+        expect.fail('Expected error to be thrown');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Error);
+      }
+    });
+
+    it('should parse WWW-Authenticate header and generate correct Authorization header', async () => {
+      // Create a mock that tracks the Authorization headers sent
+      let capturedAuthHeaders: string[] = [];
+      const authHeaderTestFetch = sinon.stub().callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: Instead of re-supplying another copy of fetch, with the only difference of capturing auth-headers. Could you do following:
1. Either capture headers in createMockFetch(). All other tests can also use it, and reduces code duplication.
2. Just check args to the mock Fetch stub: stub.getCall(n).args

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      expect(capturedAuthHeaders[1]).to.match(/^Agentic .+$/); // Second call: with Agentic auth header
+
+      // Verify the result
+      expect(isSuccessResponse(result)).to.be.true;
+    });
+
+    it('should continue without authentication when server does not return 401', async () => {
+      // Create a mock that doesn't require authentication
+      let capturedAuthHeaders: string[] = [];
+      const noAuthRequiredFetch = sinon.stub().callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: Instead of let's re-use createMockFetch, and an optional param to check for auth. The value here could be False.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      expect(isSuccessResponse(result)).to.be.true;
+      if (isSuccessResponse(result)) {
+        // Check if result is a Message1 (which has messageId) or Task2
+        if ('messageId' in result.result) {
+          expect(result.result.messageId).to.equal('msg-no-auth-required');
+        }
+      }
+    });
+
+    it('should fail gracefully when no authHandler is provided and server returns 401', async () => {
```
swapydapy: We are not failing gracefully here, right. Just conveying the server error. The test should be called "Client pipes server errors when no auth handler is specified"

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      // Create a mock that returns 401 without authHandler
+      const noAuthHandlerFetch = sinon.stub().callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: nit: fetchWithApiError

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+            jsonrpc: '2.0',
+            error: {
+              code: -32001,
+              message: 'Authentication required'
+            },
+            id: requestId
+          });
+          
+          // Create a Response that can be read multiple times
+          const stream = new ReadableStream({
```
swapydapy: not needed

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+
+    it('should support direct function calls', async () => {
+      const response = await authHandlingFetch('https://test.example.com/api');
+      expect(response).to.be.instanceOf(Response);
+    });
+  });
+
+  describe('Header Merging', () => {
+    it('should merge auth headers with provided headers', async () => {
+      const authHandlerSpy = sinon.spy(authHandler, 'headers');
```
swapydapy: Don't think auth-handler added any headers, then how is this test asserting the merge behaviour.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      const emptyAuthFetch = createAuthenticatingFetchWithRetry(mockFetch, emptyAuthHandler);
+      
+      await emptyAuthFetch('https://test.example.com/api');
+      
+      const fetchCall = mockFetch.getCall(0);
+      expect(fetchCall.args[1]).to.exist;
+    });
+  });
+
+  describe('Authentication Retry Logic', () => {
```
swapydapy: This whole is repeat of tests written for A2A client. Please remove these or refactor those test to be at this level, just test that A2AClient uses custom fetch.

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+        await errorAuthFetch('https://test.example.com/api');
+        expect.fail('Expected error to be thrown');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Error);
+        expect((error as Error).message).to.include('Auth handler error');
+      }
+    });
+  });
+
+  describe('Integration with A2AClient', () => {
```
swapydapy: remove this one

test/client/client_auth.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+  describe('Success Callback', () => {
+    it('should call onSuccess when retry succeeds', async () => {
+      const successAuthHandler = new MockAuthHandler();
+      const onSuccessSpy = sinon.spy(successAuthHandler, 'onSuccessfulRetry');
+      
+      const successFetch = createAuthenticatingFetchWithRetry(mockFetch, successAuthHandler);
+      
+      // Mock fetch to return 401 first, then 200
+      let callCount = 0;
+      const successMockFetch = sinon.stub().callsFake(async (url: string, options?: RequestInit) => {
```
swapydapy: Should use existing mockFetch.

swapydapy (changes requested):

swapydapy:
The main code looks fine now., though the tests need to cleaned-up.

As mentioned, the tests needs to assert intended library behaviour. The library is general in nature and allows retrying a request when AuthHandler deems a response to be retryable.

It has not specific to one way of auth scheme or failures.

The PR can be merged if you can cleanup the tests. A lot of folks are asking for this auth support. Thank you for your contribution.

mprynce:
> The main code looks fine now., though the tests need to cleaned-up.
> 
> As mentioned, the tests needs to assert intended library behaviour. The library is general in nature and allows retrying a request when AuthHandler deems a response to be retryable.
> 
> It has not specific to one way of auth scheme or failures.
> 
> The PR can be merged if you can cleanup the tests. A lot of folks are asking for this auth support. Thank you for your contribution.

Thank you so much for all the attention to detail and great feedback.  I'll get to work on this ASAP!

swapydapy:
Let me know the ETA you have in mind for addressing these comments. Might need to send a PR to allow specifying custom fetch in client.

mprynce:
> Let me know the ETA you have in mind for addressing these comments. Might need to send a PR to allow specifying custom fetch in client.

Shooting for tonight :)

src/client/auth-handler.ts
mprynce: Removed Agentic JWT example and replaced with general description

src/client/auth-handler.ts
mprynce: Removed, and already in place in the onSuccessfulRetry() docs

src/client/auth-handler.ts
mprynce: Revised to suggested wording

test/client/util.ts
mprynce: Removed unecessry second createFreshMockFetch, and removed use of ReadableStream when simple string for response body works

test/client/client_auth.spec.ts
mprynce: Removed

test/client/client_auth.spec.ts
mprynce: Fixed test to check try and test for merged headers

test/client/client_auth.spec.ts
mprynce: Removed

test/client/client_auth.spec.ts
mprynce: Removed

test/client/client_auth.spec.ts
mprynce: Updated to use existing mockFetch

test/client/client_auth.spec.ts
mprynce: Renamed the test to the suggestion and relaxed the test to accept any auth header in second response

test/client/client_auth.spec.ts
mprynce: Simplified code while retaining defensive code. Tested without .reset() and .callsFake and unit tests fail.

Updated test to verify token generated from first call is used in second call.

test/client/client_auth.spec.ts
mprynce: Removed Authorization header scheme and token verification from tests

test/client/client_auth.spec.ts
mprynce: Revised to conventional Bearer

test/client/client_auth.spec.ts
mprynce: Removed reset() and revised callCount to match new values

test/client/client_auth.spec.ts
mprynce: Big refactor of the mock fetch factories into a common createMockFetch that is configurable.  Also attached capturedAuthHeaders to the mock to reduce code duplication

test/client/client_auth.spec.ts
mprynce: Yes, made it configurable

test/client/client_auth.spec.ts
mprynce: Renamed

test/client/client_auth.spec.ts
mprynce: Renamed

swapydapy (approved):

PR #49, opened 2025-07-10
fix: setting context id in _createRequestContext

### LINKED ISSUE #41: [Question]: Why _createRequestContext sets different context IDs for task and incoming message?
In `default_request_handler._createRequestContext` we have this code block:
```typescript
        // Ensure contextId is present
        const messageForContext = { ...incomingMessage };
        if (!messageForContext.contextId) {
            messageForContext.contextId = task?.contextId || uuidv4();
        }

        const contextId = incomingMessage.contextId || uuidv4();
``` 

If there is no previous context id and no previous task, different context ids will be created. Is there a reason for this?

clriesco:
# Description

- [X] Follow the [`CONTRIBUTING` Guide](https://github.com/google-a2a/a2a-js/blob/main/CONTRIBUTING.md).
- [X] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [X] Ensure the tests and linter pass
- [ ] Appropriate docs were updated (if necessary)

Fixes #41 🦕

### DIFF
--- a/test/server/default_request_handler.spec.ts
+++ b/test/server/default_request_handler.spec.ts
@@ -4,7 +4,8 @@ import sinon, { SinonStub, SinonFakeTimers } from 'sinon';
 
 import { AgentExecutor } from '../../src/server/agent_execution/agent_executor.js';
 import { describe, beforeEach, afterEach, it } from 'node:test';
-import { RequestContext, ExecutionEventBus, TaskStore, InMemoryTaskStore, DefaultRequestHandler, AgentCard, Artifact, Message, MessageSendParams, PushNotificationConfig, Task, TaskIdParams, TaskPushNotificationConfig, TaskState, TaskStatusUpdateEvent } from '../../src/index.js';
+import { RequestContext, ExecutionEventBus, TaskStore, InMemoryTaskStore, DefaultRequestHandler } from '../../src/server/index.js';
+import { AgentCard, Artifact, Message, MessageSendParams, PushNotificationConfig, Task, TaskIdParams, TaskPushNotificationConfig, TaskState, TaskStatusUpdateEvent } from '../../src/index.js';
 import { DefaultExecutionEventBusManager, ExecutionEventBusManager } from '../../src/server/events/execution_event_bus_manager.js';
 import { A2ARequestHandler } from '../../src/server/request_handler/a2a_request_handler.js';
 
@@ -599,13 +600,14 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
             message: {
                 messageId: 'msg-ctx2',
                 role: 'user',
-                parts: [{ kind: 'text', text: 'Hola' }],
+                parts: [{ kind: 'text', text: 'Hi' }],
                 kind: 'message',
                 taskId,
             },
         };
+        let capturedContextId: string | undefined;
         (mockAgentExecutor.execute as SinonStub).callsFake(async (ctx, bus) => {
-            expect(ctx.contextId).to.equal(taskContextId);
+            capturedContextId = ctx.contextId;
             bus.publish({
                 id: ctx.taskId,
                 contextId: ctx.contextId,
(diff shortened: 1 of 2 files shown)


test/server/default_request_handler.spec.ts
```diff
@@ -562,4 +562,82 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+            message: {
+                messageId: 'msg-ctx3',
+                role: 'user',
+                parts: [{ kind: 'text', text: 'Hola' }],
+                kind: 'message',
+            },
+        };
+        (mockAgentExecutor.execute as SinonStub).callsFake(async (ctx, bus) => {
+            expect(ctx.contextId).to.be.a('string').and.not.empty;
+            expect(ctx.contextId).to.not.equal('incoming-ctx-id');
```
swapydapy: The below two assertions are not needed, as those string values bear no meaning in this isolated testcase.

test/server/default_request_handler.spec.ts
```diff
@@ -562,4 +562,82 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+            message: {
+                messageId: 'msg-ctx2',
+                role: 'user',
+                parts: [{ kind: 'text', text: 'Hola' }],
+                kind: 'message',
+                taskId,
+            },
+        };
+        (mockAgentExecutor.execute as SinonStub).callsFake(async (ctx, bus) => {
+            expect(ctx.contextId).to.equal(taskContextId);
```
swapydapy: My sense is, we should not put expect statements inside a stub for a fake. Reason being, that if there's a bug and that method ("execute") itself is not called, then the "expect" will also not be called. The testcase would happily pass, and the bug will not be caught.

Instead, why don't we capture the contextId from the fake. Then, after the sendMessage method call, add an expect for captured `contextId`. This way, we always execute the expect assertion.

swapydapy (changes requested):

test/server/default_request_handler.spec.ts
clriesco: Done

test/server/default_request_handler.spec.ts
clriesco: Done

test/server/default_request_handler.spec.ts
```diff
@@ -562,4 +563,84 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+            message: {
+                messageId: 'msg-ctx',
+                role: 'user',
+                parts: [{ kind: 'text', text: 'Hola' }],
+                kind: 'message',
+                contextId: 'incoming-ctx-id',
+            },
+        };
+        (mockAgentExecutor.execute as SinonStub).callsFake(async (ctx, bus) => {
+            expect(ctx.contextId).to.equal('incoming-ctx-id');
```
swapydapy: nit: could you also capture the contextID here as well and add anexpect outside the fake function.

test/server/default_request_handler.spec.ts
```diff
@@ -562,4 +563,84 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+            bus.publish({
+                id: ctx.taskId,
+                contextId: ctx.contextId,
+                status: { state: "submitted" },
+                kind: 'task'
+            });
+            bus && bus.finished && bus.finished();
+        });
+        await handler.sendMessage(params);
+        expect(capturedContextId).to.be.a('string').and.not.empty.and.not.equal('incoming-ctx-id');
```
swapydapy: Any reason why we have added this check for "and.not.equal('incoming-ctx-id')"?

The executor has never seen "incoming-ctx-id'". Unless you create a task with this Id, and then start a new message, then you can possibly add this check, to make sure a new contextID is generated.

swapydapy (changes requested):

swapydapy:
The code changes look fine. I think, the commits from PR https://github.com/a2aproject/a2a-js/pull/61, also got merged here.

Could you submit that PR first.

clriesco:
yes, got a mess in git. Now the branch is reseted and rebased.

swapydapy:
Thanks. Need to resolve some conflicts.

PR #102, opened 2025-08-08
fix: fix Incorrect Well-Known Path for Agent Card

### LINKED ISSUE #99: [Bug]: Incorrect Well-Known Path for Agent Card
### Description
According to the [v3.0 A2A protocol specification](https://a2a-protocol.org/latest/specification/#53-recommended-location), the Agent Card should be served at the following well-known path: `/.well-known/agent-card.json`. However, in version 0.3.0, the current implementation still uses the outdated path: `/.well-known/agent.json`

### Source
https://github.com/a2aproject/a2a-js/blob/main/src/server/express/a2a_express_app.ts#L32
The outdated path is defined in the following file:
```ts
router.get("/.well-known/agent.json", async (req: Request, res: Response) => {
    try {
        const agentCard = await this.requestHandler.getAgentCard();
        res.json(agentCard);
    } catch (error: any) {
        console.error("Error fetching agent card:", error);
        res.status(500).json({ error: "Failed to retrieve agent card" });
    }
});
```

Affected Version:
v0.3.0

### Relevant log output

```shell

```

### Code of Conduct

ronantakizawa:
# Description

Fix the well-known path for agent card from `/.well-known/agent.json` to `/.well-known/agent-card.json`

Fixes #99  🦕

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -31,7 +31,7 @@ import {
   A2AError,
   SendMessageSuccessResponse
 } from '../types.js'; // Assuming schema.ts is in the same directory or appropriately pathed
-import { AGENT_CARD_RELATIVE_PATH } from "../constants.js";
+import { AGENT_CARD_PATH } from "../constants.js";
 
 // Helper type for the data yielded by streaming methods
 type A2AStreamEventData = Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent;
@@ -44,7 +44,6 @@ export class A2AClient {
   private agentBaseUrl: string;
   private agentCardPath: string;
   private agentCardPromise: Promise<AgentCard>;
-  private static readonly DEFAULT_AGENT_CARD_PATH = AGENT_CARD_RELATIVE_PATH;
   private requestIdCounter: number = 1;
   private serviceEndpointUrl?: string; // To be populated from AgentCard after fetching
 
@@ -56,7 +55,7 @@ export class A2AClient {
    * @param agentBaseUrl The base URL of the A2A agent (e.g., https://agent.example.com)
    * @param agentCardPath path to the agent card, defaults to .well-known/agent-card.json
    */
-  constructor(agentBaseUrl: string, agentCardPath: string = A2AClient.DEFAULT_AGENT_CARD_PATH) {
+  constructor(agentBaseUrl: string, agentCardPath: string = AGENT_CARD_PATH) {
     this.agentBaseUrl = agentBaseUrl.replace(/\/$/, ""); // Remove trailing slash if any
     this.agentCardPath = agentCardPath.replace(/^\//, ""); // Remove leading slash if any
     this.agentCardPromise = this._fetchAndCacheAgentCard();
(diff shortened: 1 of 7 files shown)


ronantakizawa:
@swapydapy

src/server/express/a2a_express_app.ts
```diff
@@ -29,7 +29,7 @@ export class A2AExpressApp {
         const router = express.Router();
         router.use(express.json(), ...(middlewares ?? []));
 
-        router.get("/.well-known/agent.json", async (req: Request, res: Response) => {
+        router.get("/.well-known/agent-card.json", async (req: Request, res: Response) => {
```
swapydapy: Can we add a constant and also make it a parameter in the `setupRoutes()` method.

swapydapy (changes requested):

swapydapy:
Thanks for catching this.

src/server/express/a2a_express_app.ts
ronantakizawa: @swapydapy done!

src/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * The well-known path for the agent card endpoint
+ */
+export const AGENT_CARD_PATH = "/.well-known/agent-card.json";
+
+/**
+ * The relative well-known path for the agent card (without leading slash)
+ */
+export const AGENT_CARD_RELATIVE_PATH = ".well-known/agent-card.json";
```
swapydapy: Let's just have one constant: AGENT_CARD_RELATIVE_PATH, the second one is redundant.

src/client/client.ts
```diff
@@ -43,17 +44,17 @@ export class A2AClient {
   private agentBaseUrl: string;
   private agentCardPath: string;
   private agentCardPromise: Promise<AgentCard>;
-  private static readonly DEFAULT_AGENT_CARD_PATH = ".well-known/agent.json";
+  private static readonly DEFAULT_AGENT_CARD_PATH = AGENT_CARD_RELATIVE_PATH;
```
swapydapy: Any reason to keep this static class member? we can get rid of this as we now already have a constant.

src/server/express/a2a_express_app.ts
```diff
@@ -19,17 +20,19 @@ export class A2AExpressApp {
...
      * @param middlewares Optional array of Express middlewares to apply to the A2A routes.
+     * @param agentCardPath Optional custom path for the agent card endpoint (defaults to /.well-known/agent-card.json).
      * @returns The Express app with A2A routes.
      */
     public setupRoutes(
         app: Express,
         baseUrl: string = "",
-        middlewares?: Array<RequestHandler | ErrorRequestHandler>
+        middlewares?: Array<RequestHandler | ErrorRequestHandler>,
+        agentCardPath: string = AGENT_CARD_PATH
```
swapydapy, line 30: Would you be able to please add a test for A2AExpressApp, which verifies this functionality. I know we don't have a test for A2AExpressApp yet, so might need you to set it up. You can take it up in a separate PR as well.

src/constants.ts
swapydapy: could you also export this constant in https://github.com/a2aproject/a2a-js/blob/main/src/index.ts

swapydapy (changes requested):

ronantakizawa:
@swapydapy made your suggested changes. I'll make a new issue for adding tests for  A2AExpressApp.

src/server/express/a2a_express_app.ts
```diff
@@ -19,17 +20,19 @@ export class A2AExpressApp {
...
      * @param middlewares Optional array of Express middlewares to apply to the A2A routes.
+     * @param agentCardPath Optional custom path for the agent card endpoint (defaults to /.well-known/agent-card.json).
      * @returns The Express app with A2A routes.
      */
     public setupRoutes(
         app: Express,
         baseUrl: string = "",
-        middlewares?: Array<RequestHandler | ErrorRequestHandler>
+        middlewares?: Array<RequestHandler | ErrorRequestHandler>,
+        agentCardPath: string = `/${AGENT_CARD_PATH}`
```
swapydapy: One small nit: Can we follow the same constant style across the method API and the default constant. This way users can directly specify the constant here.

Can we make this default value to be AGENT_CARD_PATH. Then in router.get(`/${agentCardPath}`

swapydapy (changes requested):

ronantakizawa:
@swapydapy fixed the constant style and rebased to `main`

swapydapy (approved):

PR #108, opened 2025-08-12
feat: Added push notifications support

### LINKED ISSUE #89: [Feat]: Push Notifications Support

I want to implement push notifications for a2a agent. push notification configs are only stored in memory right now.

[https://github.com/a2aproject/a2a-js/blob/main/src/server/request_handler/default_request_handler.ts](https://github.com/a2aproject/a2a-js/blob/main/src/server/request_handler/default_request_handler.ts)

Create base interfaces and a defaultPushNotificationHandler, 
defaultRequestHandler should have the ability to handle sending push notifications if configuration provided.

Create my own request handler to support the push notifications protocol support.

### Additional context

The A2A platform supports sending PushNotificationConfig so clients can be updated of the tasks.

### Code of Conduct

botraunak:
# Description

Added push notifications support in the Server sdk.
Defines the configuration for setting up push notifications for task updates.

Added an `InMemoryPushNotificationStore`, refactored from within the `DefaultRequestHandler`.
Provided interface for push notification store which can be extended
Added a default implementation in `DefaultRequestHandler`.

Added support when pushNotification config available in `messageSendParams`

Existing push notification tests still pass
Added new test to ensure push notifications are being called

Fixes #89  🦕

### DIFF
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -36,6 +36,10 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
         pushConfig: PushNotificationConfig
     ): Promise<boolean> {
         const url = pushConfig.url;
+        const controller = new AbortController();
+        // Abort the request if it takes longer than 5 seconds.
+        const timeoutId = setTimeout(() => controller.abort(), 5000);
+
         try {
             const headers: Record<string, string> = {
                 'Content-Type': 'application/json'
@@ -48,7 +52,8 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             const response = await fetch(url, {
                 method: 'POST',
                 headers,
-                body: JSON.stringify(task)
+                body: JSON.stringify(task),
+                signal: controller.signal
             });
 
             if (!response.ok) {
@@ -60,6 +65,8 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
         } catch (error) {
             console.error(`Error sending push notification for task_id=${task.id} to URL: ${url}. Error:`, error);
             return false;
+        } finally {
+            clearTimeout(timeoutId);
         }
     }
 }
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -2,41 +2,55 @@ import { Task, PushNotificationConfig } from "../../types.js";
 import { PushNotificationSender } from "./push_notification_sender.js";
 import { PushNotificationStore } from "./push_notification_store.js";
 
+export interface DefaultPushNotificationSenderOptions {
+    /**
+     * Timeout in milliseconds for the abort controller. Defaults to 5000ms.
+     */
+    timeout?: number;
+    /**
+     * Custom header name for the token. Defaults to 'X-A2A-Notification-Token'.
+     */
+    tokenHeaderName?: string;
+}
+
 export class DefaultPushNotificationSender implements PushNotificationSender {
 
     private readonly pushNotificationStore: PushNotificationStore;
+    private readonly options: Required<DefaultPushNotificationSenderOptions>;
     
-    constructor(pushNotificationStore: PushNotificationStore) {
+    constructor(pushNotificationStore: PushNotificationStore, options: DefaultPushNotificationSenderOptions = {}) {
         this.pushNotificationStore = pushNotificationStore;
+        this.options = {
+            timeout: 5000,
+            tokenHeaderName: 'X-A2A-Notification-Token',
+            ...options
+        };
     }
 
     async send(task: Task): Promise<void> {
         const pushConfigs = await this.pushNotificationStore.load(task.id);
         if (!pushConfigs || pushConfigs.length === 0) {
-            return;
+            return Promise.resolve();
         }
 
         pushConfigs.forEach(pushConfig => {
             this._dispatchNotification(task, pushConfig)
-                .then(success => {
-                    if (!success) {
-                        console.warn(`Push notification failed to send for task_id=${task.id} to URL: ${pushConfig.url}`);
-                    }
-                })
                 .catch(error => {
                     console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
                 });
         });
+        // Return a resolved promise to indicate success
+        return Promise.resolve();
     }
 
     private async _dispatchNotification(
         task: Task, 
         pushConfig: PushNotificationConfig
-    ): Promise<boolean> {
+    ): Promise<void> {
         const url = pushConfig.url;
         const controller = new AbortController();
-        // Abort the request if it takes longer than 5 seconds.
-        const timeoutId = setTimeout(() => controller.abort(), 5000);
+        // Abort the request if it takes longer than the configured timeout.
+        const timeoutId = setTimeout(() => controller.abort(), this.options.timeout);
 
         try {
             const headers: Record<string, string> = {
@@ -44,7 +58,7 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             };
             
             if (pushConfig.token) {
-                headers['X-A2A-Notification-Token'] = pushConfig.token;
+                headers[this.options.tokenHeaderName] = pushConfig.token;
             }
 
             const response = await fetch(url, {
@@ -59,12 +73,12 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             }
 
             console.info(`Push notification sent for task_id=${task.id} to URL: ${url}`);
-            return true;
         } catch (error) {
+            // Ignore errors
             console.error(`Error sending push notification for task_id=${task.id} to URL: ${url}. Error:`, error);
-            return false;
         } finally {
             clearTimeout(timeoutId);
         }
+        return Promise.resolve();
     }
 }
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -39,8 +39,6 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
                     console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
                 });
         });
-        // Return a resolved promise to indicate success
-        return Promise.resolve();
     }
 
     private async _dispatchNotification(
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -30,7 +30,7 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
     async send(task: Task): Promise<void> {
         const pushConfigs = await this.pushNotificationStore.load(task.id);
         if (!pushConfigs || pushConfigs.length === 0) {
-            return Promise.resolve();
+            return;
         }
 
         pushConfigs.forEach(pushConfig => {
@@ -77,6 +77,5 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
         } finally {
             clearTimeout(timeoutId);
         }
-        return Promise.resolve();
     }
 }
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -16,19 +16,17 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             return;
         }
 
-        const promises = pushConfigs.map(pushConfig => 
+        pushConfigs.forEach(pushConfig => {
             this._dispatchNotification(task, pushConfig)
-        );
-        
-        const results = await Promise.allSettled(promises);
-        
-        const failedCount = results.filter(result => 
-            result.status === 'rejected' || (result.status === 'fulfilled' && !result.value)
-        ).length;
-        
-        if (failedCount > 0) {
-            console.warn(`Some push notifications failed to send for task_id=${task.id}. ${failedCount}/${results.length} failed.`);
-        }
+                .then(success => {
+                    if (!success) {
+                        console.warn(`Push notification failed to send for task_id=${task.id} to URL: ${pushConfig.url}`);
+                    }
+                })
+                .catch(error => {
+                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
+                });
+        });
     }
 
     private async _dispatchNotification(
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -1,0 +1,65 @@
+import { Task, PushNotificationConfig } from "../../types.js";
+import { PushNotificationSender } from "./push_notification_sender.js";
+import { PushNotificationStore } from "./push_notification_store.js";
+
+export class DefaultPushNotificationSender implements PushNotificationSender {
+
+    private readonly pushNotificationStore: PushNotificationStore;
+    
+    constructor(pushNotificationStore: PushNotificationStore) {
+        this.pushNotificationStore = pushNotificationStore;
+    }
+
+    async send(task: Task): Promise<void> {
+        const pushConfigs = await this.pushNotificationStore.load(task.id);
+        if (!pushConfigs || pushConfigs.length === 0) {
+            return;
+        }
+
+        const promises = pushConfigs.map(pushConfig => 
+            this._dispatchNotification(task, pushConfig)
+        );
+        
+        const results = await Promise.allSettled(promises);
+        
+        const failedCount = results.filter(result => 
+            result.status === 'rejected' || (result.status === 'fulfilled' && !result.value)
+        ).length;
+        
+        if (failedCount > 0) {
+            console.warn(`Some push notifications failed to send for task_id=${task.id}. ${failedCount}/${results.length} failed.`);
+        }
+    }
+
+    private async _dispatchNotification(
+        task: Task, 
+        pushConfig: PushNotificationConfig
+    ): Promise<boolean> {
+        const url = pushConfig.url;
+        try {
+            const headers: Record<string, string> = {
+                'Content-Type': 'application/json'
+            };
+            
+            if (pushConfig.token) {
+                headers['X-A2A-Notification-Token'] = pushConfig.token;
+            }
+
+            const response = await fetch(url, {
+                method: 'POST',
+                headers,
+                body: JSON.stringify(task)
+            });
+
+            if (!response.ok) {
+                throw new Error(`HTTP ${response.status}: ${response.statusText}`);
+            }
+
+            console.info(`Push notification sent for task_id=${task.id} to URL: ${url}`);
+            return true;
+        } catch (error) {
+            console.error(`Error sending push notification for task_id=${task.id} to URL: ${url}. Error:`, error);
+            return false;
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,65 @@
+import { Task, PushNotificationConfig } from "../../types.js";
+import { PushNotificationSender } from "./push_notification_sender.js";
+import { PushNotificationStore } from "./push_notification_store.js";
+
+export class DefaultPushNotificationSender implements PushNotificationSender {
```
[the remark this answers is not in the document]
botraunak: console logging is the standard being used in `DefaultRequestHandler`

src/server/push_notification/push_notification_store.ts
```diff
@@ -0,0 +1,58 @@
...
+        // Remove existing config with the same ID if it exists
+        const existingIndex = configs.findIndex(config => config.id === pushNotificationConfig.id);
+        if (existingIndex !== -1) {
+            configs.splice(existingIndex, 1);
+        }
+        
+        // Add the new/updated config
+        configs.push(pushNotificationConfig);
+        this.store.set(taskId, configs);
+    }
```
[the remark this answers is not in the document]
botraunak: This is the backwards compatible logic that was inferred from the inline push notification config store

src/server/push_notification/default_push_notification_sender.ts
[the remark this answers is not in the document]
swapydapy: This is a good idea, but can be added in future.

src/server/request_handler/default_request_handler.ts
```diff
@@ -19,22 +22,26 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+    private readonly pushNotificationStore: PushNotificationStore;
+    private readonly pushNotificationSender: PushNotificationSender;
 
 
     constructor(
         agentCard: AgentCard,
         taskStore: TaskStore,
         agentExecutor: AgentExecutor,
         eventBusManager: ExecutionEventBusManager = new DefaultExecutionEventBusManager(),
+        pushNotificationStore: PushNotificationStore = new InMemoryPushNotificationStore(),
```
swapydapy: Both store & sender should be optional. Constructor should check if pushNotification capability is enabled. Based on that, it should check the store & sender

test/server/default_request_handler.spec.ts
```diff
@@ -87,6 +91,8 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
     // Before each test, reset the components to a clean state
     beforeEach(() => {
         mockTaskStore = new InMemoryTaskStore();
+        mockPushNotificationStore = new InMemoryPushNotificationStore();
```
swapydapy: Lets not instantiate these for each testcase. Only create these for the push-notification related testcases.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,72 @@
...
+        const pushConfigs = await this.pushNotificationStore.load(task.id);
+        if (!pushConfigs || pushConfigs.length === 0) {
+            return;
+        }
+
+        const promises = pushConfigs.map(pushConfig => 
+            this._dispatchNotification(task, pushConfig)
+        );
+        
+        const results = await Promise.allSettled(promises);
```
swapydapy: We should not await the push notifications. This would block the processing of executor events. Push notificationss should be sent in the background.

src/server/request_handler/default_request_handler.ts
```diff
@@ -512,4 +514,20 @@ export class DefaultRequestHandler implements A2ARequestHandler {
             eventQueue.stop();
         }
     }
+
+    private async _sendPushNotificationIfNeeded(event: AgentExecutionEvent): Promise<void> {
```
swapydapy, line 523: We should also check if the agent supports pushNotifications.

src/server/request_handler/default_request_handler.ts
```diff
@@ -512,4 +514,20 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+        if (event.kind === 'status-update') {
```
swapydapy: Currently Push notifs should be sent for all events.

swapydapy (changes requested):

src/server/request_handler/default_request_handler.ts
botraunak: Added the check from agentCard

src/server/request_handler/default_request_handler.ts
botraunak: Sending push notifications for all events

src/server/push_notification/default_push_notification_sender.ts
botraunak: You're right although this should be called in a fire and forget manner from the _sendPushNotificationsIfNeeded method from the DefaultRequestHandler, I have updated both the default implementation and sending to be in background.

botraunak:
@swapydapy I have updated the pull request with the necessary changes please review.

test/server/default_request_handler.spec.ts
botraunak: Resolved in the test cases, instantiated only for new push notification sending test case.

src/server/request_handler/default_request_handler.ts
botraunak: Made changes on constructor based on the agent capabilities

test/server/default_request_handler.spec.ts
```diff
@@ -571,6 +579,66 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+        });
+
+        const result = await handler.sendMessage(params);
+        const taskResult = result as Task;
+
+        assert.equal(taskResult.id, taskResult.id);
+        assert.equal(taskResult.status.state, "completed");
+
+        assert.isTrue((mockPushNotificationSender as MockPushNotificationSender).send.calledThrice);
+        assert.equal((mockPushNotificationSender as MockPushNotificationSender).send.firstCall.args[0].id, taskResult.id);
```
swapydapy: We should assert entire task object for all three calls, and not just the taskId.

test/server/default_request_handler.spec.ts
```diff
@@ -571,6 +579,66 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+        assert.equal((mockPushNotificationSender as MockPushNotificationSender).send.secondCall.args[0].id, taskResult.id);
+        assert.equal((mockPushNotificationSender as MockPushNotificationSender).send.thirdCall.args[0].id, taskResult.id);
```
swapydapy: We should also test whether the push notif was actually sent or not. Can we start a server in this test and listen to push events, and verify the payloads. That would test the actual functionality, currently we just check whether the function was called or not.

swapydapy:
Thank you. Requested one additional check for the tests.

test/server/default_request_handler.spec.ts
botraunak: Those tests should not belong in the default_request_handler spec as they shouldn't affect the core functionality of request handler, mocking and simply verifying the calls should be enough. we should create a different tests if we want to add tests for the default_notification_sender implementation. What are your thoughts? @swapydapy

swapydapy: Yes, we can do that too. We should verify that the notifs a re actually sent.

test/server/default_request_handler.spec.ts
```diff
@@ -571,6 +579,66 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+                taskId: ctx.taskId,
+                contextId,
+                kind: 'status-update',
+                status: { state: "completed" },
+                final: true
+            });
+            bus.finished();
+        });
+
+        const result = await handler.sendMessage(params);
```
swapydapy: Please add a testcase for sendMessageStream as well.

swapydapy:
@botraunak any update.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,70 @@
...
+                .catch(error => {
+                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
+                });
+        });
+    }
+
+    private async _dispatchNotification(
+        task: Task, 
+        pushConfig: PushNotificationConfig
+    ): Promise<boolean> {
```
swapydapy: There's no need to return a boolean. Instead, lets convey that through the promise itself. If dispatch is success, the promise is resolved otherwise rejected.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,70 @@
...
+        const url = pushConfig.url;
+        const controller = new AbortController();
+        // Abort the request if it takes longer than 5 seconds.
+        const timeoutId = setTimeout(() => controller.abort(), 5000);
```
swapydapy: Thanks for adding a timeout. Instead of hard-coded value of 5s, can we make it a configurable param of `DefaultPushNotificationSender` with default value of 5s.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        try {
+            const headers: Record<string, string> = {
+                'Content-Type': 'application/json'
+            };
+            
+            if (pushConfig.token) {
+                headers['X-A2A-Notification-Token'] = pushConfig.token;
```
swapydapy: Can we also make this header name configurable for `DefaultPushNotificationSender`.

test/server/default_request_handler.spec.ts
```diff
@@ -571,6 +579,178 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+        assert.equal((events[1] as TaskStatusUpdateEvent).status.state, "working");
+        assert.equal((events[2] as TaskStatusUpdateEvent).status.state, "completed");
+        assert.isTrue((events[2] as TaskStatusUpdateEvent).final);
+
+        // Verify push notifications were sent with complete task objects
+        assert.isTrue((mockPushNotificationSender as MockPushNotificationSender).send.calledThrice);
+        
+        // Verify first call (submitted state)
+        const firstCallTask = (mockPushNotificationSender as MockPushNotificationSender).send.firstCall.args[0] as Task;
+        assert.equal(firstCallTask.id, (events[0] as Task).id);
```
swapydapy: Instead of verifying individual properties, can we rather construct the expected Task payload and use `assert.deepEqual`

That makes reading the test much easier and verifies entire payload. This way all properties are verified and tests also detect unknown props. 

Can we do this for all three payloads & same for sendMessageStream push notification testcase.

test/server/push_notification_integration.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+                    body: req.body,
+                    headers: req.headers,
+                    url: req.url,
+                    method: req.method
+                });
+                res.status(200).json({ received: true, timestamp: new Date().toISOString() });
+            });
+
+            // Endpoint to simulate different response scenarios
+            app.post('/notify/:scenario', (req: Request, res: Response) => {
```
swapydapy, line 61: Lets just keep the `/notify/:scenario` to allow for multiple notifUrls to be added for single Task.

Lets remove the testing for slow, instead just keep the `error` scenario. As that would check the push Notif sender actually handles/ignores API errors.

test/server/push_notification_integration.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+
+            const params: MessageSendParams = {
+                message: createTestMessage('Test task with push notifications'),
+                configuration: {
+                    pushNotificationConfig: pushConfig
+                }
+            };
+
+            // Mock the agent executor to publish all three states for this test only
+            mockAgentExecutor.execute.callsFake(async (ctx, bus) => {
```
swapydapy, line 152: This same fake has been used for each testcase, can we refactor out it as constant and supply in all testcases.

test/server/push_notification_integration.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+            assert.include(states, 'working', 'Should include working state');
+            assert.include(states, 'completed', 'Should include completed state');
+            
+            // Verify first notification has correct format
+            const firstNotification = receivedNotifications[0];
+            assert.equal(firstNotification.method, 'POST');
+            assert.equal(firstNotification.url, '/notify');
+            assert.equal(firstNotification.headers['content-type'], 'application/json');
+            assert.equal(firstNotification.headers['x-a2a-notification-token'], 'test-auth-token');
+            assert.equal(firstNotification.body.id, task.id);
```
swapydapy: can we assert.deepEqual entire task payload for all 3 notifs.

test/server/push_notification_integration.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+            // Wait for async push notifications to be sent
+            await new Promise(resolve => setTimeout(resolve, 300));
+
+            // Should now have notifications from both endpoints
+            const allEndpoints = receivedNotifications.map(n => n.url);
+            assert.include(allEndpoints, '/notify', 'Should notify primary endpoint');
+            assert.include(allEndpoints, '/notify/second', 'Should notify second endpoint');
+        });
+
+        it('should handle slow push notification endpoints gracefully', async () => {
```
swapydapy: Good testcase, though instead of slow, can we just check for `error` scenario. That validates that pushSender can ignore API failures and doesn't fail itself.

test/server/push_notification_integration.spec.ts
botraunak: Removed the slow scenario

test/server/push_notification_integration.spec.ts
botraunak: Refactored the mocks

test/server/push_notification_integration.spec.ts
botraunak: added deepequal checks for body

test/server/push_notification_integration.spec.ts
botraunak: Removed the slow testcase scenario

test/server/push_notification_integration.spec.ts
swapydapy: There's also the 'unauthorized' scenario as well, could you remove that too.

test/server/default_request_handler.spec.ts
```diff
@@ -571,6 +516,181 @@ describe('DefaultRequestHandler as A2ARequestHandler', () => {
...
+                kind: 'status-update',
+                status: { state: "completed" },
+                final: true
+            });
+            bus.finished();
+        });
+
+        const result = await handler.sendMessage(params);
+        const taskResult = result as Task;
+        const task = await mockTaskStore.load(taskResult.id);
```
swapydapy: Instead of loading it from taskStore, could you keep a copy of Task object supplied to the MockAgentExecutor. This way the test assertion doesn't not depend on the task stored in mockStore. As that can lead of to bugs. If there's a bug in taskStore, it won't be caught.

Instead we should treat the SDK as blackbox. Suuply the same copy of task to `MockAgentExecutor` and assert that same is being returned as push notification.

swapydapy: Please do the same for sendMessageStream testcase.

test/server/push_notification_integration.spec.ts
swapydapy: Actually I meant the mocks in `mockAgentExecutor.execute.callsFake`

The same "submitted", "working" & "completed" are used across multiple testcases in `push_notification_integration.spec.ts` & `default_request_hanlder`.

Can we refactor this fakeFunction as constant and re-use it.

const fakeExecute = async (ctx, bus) => {}

test/server/push_notification_integration.spec.ts
```diff
@@ -0,0 +1,647 @@
...
+
+            // Send a message to trigger notifications
+            await handler.sendMessage(params);
+
+            // Wait for async push notifications to be sent
+            await new Promise(resolve => setTimeout(resolve, 300));
+
+            // Should now have notifications from both endpoints
+            const allEndpoints = receivedNotifications.map(n => n.url);
+            assert.include(allEndpoints, '/notify', 'Should notify primary endpoint');
```
swapydapy: can we check the counts as well

swapydapy (changes requested):

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            console.info(`Push notification sent for task_id=${task.id} to URL: ${url}`);
+        } catch (error) {
+            // Ignore errors
+            console.error(`Error sending push notification for task_id=${task.id} to URL: ${url}. Error:`, error);
+        } finally {
+            clearTimeout(timeoutId);
+        }
+        return Promise.resolve();
```
swapydapy: no need for this `return Promise.resolve();`

swapydapy:
The new tests make this very robust. Just final nits to polish the testcases, and we can merge this.

test/server/default_request_handler.spec.ts
botraunak: Task id is generated by the implementation of the requestHandler Request Context, I cannot create a copy of the task object as I would not know what taskId will be assigned, so I will anyways figure that out through the task store. So creating an reference object is moot as the dependency is still there on the Task store. We can supply the taskId and contextId when sending a message but we will still need to save it in taskStore as the requestContext building requires this. 
So these tests shouldn't really be concerned with the correctness of TaskStore implementation, we are already testing for those in the defaultRequestHandler specs get task method. So if any errors in future if shall arise in the task store implementation should be caught through those tests.

test/server/push_notification_integration.spec.ts
botraunak: done.

src/server/push_notification/default_push_notification_sender.ts
botraunak: done

botraunak:
@swapydapy Made the relevant changes, please help resolve the review comment for task store dependency removal, Would like to get it merged.

src/server/push_notification/default_push_notification_sender.ts
swapydapy: seems it got missed. Could we remove this `return Promise.resolve();`

test/server/default_request_handler.spec.ts
swapydapy: During testing a component/system, we ideally assert the expected final behaviours. We avoid interacting with its internal dependencies. The reason for that is that the contract between the internal dependencies can change without affecting the final behaviours of the system under test.

Lets say the taskStore contract totally changes by storing artifacts separately in another table, and only keeping reference with the task. The taskstore interface will be changed and the associated tests. Everything is fine on that front. Now there's a bug in push_notification which is not sending full artifacts and only the artifact reference, this current test will still succeed. As it is loading Task definition from the taskStore. If we had asserted final push payload based on our expectation of task, the test would fail, and thereby catch the bug. Hence, always better to assert on final behaviour without involving internal implementation details.

I see your point regarding taskId not known beforehand. That can be resolved by two ways:
1. We are supplying the fakeTaskExecute. We can record what was the taskID supplied and associated task Objects which were emitted. Use that to create expectations. Here we are using internal details, but our own supplied executor.
2. The task objects, without the taskID, in fakeTaskExecute can be exposed as constants. Use that to assert the push payload but ignore the taskID. Though this is vulnerable to incorrect taskID being passed to push notifs.

swapydapy (changes requested):

swapydapy (approved):

PR #115, opened 2025-08-18
chore: repin express to 5.1 and express types to 5.0.3

adamfweidman:
repin express to ^5.1.0 and express types to ^5.0.3

This is necessary in order to depend on this package in gemini-cli which already has a dependency on 5.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "node": ">=18"
       },
       "peerDependencies": {
-        "express": "^5.1.0"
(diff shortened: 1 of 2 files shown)


swapydapy:
Since the setup routes method compatible with both express version 4 & 5, let's add support for both versions of express like this.

Also please test this locally, by importing into another sample project to assert this works.

```
{
  "peerDependencies": {
    "express": "^4.21.2 || ^5.0.1"
  }
}
```

PR #117, opened 2025-08-19
fix: Improve fetch implementation handling

### LINKED ISSUE #116: [Bug]: Illegal invocation when fetchImpl is not provided (due to saved native fetch reference)
### What happened?

When `fetchImpl` is not explicitly provided to `A2AClient`, an Illegal invocation error occurs. It appears this happens because the saved reference to fetch loses its window context.
```
Error fetching or parsing Agent Card: TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
```

### Suggested Fix
 
Ensure the fetch reference retains the correct context. Two possible approaches:
1. Bind window when saving the fetch reference in the A2AClient constructor:
```
this.fetchImpl = options?.fetchImpl ??
      (typeof window !== "undefined" ? window.fetch.bind(window) : fetch);
```
2. Avoid saving fetch as a reference; instead, call it dynamically when needed, if custom fetch is not provided.

### Code of Conduct

ajaynagotha:
# Description
This change improves flexibility for different environments and error handling when no fetch implementation is available.

Fixes #116  🦕

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -387,7 +387,7 @@ export class A2AClient {
       body: JSON.stringify(rpcRequest),
     });
 
- if (!response.ok) {
+  if (!response.ok) {
       let errorBody = "";
       try {
         errorBody = await response.text();
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -387,7 +387,7 @@ export class A2AClient {
       body: JSON.stringify(rpcRequest),
     });
 
-  if (!response.ok) {
+    if (!response.ok) {
       let errorBody = "";
       try {
         errorBody = await response.text();
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -48,9 +48,9 @@ export class A2AClient {
   private agentCardPromise: Promise<AgentCard>;
   private requestIdCounter: number = 1;
   private serviceEndpointUrl?: string; // To be populated from AgentCard after fetching
-  private fetchImpl: typeof fetch;
+  private customFetchImpl?: typeof fetch;
 
-  /**
+ /**
    * Constructs an A2AClient instance.
    * It initiates fetching the agent card from the provided agent baseUrl.
    * The Agent Card is fetched from a path relative to the agentBaseUrl, which defaults to '.well-known/agent-card.json'.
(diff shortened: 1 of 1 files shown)


src/client/client.ts
```diff
@@ -59,10 +59,34 @@ export class A2AClient {
...
+   * @returns The fetch implementation.
+   * @param args Arguments to pass to the fetch implementation.
+   * @throws If no fetch implementation is available.
+   */
+  private _fetch(...args: Parameters<typeof fetch>): ReturnType<typeof fetch> {
+    if (this.customFetchImpl) {
+      return this.customFetchImpl(...args);
+    }
+    if (typeof window !== 'undefined' && typeof window.fetch === 'function') {
+      return window.fetch.bind(window)(...args);
```
swapydapy: Why are we binding the window object?

src/client/client.ts
```diff
@@ -59,10 +59,34 @@ export class A2AClient {
...
+    }
+    if (typeof fetch === 'function') {
```
swapydapy, line 83: why is this check needed?

ajaynagotha: > why is this check needed?

In case someone is using library in environment where fetch is not globally exposed

src/client/client.ts
ajaynagotha: We are making sure that fetch's context to window always should be there. Making sure it is not detached and if it is then we are re adding it before using.

ajaynagotha: We can remove every condition and we can just check which says typeof fetch should be function

ajaynagotha: it is just that currently we are putting extra security in place

src/client/client.ts
ajaynagotha: We are making sure it does not throw invocation or type of undefined error

ajaynagotha: We are giving clear message so that they can add fetch using some library

src/client/client.ts
swapydapy: This binding is not necessary right since now we are directly calling the `fetch` method. That `fetch` is already in global scope.

src/client/client.ts
```diff
@@ -59,10 +59,34 @@ export class A2AClient {
...
+      return fetch(...args);
+    }
+    throw new Error(
+      'A `fetch` implementation was not provided and is not available in the global scope. ' +
+      'Please provide a `fetchImpl` in the A2AClientOptions. ' +
+      'For Node.js environments, you can use a library like `node-fetch`.'
```
swapydapy: Node-18 already has native fetch. So rephrase this message as "for earlier Node.js environments"

src/client/client.ts
swapydapy: This entire `if` condition is not needed.

src/client/client.ts
swapydapy: Please do proper due-diligence of the generated code from LLMs.

src/client/client.ts
```diff
@@ -59,10 +59,34 @@ export class A2AClient {
...
+    if (this.customFetchImpl) {
+      return this.customFetchImpl(...args);
+    }
+    if (typeof window !== 'undefined' && typeof window.fetch === 'function') {
+      return window.fetch.bind(window)(...args);
+    }
+    if (typeof fetch === 'function') {
+      return fetch(...args);
+    }
+    throw new Error(
```
swapydapy, line 86: Please add a testcase for this error scenario.

ajaynagotha: Sure

src/client/client.ts
ajaynagotha: Sure.

src/client/client.ts
ajaynagotha: This is done

src/client/client.ts
ajaynagotha: This is done

src/client/client.ts
ajaynagotha: This is done

test/client/client.spec.ts
```diff
@@ -52,6 +52,33 @@ describe('A2AClient Basic Tests', () => {
...
+        global.fetch = undefined;
+
+        // Act: Instantiate the client without providing a custom fetch implementation.
+        // The constructor kicks off an async operation that will fail.
+        const clientWithoutFetch = new A2AClient('https://test-agent.example.com');
+        // Assert: Check that any method relying on the agent card fetch rejects with the expected error.
+        await clientWithoutFetch.getAgentCard();
+        expect.fail('Expected an error to be thrown but it was not.');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Error);
```
swapydapy: Instead of putting an expect inside the catch block, capture that error in a variable. Then assert that variable after the try/catch block. This way the asserts will always run. That captures the intent of the testcase, much better.

In current code, if for some reason the catch block is not called, the expectation will not be run. The added `expect.fail` is just a stop gap solution..

ajaynagotha: Sure I will fix this by today

ajaynagotha: This is fixed as requested

ajaynagotha:
> Please update to latest branch

This is also done and conflicts are also resolved

ajaynagotha:
@swapydapy I have made changes as requested pls check and re review. Thanks for all the guidance

swapydapy (approved):

PR #121, opened 2025-08-25
feat!: Update A2AClient constructor to initialize with full agentCardUrl

adamfweidman:
# Description

- Breaking Change: A2AClient Constructor initialized with full agentCard rather than `baseUrl` and `agent_card_path`. This previously led to an error where the baseUrl had to be the baseUrl of the full agent card, rather than the serviceUrl (which is set by the AgentCard itself). 
- Static Method to load A2AClient from agentCardUrl.
- Cleanup trailing spaces at the start of new lines

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -43,112 +43,64 @@ export interface A2AClientOptions {
  * A2AClient is a TypeScript HTTP client for interacting with A2A-compliant agents.
  */
 export class A2AClient {
-  private agentCardPromise: Promise<AgentCard>;
+  private agentCard: AgentCard;
   private requestIdCounter: number = 1;
-  private serviceEndpointUrl?: string; // To be populated from AgentCard after fetching
+  private serviceEndpointUrl: string;
   private fetchImpl: typeof fetch;
 
   /**
-   * Constructs an A2AClient instance.
-   * It initiates fetching the agent card from the provided agent card URL.
-   * The `url` field from the Agent Card will be used as the RPC service endpoint.
-   * @param agentCardUrl The URL of the agent card (e.g., https://agent.example.com/.well-known/agent-card.json)
-   * @param options Optional. The options for the A2AClient including the fetch implementation and authentication handler.
-   */
-  constructor(agentCardUrl: string, options?: A2AClientOptions) {
-    this.fetchImpl = options?.fetchImpl ?? fetch;
-    this.agentCardPromise = this._fetchAndCacheAgentCard(agentCardUrl);
-  }
-
-  /**
-   * Creates an A2AClient instance from an already-fetched AgentCard.
-   * This method is useful when the AgentCard is obtained through means other than
-   * the client's own fetching mechanism (e.g., from a discovery service).
+   * Constructs an A2AClient instance from an AgentCard.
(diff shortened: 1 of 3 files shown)


src/client/client.ts
```diff
@@ -52,27 +50,49 @@ export class A2AClient {
...
+    }
+
+    // Create an instance without calling the constructor to avoid initial fetch
+    const client = Object.create(A2AClient.prototype) as A2AClient;
+
+    // Set necessary properties
+    client.fetchImpl = options?.fetchImpl ?? fetch;
+    client.agentCardPromise = Promise.resolve(agentCard);
+    client.serviceEndpointUrl = agentCard.url;
+    client.requestIdCounter = 1;
```
swapydapy: requestIdCounter??

src/client/client.ts
```diff
@@ -52,27 +50,49 @@ export class A2AClient {
...
+   * @param options Optional. The options for the A2AClient including the fetch implementation.
+   * @returns A new A2AClient instance.
+   */
+  public static fromAgentCard(agentCard: AgentCard, options?: A2AClientOptions): A2AClient {
+    if (!agentCard.url) {
+      throw new Error("Provided Agent Card does not contain a valid 'url' for the service endpoint.");
+    }
+
+    // Create an instance without calling the constructor to avoid initial fetch
+    const client = Object.create(A2AClient.prototype) as A2AClient;
```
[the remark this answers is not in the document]
swapydapy: +1 Let's avoid this approach, instead modify the constructor itself.

1. It should take in the agentCard.
2. Let's avoid this _fetchAndCacheAgentCard(). Just expect a resolved agentCard in constructor and validate it there.
3. Lets add a static method `fromCardUrl()`, which takes in a full agent card URL.

src/client/client.ts
```diff
@@ -31,116 +31,76 @@ import {
...
-   * The Agent Card is fetched from a path relative to the agentBaseUrl, which defaults to '.well-known/agent-card.json'.
-   * The `url` field from the Agent Card will be used as the RPC service endpoint.
-   * @param agentBaseUrl The base URL of the A2A agent (e.g., https://agent.example.com)
-   * @param options Optional. The options for the A2AClient including the fetch implementation, agent card path, and authentication handler.
+   * Constructs an A2AClient instance from an AgentCard.
+   * @param agentCard The AgentCard object.
+   * @param options Optional. The options for the A2AClient including the fetch/auth implementation.
    */
-  constructor(agentBaseUrl: string, options?: A2AClientOptions) {
+  constructor(agentCard: AgentCard, options?: A2AClientOptions) {
```
swapydapy: Just realised, for backward compatibility, could you check if `agentCard` is instance if string. If yes, print out a warning to instead use `fromCardUrl` method, and call the fromCardUrl method. We can remove this backward compatibility support in future.

Please a test for this support too. Thank you!

adamfweidman: Done

swapydapy (approved):

ajaynagotha:
@swapydapy I wonder why readme is not updated for this breaking change

ajaynagotha:
@adamfweidman I think you should update readme for this change

PR #141, opened 2025-10-24
fix: ensure push notifications order

### LINKED ISSUE #140: [Bug]: SendPushNotification does not implement any check to avoid race conditions
### What happened?

SendPushNotification does not guarantee that notifications will be sent in the correct order, which can lead to a race condition. This is already happening, as the RunUnitTest action appears to be flaky.

- https://github.com/a2aproject/a2a-js/actions/runs/18750636883 (see top right corner for first two failed attempts)
- https://github.com/a2aproject/a2a-js/actions/runs/17686476400

### Relevant log output

```shell

```

### Code of Conduct

guglielmo-san:
# Description

This PR introduces a promise chain (`this.notificationChain`) to serialize the batches of notifications. This ensures that only one batch at a time is processed, in a predictable First-In, First-Out (FIFO) order. The test was already ensuring it and it caused some flakiness before, examples:
- https://github.com/a2aproject/a2a-js/actions/runs/18750636883 (see top right corner for first two failed attempts)
- https://github.com/a2aproject/a2a-js/actions/runs/17686476400

The `push_notification_integration.spec.ts` test file has been modified to simulate a more deterministic delay (100ms) on the completion of the first received notification. This guarantee a 100% test failure when the notification serialization is not implemented.

Fixes #140  🦕

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,28 @@ const client = await A2AClient.fromCardUrl(
 await client.sendMessage({ message: { messageId: uuidv4(), role: "user", parts: [{ kind: "text", text: "A message requiring custom headers." }], kind: "message" } });
 ```
 
+### Example: Specifying a Timeout
+
+This example creates a `fetch` wrapper that sets a timeout for every outgoing request.
+
+```typescript
+import { A2AClient } from "@a2a-js/sdk/client";
+
+// 1. Create a wrapper around the global fetch function.
+const fetchWithTimeout: typeof fetch = async (url, init) => {
+  return fetch(url, { ...init, signal: AbortSignal.timeout(5000)});
+};
+
+// 2. Provide the custom fetch implementation to the client.
+const client = await A2AClient.fromCardUrl(
+  "http://localhost:4000/.well-known/agent-card.json",
+  { fetchImpl: fetchWithTimeout }
+);
+
+// Now, all requests made by this client instance will have a configured timeout.
+await client.sendMessage({ message: { messageId: uuidv4(), role: "user", parts: [{ kind: "text", text: "A message requiring custom headers." }], kind: "message" } });
+```
+
 ### Using the Provided `AuthenticationHandler`
 
 For advanced authentication scenarios, the SDK includes a higher-order function `createAuthenticatingFetchWithRetry` and an `AuthenticationHandler` interface. This utility automatically adds authorization headers and can retry requests that fail with authentication errors (e.g., 401 Unauthorized).
--- a/package.json
+++ b/package.json
@@ -68,8 +68,9 @@
     "test": "mocha test/**/*.spec.ts",
     "coverage": "c8 npm run test",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",
-    "sample:cli": "tsx src/samples/cli.ts",
-    "sample:movie-agent": "tsx src/samples/agents/movie-agent/index.ts"
+    "a2a:cli": "tsx src/samples/cli.ts",
+    "agents:movie-agent": "tsx src/samples/agents/movie-agent/index.ts",
+    "agents:sample-agent": "tsx src/samples/agents/sample-agent/index.ts"
   },
   "dependencies": {
     "uuid": "^11.1.0"
--- a/src/samples/agents/movie-agent/index.ts
+++ b/src/samples/agents/movie-agent/index.ts
@@ -312,7 +312,10 @@ async function main() {
 
   // 5. Start the server
   const PORT = process.env.PORT || 41241;
-  expressApp.listen(PORT, () => {
+  expressApp.listen(PORT, (err) => {
+    if (err) {
+      throw err;
+    }
     console.log(`[MovieAgent] Server using new framework started on http://localhost:${PORT}`);
     console.log(`[MovieAgent] Agent Card: http://localhost:${PORT}/.well-known/agent-card.json`);
     console.log('[MovieAgent] Press Ctrl+C to stop the server');
--- a/src/samples/agents/sample-agent/README.md
+++ b/src/samples/agents/sample-agent/README.md
@@ -1,0 +1,9 @@
+# Sample Agent
+
+This agent is a sample to emulate the task flow in a streaming scenario. To run:
+
+```bash
+npm run agents:sample-agent
+```
+
+The agent will start on `http://localhost:41241`.
--- a/src/samples/agents/sample-agent/index.ts
+++ b/src/samples/agents/sample-agent/index.ts
@@ -1,0 +1,201 @@
+import express from "express";
+import { v4 as uuidv4 } from 'uuid'; // For generating unique IDs
+
+import {
+  AgentCard,
+  Task,
+  TaskStatusUpdateEvent,
+  Message
+} from "../../../index.js";
+import {
+  InMemoryTaskStore,
+  TaskStore,
+  AgentExecutor,
+  RequestContext,
+  ExecutionEventBus,
+  DefaultRequestHandler
+} from "../../../server/index.js";
+import { A2AExpressApp } from "../../../server/express/index.js";
+
+/**
+ * SampleAgentExecutor implements the agent's core logic.
+ */
+class SampleAgentExecutor implements AgentExecutor {
+
+  public cancelTask = async (
+    taskId: string,
+    eventBus: ExecutionEventBus,
+  ): Promise<void> => { };
+
+  async execute(
+    requestContext: RequestContext,
+    eventBus: ExecutionEventBus
+  ): Promise<void> {
+    const userMessage = requestContext.userMessage;
+    const existingTask = requestContext.task;
+
+    // Determine IDs for the task and context
+    const taskId = requestContext.taskId;
+    const contextId = requestContext.contextId;
+
+    console.log(
+      `[SampleAgentExecutor] Processing message ${userMessage.messageId} for task ${taskId} (context: ${contextId})`
+    );
+
+    // 1. Publish initial Task event if it's a new task
+    if (!existingTask) {
+      const initialTask: Task = {
+        kind: 'task',
+        id: taskId,
+        contextId: contextId,
+        status: {
+          state: 'submitted',
+          timestamp: new Date().toISOString(),
+        },
+        history: [userMessage], // Start history with the current user message
+        metadata: userMessage.metadata, // Carry over metadata from message if any
+      };
+      eventBus.publish(initialTask);
+    }
+
+    // 2. Publish "working" status update
+    const workingStatusUpdate: TaskStatusUpdateEvent = {
+      kind: 'status-update',
+      taskId: taskId,
+      contextId: contextId,
+      status: {
+        state: 'working',
+        message: {
+          kind: 'message',
+          role: 'agent',
+          messageId: uuidv4(),
+          parts: [{ kind: 'text', text: 'Processing your question' }],
+          taskId: taskId,
+          contextId: contextId,
+        },
+        timestamp: new Date().toISOString(),
+      },
+      final: false,
+    };
+    eventBus.publish(workingStatusUpdate);
+
+    // 3. Publish final task status update
+    const agentReplyText = this.parseInputMessage(userMessage);
+    console.info(`[SampleAgentExecutor] Prompt response: ${agentReplyText}`);
+ 
+    const agentMessage: Message = {
+      kind: 'message',
+      role: 'agent',
+      messageId: uuidv4(),
+      parts: [{ kind: 'text', text: agentReplyText }],
+      taskId: taskId,
+      contextId: contextId,
+    };
+
+    const finalUpdate: TaskStatusUpdateEvent = {
+      kind: 'status-update',
+      taskId: taskId,
+      contextId: contextId,
+      status: {
+        state: 'completed',
+        message: agentMessage,
+        timestamp: new Date().toISOString(),
+      },
+      final: true,
+    };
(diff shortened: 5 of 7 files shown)


src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -33,11 +34,10 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
-        pushConfigs.forEach(pushConfig => {
-            this._dispatchNotification(task, pushConfig)
-                .catch(error => {
-                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
-                });
+        // Ensures that notification are sent sequentially.
+        this.notificationChain = this.notificationChain.then(async () => {
+            const dispatches = pushConfigs.map(pushConfig => this._dispatchNotification(task, pushConfig));
+            await Promise.all(dispatches);
         });
```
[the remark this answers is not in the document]
ishymko: I'd put a catch explicitly on `_dispatchNotification` and remove it from the method itself, it doesn't throw now but it shouldn't be its' responsibility to never throw.

test/server/push_notification_integration.spec.ts
```diff
@@ -47,7 +48,12 @@ describe('Push Notification Integration Tests', () => {
             app.use(express.json());
 
             // Endpoint to receive push notifications
-            app.post('/notify', (req: Request, res: Response) => {
+            app.post('/notify', async (req: Request, res: Response) => {
+                const status = req.body.status.state;
+                // Simulate delay for 'submitted' status to test correct ordering of notifications
+                if (status === 'submitted') {
+                    await new Promise(resolve => setTimeout(resolve, 100))
+                }
```
ishymko: Can this be moved into `/notify/:scenario`? Seems like it's there for cases like this.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -33,11 +34,10 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             return;
         }
 
-        pushConfigs.forEach(pushConfig => {
-            this._dispatchNotification(task, pushConfig)
-                .catch(error => {
-                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
-                });
+        // Ensures that notification are sent sequentially.
```
ishymko: nit: `in the same order they are sent`?

test/server/push_notification_integration.spec.ts
ishymko, line 1: Let's revert formatting changes in this file for now, we should run some repo-wise formatting separately.

src/server/push_notification/default_push_notification_sender.ts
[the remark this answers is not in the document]
guglielmo-san: Done

src/server/push_notification/default_push_notification_sender.ts
guglielmo-san: Done

test/server/push_notification_integration.spec.ts
guglielmo-san: Done

test/server/push_notification_integration.spec.ts
```diff
@@ -58,9 +58,13 @@ describe('Push Notification Integration Tests', () => {
             });
 
             // Endpoint to simulate different response scenarios
-            app.post('/notify/:scenario', (req: Request, res: Response) => {
+            app.post('/notify/:scenario', async(req: Request, res: Response) => {
```
ishymko: nit: formatting `async (`

test/server/push_notification_integration.spec.ts
guglielmo-san: Done

test/server/push_notification_integration.spec.ts
ishymko, line 165: Given that we now added 100ms into notifications sending, should we increase it? Probably worth looking into making it deterministic separately.

test/server/push_notification_integration.spec.ts
```diff
@@ -620,4 +624,4 @@ describe('Push Notification Integration Tests', () => {
             });
         });
     });
-});
+});
```
ishymko: nit: formatting: let's keep new line at the end of file as before

test/server/push_notification_integration.spec.ts
```diff
@@ -58,9 +58,13 @@ describe('Push Notification Integration Tests', () => {
             });
 
             // Endpoint to simulate different response scenarios
-            app.post('/notify/:scenario', (req: Request, res: Response) => {
+            app.post('/notify/:scenario', async(req: Request, res: Response) => {
                 const scenario = req.params.scenario;
-                
+                // Simulate delay for 'submitted' status to test correct ordering of notifications
+                if(scenario === 'delay_on_submitted' && req.body.status.state === 'submitted') {
```
ishymko: nit: formatting `if (`

ishymko (approved):

test/server/push_notification_integration.spec.ts
guglielmo-san: Done

test/server/push_notification_integration.spec.ts
guglielmo-san: Done

test/server/push_notification_integration.spec.ts
guglielmo-san: Done

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -33,11 +34,12 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
             return;
         }
 
-        pushConfigs.forEach(pushConfig => {
-            this._dispatchNotification(task, pushConfig)
-                .catch(error => {
-                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
-                });
+        // Ensures that notification are delivered in the same order they are sent.
+        this.notificationChain = this.notificationChain.then(async () => {
```
swapydapy: Instead of using a promise chain, whose length will keep on increasing as notifs arrive, can we rather use a queue, denoted by a linked list.

As notifs get delivered, they can be removed from the queue. A dispatcher can keep reading from that queue.

Also, having a global queue is going to harm the throughput. This is going to be especially bad, when an A2A server is processing multiple tasks in parallel. Instead a task level queue should be added.

The entire task level notif queue can be cleaned once there are no more notifs to be delivered.

test/server/push_notification_integration.spec.ts
```diff
@@ -58,9 +58,13 @@ describe('Push Notification Integration Tests', () => {
             });
 
             // Endpoint to simulate different response scenarios
-            app.post('/notify/:scenario', (req: Request, res: Response) => {
+            app.post('/notify/:scenario', async (req: Request, res: Response) => {
                 const scenario = req.params.scenario;
-                
+                // Simulate delay for 'submitted' status to test correct ordering of notifications
+                if (scenario === 'delay_on_submitted' && req.body.status.state === 'submitted') {
+                    await new Promise(resolve => setTimeout(resolve, 100));
```
swapydapy: Lets just minimise the delay to be 10ms. We should keep tests super fast.

swapydapy: Also there are other techniques to make testing time related code to be more deterministic, like simulated clock. But I guess can we keep using settimeout for now.

src/server/push_notification/default_push_notification_sender.ts
ishymko: > having a global queue is going to harm the throughput

Very good point, indeed there is no reason to serialize notifications globally, we just need it per task.

> promise chain, whose length will keep on increasing as notifs arrive

I don't think this is the case here actually. As only one the most recent reference is kept, GC will get rid of resolved promises as they run, so the behaviour will be exactly as you described: a queue and JS runtime acting as a dispatcher. A small test:

```js
class Queue {
    constructor() { this.promiseChain = Promise.resolve(); }
    add(fn) { this.promiseChain = this.promiseChain.then(fn) }
}

function logMemory(label) {
    const mem = process.memoryUsage().heapUsed / 1024 / 1024;
    console.log(`${label}: ${mem.toFixed(2)} MB`);
}

const queue = new Queue();
for (let i = 1; i <= 1_000_000; i++) {
    if (i % 50_000 == 0) {
        queue.add(() => new Promise(res => setTimeout(res, 1000)));
        queue.promiseChain.then(() => logMemory(`Iteraion ${i}`))
        await new Promise(res => setTimeout(res, 1000))
    } else {
        queue.add(() => new Promise(res => res()))
    }
}
```

outputs

```
Iteraion 50000: 22.17 MB
Iteraion 100000: 28.20 MB
Iteraion 150000: 13.25 MB
Iteraion 200000: 14.87 MB
Iteraion 250000: 17.85 MB
Iteraion 300000: 17.87 MB
Iteraion 350000: 23.64 MB
Iteraion 400000: 24.92 MB
Iteraion 450000: 13.29 MB
Iteraion 500000: 14.92 MB
Iteraion 550000: 17.80 MB
Iteraion 600000: 17.86 MB
Iteraion 650000: 13.34 MB
Iteraion 700000: 14.98 MB
Iteraion 750000: 17.99 MB
Iteraion 800000: 18.00 MB
Iteraion 850000: 13.34 MB
Iteraion 900000: 14.91 MB
Iteraion 950000: 17.83 MB
Iteraion 1000000: 13.63 MB
```

While if it constantly increased, it'd look like this:

```js
const queue = []; // Just keep all refs to measure memory for 1M promises.
for (let i = 1; i <= 1_000_000; i++) {
    queue.push(() => new Promise(res => res()))
    if (i % 50_000 == 0) {
        logMemory(`Iteration ${i}`)
        await new Promise(res => setTimeout(res, 1000))
    }
}
```

outputs

```
Iteration 50000: 7.13 MB
Iteration 100000: 11.75 MB
Iteration 150000: 14.83 MB
Iteration 200000: 19.47 MB
Iteration 250000: 22.16 MB
Iteration 300000: 27.79 MB
Iteration 350000: 30.47 MB
Iteration 400000: 37.57 MB
Iteration 450000: 40.25 MB
Iteration 500000: 42.98 MB
Iteration 550000: 42.69 MB
Iteration 600000: 52.00 MB
Iteration 650000: 54.68 MB
Iteration 700000: 57.35 MB
Iteration 750000: 60.02 MB
Iteration 800000: 62.68 MB
Iteration 850000: 65.35 MB
Iteration 900000: 77.98 MB
Iteration 950000: 80.65 MB
Iteration 1000000: 83.33 MB
``` 

What do you think?

swapydapy: Hmm... I see, since we re-assign `this.promiseChain` to new then-promise, the earlier promise object is garbage collected. This is subtle and not explicit. Can we add a big comment explaining this, to make sure future readers are aware of this.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -33,11 +35,25 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
+        // Chain promises to ensure notifications for the same task are sent sequentially.
+        // The `finally` block below will clean up the entry from the map once the
+        // promise chain for this task is complete.
+        const newPromise = this.notificationChain.get(task.id)!.then(async () => {
+            const dispatches = pushConfigs.map(pushConfig => this._dispatchNotification(task, pushConfig).catch((error) => { 
+                console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error);
+             }));
+            await Promise.all(dispatches);
+        })
+        this.notificationChain.set(task.id, newPromise);
```
ishymko: How about something like


```suggestion
        const lastPromise = this.notificationChain.get(task.id) ?? Promise.resolve();
        // Chain promises to ensure notifications for the same task are sent sequentially.
        // The `finally` block below will clean up the entry from the map once the
        // promise chain for this task is complete.
        const newPromise = lastPromise.then(async () => {
            const dispatches = pushConfigs.map(async pushConfig => {
                try {
                    await this._dispatchNotification(task, pushConfig)
                } catch (error) {
                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error)
                }
            });
            await Promise.all(dispatches);
        })
        this.notificationChain.set(task.id, newPromise);
```

test/server/push_notification_integration.spec.ts
guglielmo-san: Creted a new issue to keep track of this
https://github.com/a2aproject/a2a-js/issues/148

src/server/push_notification/default_push_notification_sender.ts
guglielmo-san: Sure, looks better

src/server/push_notification/default_push_notification_sender.ts
guglielmo-san: Done

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -33,11 +35,27 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
+            const dispatches = pushConfigs.map(async pushConfig => {
+                try {
+                    await this._dispatchNotification(task, pushConfig)
+                } catch (error) {
+                    console.error(`Error sending push notification for task_id=${task.id} to URL: ${pushConfig.url}. Error:`, error)
+                }
+            });
+            await Promise.all(dispatches);
+        })
+        this.notificationChain.set(task.id, newPromise);
```
swapydapy, line 52: Can you add a comment explaining that previous promise objects will be garbage collected, as they get resolved. This will make sure memory doesn't linearly increase with number of notifs sent to a task.

Issue #137, opened 2025-10-09
[Feat]: Support HTTP+JSON/REST


alirussell:
### Is your feature request related to a problem? Please describe.

Im trying to use the client to connect to a server that only supports HTTP+JSON/REST, but the client doesn't seem to support this at all, and defaults to JSON-RPC.

### Describe the solution you'd like

The client should detect what transports protocols are supported by the agent card and then use the most appropriate one.

### Describe alternatives you've considered

Building my own client directly using the REST endpoints.

### Additional context

I am building a more generic client for our platform, which needs to be able to support all transports.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tchapacan:
I'm planning to open a PR for the client support in the continuation of the server part, commenting here just for visibility and make sure no duplicate work will be done.

PR #142, opened 2025-10-26
feat: implement server http+json

tchapacan:
# Description

## Add HTTP+REST ServerSide Transport Support

- Server: REST endpoints/routes addition by taking advantage of pre-existing express from the codebase. Following this [doc](https://a2a-protocol.org/latest/specification/#356-method-mapping-reference-table) and internally rerouting to each a2a handler method containing the buisness logic

Some part of the code has been generated using generative AI to speed up the development, I have reviewed it carefully, but this should be taken into account into the review. Keep in mind this is my first time contributing to the repo and this is an attempt to help a2a-js support REST (HTTP+JSON) as per describe in the [a2a documentation](https://a2a-protocol.org/latest/specification/#323-httpjsonrest-transport). I'll be happy to refacto it during review process, don't hesitate to challenge it.

---

Re #137 🦕

Release-As: 0.3.6


ishymko:
Hi @tchapacan, as a heads up we've recently merged ESLint and Prettier integration (#183, #185) which caused substantial diffs across the whole codebase. Those are only formatting and "unused imports" kind of changes, so I you should be good with just resolving all the conflicts in favor of your local state and applying `npm run lint:fix` on top of that.

tchapacan:
Hello @ishymko 

Thanks for the heads-up! Will do for the rebase and the linting no problem about that!

I'm able to reproduce the behavior of the tck CI run by using the correct tag 0.3.0.beta3 locally! I missed the details to switch to the correct tag my bad sorry!

I was able to dig a little further and it seems the object message send by the tck cli is not same as the one expected by the server. Using the `0.3.0.beta3` tag it fails, but it seems to have been fixed in `main` branch, so maybe there is just a new release to perform? 

Following the cli arguments you recommended and by performing some adjustment into the agentcard by following more rigorously this part of the [spec](https://a2a-protocol.org/latest/specification/#56-transport-declaration-and-url-relationships), here what I got now :

`./run_tck.py --sut-url URL --category mandatory --transports jsonrpc,rest`

#### Using tag 0.3.0.beta3 (failing):

- JSONRPC: 
```
{
  kind: 'message',
  messageId: 'text-message-id-54f8f97e-994f-4270-8d35-55623ecf6c5f',
  role: 'user',
  parts: [ { kind: 'text', text: 'Hello from TCK!' } ],
  contextId: '16f7759f-7f57-469a-8e1c-54a169300150'
}
```

- REST:
```
{
  messageId: 'text-message-id-394c0fa2-7548-4c9f-ab01-776bc8ec0426',
  role: 'ROLE_USER',
  content: [ { text: 'Hello from TCK!' } ],
  contextId: '8d0ac3ed-1fcf-45f1-920b-671a5a0f936f'
}
```

#### Using main branch (all good):

- JSONRPC:
```
{
  kind: 'message',
  messageId: 'text-message-id-851f3e53-ac53-42ab-877d-5ac943b986e3',
  role: 'user',
  parts: [ { kind: 'text', text: 'Hello from TCK!' } ],
  contextId: '054bbefe-7e20-4abb-9974-de00f29f795d'
}
```

- REST:
```
{
  messageId: 'text-message-id-458fd959-e5d9-4079-a88e-4ba1ca1a6aa3',
  role: 'ROLE_USER',
  parts: [ { text: 'Hello from TCK!' } ],
  contextId: '1a7060d9-23dc-4f4f-99e5-f5554dc66f31'
}
```

I will remove the transformation middleware in my next commits/push and try my best to propose some relevant type for REST, I'll work on it in the next days.

.github/workflows/run-tck.yaml
```diff
@@ -10,11 +10,13 @@ on:
 env:
   # TODO once we have the TCK for 0.4.0 we will need to look at the branch to decide which TCK version to run.
   # Tag of the TCK
-  TCK_VERSION: 0.3.0.beta3
+  TCK_VERSION: 601bf0d6f63baafb079c2ab23f53d73ce03e5461
```
ishymko, line 14: Let's add a comment here why it was done, maybe referencing this PR comment(s) with our discussion?

src/server/express/a2a_express_app.ts
```diff
@@ -48,6 +49,7 @@ export class A2AExpressApp {
       })
     );
     router.use(`/${agentCardPath}`, agentCardHandler({ agentCardProvider: this.requestHandler }));
+    router.use(httpRestHandler({ requestHandler: this.requestHandler }));
```
ishymko: Let's remove it from the `A2AExpressApp` to avoid registering unexpected new routes by default. We're going to move to the new handlers to be explicit and deprecate the `A2AExpressApp`.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,477 @@
...
+export const restErrorHandler: ErrorRequestHandler = (
+  err: Error,
+  _req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (err instanceof SyntaxError && 'body' in err) {
+    const a2aError = A2AError.parseError('Invalid JSON payload.');
+    return res.status(400).json({
+      error: a2aError.toJSONRPCError(),
```
ishymko: We shouldn't be using JSON-RPC errors here, current v0.3 spec doesn't specify the payload of HTTP+JSON errors, but using JSON-RPC still doesn't feel right to me. Let's follow [Python implementation](https://github.com/a2aproject/a2a-python/blob/3bfbea9ec8d7982fa73eb12d8352a581307355dc/src/a2a/utils/error_handlers.py#L54-L90) for now. 

The v1.0 is going to specify it, see in the draft [here](https://a2a-protocol.org/latest/specification/#53-method-mapping-reference) and [here](https://a2a-protocol.org/latest/specification/#116-error-handling). We will support it when getting ready for 1.0 (and when the final spec version is published). 

Applies to other places using `toJSONRPCError()`.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,477 @@
...
+
+  return router;
+}
+
+/**
+ * @deprecated Use httpRestHandler instead.
+ */
+export function createHttpRestRouter(requestHandler: A2ARequestHandler): Router {
+  return httpRestHandler({ requestHandler }) as Router;
+}
```
ishymko: Looks like a leftover, can we remove it?

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,477 @@
...
+  next(err);
+};
+
+/**
+ * Regular expression patterns for matching REST API routes with actions.
+ */
+const ROUTE_PATTERN = {
+  MESSAGE_ACTION: /^\/v1\/message:(send|stream)$/i,
+  TASK_ACTION: /^\/v1\/tasks\/([^/:]+):([a-z]+)$/i,
+} as const;
```
ishymko: Let's consider simplifying it so that single `router.get`/`route.post` handles one method to make the logic more explicit and avoid advanced regex matching. Code duplication between similar "groups" of actions can be extracted into helper methods.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,477 @@
...
+        if (error instanceof A2AError) throw error;
+        throw A2AError.invalidParams(
+          error instanceof Error ? error.message : 'Invalid message parameters'
+        );
+      }
+
+      switch (action) {
+        case ACTION.STREAM: {
+          await requireCapability('streaming');
+          const stream = await requestHandler.sendMessageStream(params);
```
ishymko: Here and for other `requestHandler` calls, currently it accepts `ServerCallContext` which is used down the road for things like protocol extensions (#171 - was added here) and authentication (#195), see for JSON-RPC [here](https://github.com/a2aproject/a2a-js/blob/45014ac4a192ad3c6120849577530d1d63eb27b8/src/server/express/json_rpc_handler.ts#L41-L44).

We need to support it for HTTP+JSON as well.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,477 @@
...
+  ): Promise<void> => {
+    const agentCard = await requestHandler.getAgentCard();
+    if (!agentCard.capabilities?.[capability]) {
+      const errorMessage =
+        capability === 'streaming'
+          ? 'Agent does not support streaming'
+          : 'Agent does not support push notifications';
+      throw capability === 'pushNotifications'
+        ? A2AError.pushNotificationNotSupported()
+        : A2AError.unsupportedOperation(errorMessage);
```
ishymko: It feels like `errorMessage` is not used when `capability` is set to `pushNotifications`. How about defining an error to throw via a static map `capability` -> `error` for better readability?

src/server/express/index.ts
```diff
@@ -5,3 +5,9 @@
 
 export { A2AExpressApp } from './a2a_express_app.js';
 export type { UserBuilder } from './common.js';
+export { jsonRpcHandler, jsonErrorHandler } from './json_rpc_handler.js';
```
ishymko: `jsonErrorHandler` should stay internal, it's a part of `jsonRpcHandler`.

```suggestion
export { jsonRpcHandler } from './json_rpc_handler.js';
```

src/server/express/index.ts
```diff
@@ -5,3 +5,9 @@
...
+export type { JsonRpcHandlerOptions } from './json_rpc_handler.js';
+export { agentCardHandler } from './agent_card_handler.js';
+export type { AgentCardHandlerOptions, AgentCardProvider } from './agent_card_handler.js';
+export { httpRestHandler, restErrorHandler } from './http_rest_handler.js';
```
ishymko: `restErrorHandler` should stay internal, it's a part of `httpRestHandler`.

```suggestion
export { httpRestHandler } from './http_rest_handler.js';
```

src/server/transports/http_rest_transport_handler.ts
ishymko, line 1: Can we actually introduce a class here similar to JSON-RPC which will perform the type conversion and some validation (like for history length and also capabilities check)? This way it should be possible to shrink down the express part more, the purpose of express part will be to create a transport specific object (i.e. assemble data from body and query params) and pass it down here, where it's going to be converted into the "domain" object, validated and processed.

src/server/transports/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,487 @@
...
+  if (!match) {
+    throw A2AError.methodNotFound('Invalid action path');
+  }
+  // Check if this is a task action pattern (has two groups: taskId and action)
+  const isTaskAction = pattern.source.includes('([^/:]+)') && pattern.source.includes('([a-z]+)');
+  return isTaskAction ? { taskId: match[1], action: match[2] } : { action: match[groupIndex] };
+}
+
+// ============================================================================
+// Server-Sent Events (SSE) Support
```
ishymko: Is it possible to reuse SSE code from [jsonrpc_transport_handler.ts](https://github.com/a2aproject/a2a-js/blob/main/src/server/transports/jsonrpc_transport_handler.ts)? There is an issue for this #175. If it gets too troublesome we can tackle it later, either way as long as it's not exported we can reorganize any time.

ishymko: I've tried to run TCK `capabilities` tests and got SSE/streaming for REST specifically (it passes for JSON-RPC). Currently we don't run `capabilities` in CI, only `mandatory`.

Failures:

```
FAILED tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic - AssertionError: assert 'taskId' in {'context_id': '065ae425-f162-4ab0-b618-18e7c29bcce3', 'final': False, 'kind': 'status-update', 'status': {'message': ... 'text', 'text': 'Processing your question'}], ...}, 'state': 'working', 'timestamp': '2025-11-26T10:45:11.634Z'}, ...}
FAILED tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance - tck.transport.base_client.TransportError: [REST] Unexpected error in REST streaming: Event loop is closed
```

I run it against your branch (TCK against the commit set for CI in this PR) by:

1) `cd tck && npm run tck:sut-agent`
2) `uv run ./run_tck.py --sut-url http://localhost:41241/a2a/jsonrpc --category capabilities --transports jsonrpc,rest`

Could you please check it? If it's some HTTP+JSON implementation bug and you'll be able to get it working, let's include them in CI (there are only two failures in this category). If it's a bug in TCK we can take another look and maybe report it in their repo.

src/server/transports/rest/rest_types.ts
ishymko, line 1: Can be moved to `/transports` I think, i.e. under a new subcatalog `transports/rest` together with the other file.

ishymko (commented):
Hello @tchapacan, thanks for wiring up everything together, this is a solid work!

First pass done, added a few comments on potential reshuffling things a bit, also let's double check SSE/streaming implementation as for some reason it fails TCK test (the one we don't run now, you'll find more details in the specific comment).

> Let's not forget the version of tck as well that is sticked into the github workflow currently to this [commit](https://github.com/a2aproject/a2a-tck/tree/601bf0d6f63baafb079c2ab23f53d73ce03e5461), if it goes in I can open a small issue for that to avoid It's forgotten later.

Sounds good, let's create a follow-up once this one is merged, thank you!

src/server/transports/http_rest_transport_handler.ts
ishymko: Added running `--capabilities` in #204 in main now.

.github/workflows/run-tck.yaml
tchapacan: fixed, added a comment referencing our discussion in this PR for more details

src/server/transports/rest/rest_types.ts
tchapacan: I have reorganized following your recommendation `transports/rest` LMK if it's ok like this. Do we need the same folder for jsonrpc even for one file ?

src/server/express/a2a_express_app.ts
tchapacan: fixed, it's removed

src/server/express/http_rest_handler.ts
tchapacan: Thanks for the details, I didn't know where to go so I reused the jsonrpc error, but I have modified it to try to stick to what was done. I added it there https://github.com/a2aproject/a2a-js/pull/142/files#diff-0a8d5095cb6c69c46617d34aa4c8120096b1aab4122acb5d16bda705d05ba9c3R35-R53 LMK if that's ok

src/server/express/http_rest_handler.ts
tchapacan: Yes forgot about that, it wasn't used anymore so removing it, fixed

src/server/express/http_rest_handler.ts
tchapacan: Fixed, now we have some explicit route for each method. I had to escape here https://github.com/a2aproject/a2a-js/pull/142/files#diff-136c39f8641a28377f16e8825d5cc02aad693dcf6c5514aafbd4cdeaf73d11dbR301-R311 (open `http_rest_handler.ts` to match the diff link as the file is large)  and everywhere else necessary, before the semi colon because i was having issue, pathtoregex used by express was taking it as a param instead of the continuation of the route

src/server/express/http_rest_handler.ts
tchapacan: I tried to add it following the existing pattern in jsonrpc, LMK if it's ok like that. One details, I had to make it optional both in `json_rpc_handler.ts` and `http_rest_handler.ts` because tck test was failing. Not sure the body is always properly sent as expected? Maybe I miss something

- https://github.com/a2aproject/a2a-js/pull/142/files#diff-209d969ec5f6e3d8f786388e1c1e68c93622d3b71f555ae2932c8437c91913acR21
- https://github.com/a2aproject/a2a-js/pull/142/files#diff-209d969ec5f6e3d8f786388e1c1e68c93622d3b71f555ae2932c8437c91913acR41
- https://github.com/a2aproject/a2a-js/pull/142/files#diff-136c39f8641a28377f16e8825d5cc02aad693dcf6c5514aafbd4cdeaf73d11dbR35
- https://github.com/a2aproject/a2a-js/pull/142/files#diff-136c39f8641a28377f16e8825d5cc02aad693dcf6c5514aafbd4cdeaf73d11dbR114

Here the logs of the failing ci https://github.com/a2aproject/a2a-js/actions/runs/19806354084/job/56741087750

src/server/express/index.ts
tchapacan: thanks it's fixed

src/server/express/index.ts
tchapacan: thanks it's fixed

src/server/transports/http_rest_transport_handler.ts
tchapacan: Ok I see, before it was some raw function exported, I have tried to follow the same pattern as jsonrpc transport handler and transformed it into a class. I have kept some part that feel closer to the express handler in the `http_rest_handler.ts` and moved a lot into the `http_rest_transport_handler.ts` LMK if it's ok like that.

src/server/transports/http_rest_transport_handler.ts
tchapacan: I have extracted some SSE code from the `jsonrpc_transport_handler.ts`, and moved it in here https://github.com/a2aproject/a2a-js/pull/142/files#diff-c9c965671d44c0c5230c9ba77895152ae411eb20715fbacc7e3113793ca0ffe5 LMK if it's ok? I have modified a little the jsonrpc handler as well to reflect the change and avoid duplication.

tchapacan: I confirm that I have the same error while trying to reproduce it. I'm gonna have a deeper look, I don't really know where it's failing yet, I might need a little bit more time. I might suspect maybe some camelCase vs snake-case but it's only a guess. I'll come back when I have more for this one and comment in this thread

Something a little odd, it seems after rebasing, tck was failing for jsonrpc because some UserBuilder method was undefined, so I've made it optional to make it pass, but don't know if it's a good idea check my comment in here https://github.com/a2aproject/a2a-js/pull/142#discussion_r2575140773 

here is the related failing ci run https://github.com/a2aproject/a2a-js/actions/runs/19806354084/job/56741087750

src/server/express/http_rest_handler.ts
tchapacan: yes good catch, i went with the static map here and putted it in the new introduce class in `http_rest_transport_handler.ts`

https://github.com/a2aproject/a2a-js/pull/142/files#diff-e33da4df125271ea190143c721b702b6b17c04af0bd55f03d906bc0a3380691aR330-R339 (open `http_rest_transport_handler.ts` as file is large to match the diff link)

src/server/transports/http_rest_transport_handler.ts
tchapacan: related to your comment here https://github.com/a2aproject/a2a-js/pull/142#discussion_r2568975264 I have added the transport method in the workflow too, to make sure we capture both jsonrpc and rest for the capabilities too

https://github.com/a2aproject/a2a-js/pull/142/files#diff-657e29e58ceab8eecf2f442816d6ff70c104022fd017f1e32a2ccb5297c4385dR106

tchapacan: @ishymko I had time to dig a little and found some interesting things. Here the CI run failure I tried to reproduce locally https://github.com/a2aproject/a2a-js/actions/runs/19806959994/job/56742596945

We were getting two failures in the CI and locally (using this https://github.com/a2aproject/a2a-tck/tree/601bf0d6f63baafb079c2ab23f53d73ce03e5461) : 

1. tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic

```sh
FAILED tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic - AssertionError: assert 'taskId' in {'context_id': '91dc4bff-122c-4db6-bfd2-4106051b3223', 'final': False, 'kind': 'status-update', 'status': {'message': ... 'text', 'text': 'Processing your question'}], ...}, 'state': 'working', 'timestamp': '2025-12-01T00:13:17.458Z'}, ...}
```

**camelCase vs snake-case** assertion into the https://github.com/a2aproject/a2a-tck/blob/main/tests/optional/capabilities/test_streaming_methods.py#L50-L152 (and a lot of test are skipped so I suspect this behavior might be present elsewhere too in this file example here too https://github.com/a2aproject/a2a-tck/blob/main/tests/optional/capabilities/test_streaming_methods.py#L558-L635)

I notice the [agnostic pattern](https://github.com/a2aproject/a2a-tck/blob/main/tests/optional/capabilities/test_streaming_methods.py#L86) that lead us in this [transport_helper](https://github.com/a2aproject/a2a-tck/blob/main/tests/utils/transport_helpers.py#L587-L617) calling the proper REST [send_streaming_message](https://github.com/a2aproject/a2a-tck/blob/main/tck/transport/rest_client.py#L246-L335) with the protobuf transformation which is snake-case for REST. But respectively the assertion in the test should be aware of it and assert snake-case too (like in the mandatory test).

`taskId` is tested both for JSON-RPC and REST here causing the test to fail for REST, so I suppose an issue should be raise into **a2a-tck** repo related to that to align everything properly

2. tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance

```sh
FAILED tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance - tck.transport.base_client.TransportError: [REST] Unexpected error in REST streaming: Event loop is closed
```

This one is a little bit sneaker and seems to come from asyncio, and sorry it's been some time I hadn't had the chance to play with some python so I might be wrong, I only have some basic/generic knowledge on eventloop (tokio, libuv etc..). When I run the test locally with more verbosity I noticed this : 

`asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function` 

So `asyncio_default_test_loop_scope` is tied to each test function, (one eventloop = one test) but the [RESTClient](https://github.com/a2aproject/a2a-tck/blob/main/tck/transport/rest_client.py#L28) singleton have the property [async_client](https://github.com/a2aproject/a2a-tck/blob/main/tck/transport/rest_client.py#L115-L128) reuse accross all tests session. There might be an edge-case where this **async_client** could still be bound to a passed event-loop of a precedent test and couldn't be reuse in the next test with a new event-loop? I blindly tried to put this in the[ pyproject.toml](https://github.com/a2aproject/a2a-tck/blob/main/pyproject.toml#L67) of **a2a-tck** and finally the test passed : 

```
asyncio_default_test_loop_scope = "session"
```

> I took the assumption that one eventloop for all the test session was enough, but I suspect if it has been choosen to scope it to a function there might be a good reason, so might be interesting to open an another issue related to that into **a2a-tck** repo and open the discussion to understand better

3. But in the end it lead me on my local environment to a 3rd error (last one i encountered) in here `tests/optional/capabilities/test_streaming_methods.py::test_tasks_resubscribe_nonexistent FAILED`

```
_________________________________________________________________________________________ test_tasks_resubscribe_nonexistent _________________________________________________________________________________________
tests/optional/capabilities/test_streaming_methods.py:467: in test_tasks_resubscribe_nonexistent
    pytest.fail("Expected error or failed status for non-existent task, but got successful events")
E   Failed: Expected error or failed status for non-existent task, but got successful events
```
Maybe a REST opiniated error is not properly raise in the **rest_client** and catch by the test in **a2a-tck**? 

> But this error only appear after switching the `asyncio_default_test_loop_scope` from `function` to `session`, so maybe the eventloop error was hiding it, or the modification caused it

Might need a little help for the next phase, do you want to report the errors or I do it to **a2a-tck** ? 


**To summarize :** 

1. Should be a **camelCase** vs **snake-case** assert issue to fix in **a2a-tck**
2. Open the discussion related to the event loop and client instanciation between test in **a2a-tck**
3. Maybe link to 2 ? Or hidden by 2, honestly I don't really now 

**Question :** 

Do we keep the capabilities tck run in the CI, or wait for it to be fixed ?

tchapacan:
@ishymko to summarize in this comment where we are currently : 

- Streaming tck error [conversation](https://github.com/a2aproject/a2a-js/pull/142#discussion_r2565047768)
-  UserBuilder switch to optional to avoid tck failure see this [conversation](https://github.com/a2aproject/a2a-js/pull/142#discussion_r2575140773)
- Rebased 2 days ago and tried to work and fix all your comments

Sorry I just noticed after that in the flow I sometimes have the bad habits to `commit --amend` all the time (to ease future rebase) and this might have break the review history, I've tried to facilitate the review by pointing the relevant part that I have modified in the different conversations. Next time I will simply add commits on top of it.

src/server/transports/http_rest_transport_handler.ts
ishymko: Great research! 

1. tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic

It appears that TCK serializes requests into snake_case, but is able to read both camelCase and snake_case due to using conversion via protobuf (i.e. [here](https://github.com/a2aproject/a2a-tck/blob/b03fefcae9767dad0e978e11573192f74252dfe3/tck/transport/rest_client.py#L213)) which supports both. I've verified [Java SDK](https://github.com/a2aproject/a2a-java) which has all TCK tests [running in CI](https://github.com/a2aproject/a2a-java/blob/main/.github/workflows/run-tck.yml) and it returns responses in camelCase (added debug print in TCK and executed against Java TCK SUT).

However this conversion via protobuf isn't done for SSE events payload: see [here](https://github.com/a2aproject/a2a-tck/blob/b03fefcae9767dad0e978e11573192f74252dfe3/tck/transport/rest_client.py#L311-L312), so it only expects camelCase as you've discovered.

In addition to that, I believe that other SDKs are able to work both snake_case and camelCase because of the same reason: using protobuf-json deserialization which understands both. From what I saw after running Java SUT TCK: TCK sends snake_case and Java replies with camelCase. And I did a quick local test by patching [this function](https://github.com/a2aproject/a2a-tck/blob/b03fefcae9767dad0e978e11573192f74252dfe3/tck/message_utils.py#L52) in TCK to set fields in camelCase - it worked with Java SDK as well.

It complicates the whole story with snake_case and camelCase, really apologize for confusion, I didn't realize it from the start when we were discussing dedicated snake_case types. Now I actually think that we can go with camelCase for both inputs and outputs and rather skip TCK tests (and start a discussion in the repo about it - I'll do it). Although we can bring Protobuf approach here as well to be compatible with both, I think at this point it's not worth it with the new spec draft [explicitly stating camelCase](https://a2a-protocol.org/latest/specification/#55-json-field-naming-convention).

2. tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance

Very interesting one, my expertise in Python asyncio is also limited, what you're saying makes sense, however I am wondering is it impacted by the previous errors in the pipeline (i.e. lifetime is broken if there are errors in previous tests) or it's an issue on it's own. Java SDK is able to pass all tests [here](https://github.com/a2aproject/a2a-java/actions/workflows/run-tck.yml), so maybe this issue manifests itself only on failures. Feel free to open an issue in the TCK repo, I think TCK will benefit from it.

---

My proposal for now is to keep going with the camelCase, unless there is an easy way to hack around supporting both on read to take advantage of TCK tests until they are updated for camelCase. You can also try patching it locally and run it manually as a smoke test. Let me know what are your thoughts on this!

src/server/express/http_rest_handler.ts
ishymko: I think it can be related to #202, there must be some place where the default user request builder was missing which errored out on each request. Maybe rebasing against main once again? There might have been some instabilities in main even in public API as long as npm package isn't published, apologize for the inconvenience.

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,609 @@
...
+/**
+ * Valid action names for REST API endpoints.
+ * These correspond to the actions in the A2A REST specification.
+ */
+export const ACTION = {
+  SEND: 'send',
+  STREAM: 'stream',
+  CANCEL: 'cancel',
+  SUBSCRIBE: 'subscribe',
+} as const;
```
ishymko: Not used it seems?

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,609 @@
...
+  }
+}
+
+// ============================================================================
+// Server-Sent Events (SSE) Support
+// ============================================================================
+
+// Re-export SSE utilities from shared module for convenience
+// Both JSON-RPC and REST transports use these utilities
+export { SSE_HEADERS, formatSSEEvent, formatSSEErrorEvent } from '../sse_utils.js';
```
ishymko: nit: I wouldn't do it, seems like we import them here only for reexport, it should be fine for express handler to use a shared sse_utils package explicitly (which is and should stay package internal).

src/server/error.ts
```diff
@@ -32,6 +32,25 @@ export class A2AError extends Error {
...
+      code: this.code,
+      message: this.message,
+    };
+
+    if (this.data !== undefined) {
+      errorObject.data = this.data;
+    }
+
+    return errorObject;
+  }
```
ishymko: Let's make it private to HTTP transport. Currently errors are tied to JSON-RPC which should be changed either way for both client and server to make them transport agnostic, I've started it already for a new client [here](https://github.com/a2aproject/a2a-js/blob/main/src/errors.ts), but for server it'd be a breaking change so we'll have to keep it for now. But let's do this conversion inside transport.

src/server/transports/sse_utils.ts
```diff
@@ -0,0 +1,63 @@
...
+// ============================================================================
+
+/**
+ * Standard HTTP headers for Server-Sent Events (SSE) streaming responses.
+ * These headers ensure proper SSE behavior across different proxies and clients.
+ */
+export const SSE_HEADERS = {
+  'Content-Type': 'text/event-stream',
+  'Cache-Control': 'no-cache',
+  Connection: 'keep-alive',
```
ishymko, line 17: nit (for consistency with other keys)
```suggestion
  'Connection': 'keep-alive',
```

src/sse_utils.ts
ishymko, line 1: Can we move it to some cross-cutting shared location for both server and client (there is no shared folder now, so just `src` is fine) to implement reading part later as well for client? Similar to how it's done in [Golang SDK](https://github.com/a2aproject/a2a-go/tree/main/internal/sse). No need to implement client part for now, but would like to put it same way so that we can keep SSE consistent and test "symmetry" of both ends.

src/server/transports/sse_utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @example
+ * ```typescript
+ * formatSSEEvent({ kind: 'message', text: 'Hello' })
+ * // Returns: "id: 1234567890\ndata: {\"kind\":\"message\",\"text\":\"Hello\"}\n\n"
+ *
+ * formatSSEEvent({ result: 'success' }, 'custom-id')
+ * // Returns: "id: custom-id\ndata: {\"result\":\"success\"}\n\n"
+ * ```
+ */
+export function formatSSEEvent(event: unknown, id?: string | number): string {
```
ishymko: nit: Seems like `id` isn't used by callers, can we remove it for now? We shouldn't export this either way so we can add if we need it later.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,471 @@
...
+/**
+ * Express error handler middleware for REST API JSON parse errors.
+ * Catches SyntaxError from express.json() and converts to A2A parse error format.
+ *
+ * @param err - Error thrown by express.json() middleware
+ * @param _req - Express request (unused)
+ * @param res - Express response
+ * @param next - Next middleware function
+ */
+export const restErrorHandler: ErrorRequestHandler = (
```
ishymko: nit: I think it shouldn't be exported here?

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,471 @@
...
+import { UserBuilder } from './common.js';
+
+/**
+ * Options for configuring the HTTP REST handler.
+ */
+export interface HttpRestHandlerOptions {
+  /** The A2A request handler implementation that processes requests */
+  requestHandler: A2ARequestHandler;
+  /** Optional function to build a User from the Express request for authentication */
+  userBuilder?: UserBuilder;
```
ishymko: Let's make it mandatory as in #202.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,471 @@
...
+        req.params.taskId,
+        req.query.historyLength,
+        context
+      );
+      sendResponse(res, HTTP_STATUS.OK, result, context);
+    })
+  );
+
+  /**
+   * POST /v1/tasks/:taskId\\:cancel
```
ishymko: nit: to avoid copying escaping characters to docs we can use [this notation](https://a2a-protocol.org/latest/specification/#53-method-mapping-reference) in comments, so here it'd be
```suggestion
   * POST /v1/tasks/{id}:cancel
```

(here and below)

src/server/transports/rest/rest_types.ts
ishymko: >  Do we need the same folder for jsonrpc even for one file ?

Yeah I think it's a good idea.

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: good catch thanks! it's removed

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: yes I have removed this, now it's direct import in the handler (and symmetric between jsonrpc and rest), fixed

src/server/error.ts
tchapacan: no problem I have moved it as a simple exported function in the transport, not in the class but in the same module, lmk if it's ok like this ?

src/server/transports/sse_utils.ts
tchapacan: thanks it's fixed

src/sse_utils.ts
tchapacan: file has been moved into `src/` while waiting introduction of a shared package between client and server

src/server/transports/sse_utils.ts
tchapacan: I have removed the `id` optional parameter, it wasn't necessary agreed. For the export I'm using it in both handler now to avoid duplication, is it ok ?

src/server/transports/rest/rest_types.ts
tchapacan: I have moved the jsonrpc too

src/server/express/http_rest_handler.ts
tchapacan: yes good catch, it's used in the same module, so removed the unecessary `export`

src/server/express/http_rest_handler.ts
tchapacan: I tried this modification but unfortunately those test are failing :

1) should cancel task and return 202 Accepted
2) should resubscribe to task updates via SSE
3) should return 501 if streaming is not supported
4) should return 404 for unknown task action (route not matched)

i think express don't support this syntax, and the escaping seems to be needed for this specific usecase

src/server/express/http_rest_handler.ts
tchapacan: I have rebased from main and made it mandatory and in the end it was a mistake from me, I completly forgot to pass it in the sut test agent. See in here : https://github.com/a2aproject/a2a-js/pull/142/files#diff-aa3c7f9434c20c5abe3aeebc31813a69e4dfcd6bee926751f3fd83b85db4cdd9R211-R220

now it's working as expected

src/server/express/http_rest_handler.ts
tchapacan: check my comment in here https://github.com/a2aproject/a2a-js/pull/142#discussion_r2590774120

it was on me i forgot to update sut agent for the test  check [here](https://github.com/a2aproject/a2a-js/pull/142/files#diff-aa3c7f9434c20c5abe3aeebc31813a69e4dfcd6bee926751f3fd83b85db4cdd9R211-R220)

src/server/express/http_rest_handler.ts
ishymko: Sorry for confusion, I meant using this syntax only in comments, not in the actual code, as escaping looks a bit strange in comments.

UPD: I see escaping is removed now in the comments, makes sense to me.

src/server/transports/http_rest_transport_handler.ts
ishymko: Created an issue in TCK repo: https://github.com/a2aproject/a2a-tck/issues/97.

tchapacan: 1. tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic

Gotcha! No worry It's a living and evolving spec, it's ok to implement and remove some stuff along the journey. As other SDK seems can handle both as input, and we have already implemented a bunch of boilerplate method for camelCase to snake-case, I've tried to take advantage of it, and support both for the input in my last push. But the output will always be camelCase. But in the end it seems to complicate the implementation with the rest_type the transformation etc.. It could be error prone to maintain in the future, and the spec draft seems quite clear to only support camelCase now. Unless you have an existing pattern you wanted to leverage with the protobuff format.

2. tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance

I'm gonna open an issue in tck and reference this conversation from the PR. Maybe there is something underneath and this early closing of the eventloop is just a symptom. Jsonrpc is working just fine and I've tried to dig a little further, it seems REST is reusing the same client versus jsonrpc use a new client each time. Modifying a global config parameter in asyncio (function vs session eventloop) I don't 100% understand can be risky, so the track of the reused client vs new client might be it, but I don't really know. Anyway I'm gonna open an issue for this specifically.

For the camelCase it's an another subject I let you take the lead as it is a discussion to align multiple sdk as per the evolving a2a spec if that's ok ?

LMK about the current attempt implementation : input(camelCase+snake-case) => output(camelCase)

otherwise I'll simplify it and go with input(camelCase) => output(camelCase)

tchapacan:
Helo @ishymko, to summarize : 

- Tried to fix all your comments, and rebased, LMK if that's ok
- Tried to implement : 
  - input: both camelCase and snake-case accepted
  - output: only camelCase returned
> I went this path to leverage the existing boilerplate method we have implemented (by cleaning some that weren't necessary anymore)
- Open an issue https://github.com/a2aproject/a2a-tck/issues/97 in **a2a-tck** related to the unexpected closed event-loop error

If the camelCase/snake-case input is to flaky and add some unecessary complexity according to the next version of the [spec](https://a2a-protocol.org/latest/specification/#55-json-field-naming-convention) that will authorize only **camelCase**, I can remove it no worry.

.github/workflows/run-tck.yaml
```diff
@@ -94,13 +97,13 @@ jobs:
...
         run: |
-          ./run_tck.py --sut-url ${{ env.SUT_JSONRPC_URL }} --category mandatory
+          ./run_tck.py --sut-url ${{ env.SUT_JSONRPC_URL }} --category mandatory --transports jsonrpc,rest
         working-directory: tck/a2a-tck
       - name: Run TCK (capabilities)
         id: run-tck-capabilities
         timeout-minutes: 5
         run: |
-          ./run_tck.py --sut-url ${{ env.SUT_JSONRPC_URL }} --category capabilities
+          ./run_tck.py --sut-url ${{ env.SUT_JSONRPC_URL }} --category capabilities --transports jsonrpc,rest
```
ishymko: Let's skip rest for now as long as it's blocked by "Event loop is closed" error. We can address it separately.

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,475 @@
...
+ * for business logic.
+ *
+ * @param options - Configuration options including the request handler
+ * @returns Express router configured with all A2A REST endpoints
+ *
+ * @example
+ * ```typescript
+ * const app = express();
+ * const requestHandler = new DefaultRequestHandler(...);
+ * app.use('/api/rest', httpRestHandler({ requestHandler }));
```
ishymko: nit: `userBuilder` is now a required parameter, let's update this example.

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,476 @@
...
+    case A2A_ERROR_CODE.INVALID_PARAMS:
+      return HTTP_STATUS.BAD_REQUEST;
+    case A2A_ERROR_CODE.METHOD_NOT_FOUND:
+    case A2A_ERROR_CODE.TASK_NOT_FOUND:
+      return HTTP_STATUS.NOT_FOUND;
+    case A2A_ERROR_CODE.TASK_NOT_CANCELABLE:
+      return HTTP_STATUS.CONFLICT;
+    case A2A_ERROR_CODE.PUSH_NOTIFICATION_NOT_SUPPORTED:
+    case A2A_ERROR_CODE.UNSUPPORTED_OPERATION:
+      return HTTP_STATUS.NOT_IMPLEMENTED;
```
ishymko: Although it's not specified in 0.3 spec, 1.0 draft treats those as 400: https://a2a-protocol.org/latest/specification/#54-error-code-mappings.

I tend to agree as it basically means that client ignored agent card information about supported capabilities and invoked unsupported methods.

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,476 @@
...
+    return this.requestHandler.getAuthenticatedExtendedAgentCard();
+  }
+
+  /**
+   * Sends a message to the agent.
+   * Accepts both snake_case and camelCase input, returns camelCase.
+   */
+  async sendMessage(
+    params: MessageSendParamsInput,
+    context?: ServerCallContext
```
ishymko: Here and below: let's make it required, we should pass it everywhere. In `JsonRpcTransportHandler` it was optional to avoid breaking changes. For new code here we should establish it as required right away.

test/server/http_rest_handler.spec.ts
```diff
@@ -0,0 +1,472 @@
...
+  describe('POST /v1/message:send', () => {
+    it('should send message and return 201 Created with Task', async () => {
+      (mockRequestHandler.sendMessage as SinonStub).resolves(testTask);
+
+      const response = await request(app)
+        .post('/v1/message:send')
+        .send({ message: restMessage })
+        .expect(201);
+
+      // REST API returns snake_case
```
ishymko: Here and below: I think this comment is outdated now.

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,476 @@
...
+      throw A2AError.invalidParams('historyLength must be a valid integer');
+    }
+    if (parsed < 0) {
+      throw A2AError.invalidParams('historyLength must be non-negative');
+    }
+    return parsed;
+  }
+
+  /**
+   * Normalizes Part input - accepts both snake_case and camelCase for file mimeType.
```
ishymko, line 358: I like the idea of supporting both! 

Let's just make sure we have proper test coverage for both formats. Unless I am missing something I think tests were not yet updated for both (IMO we may just have two test cases per each current test: with snake_case and camelCase).

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,476 @@
...
+   */
+  private normalizePart(part: PartInput): Part {
+    if (part.kind === 'text') return { kind: 'text', text: part.text };
+    if (part.kind === 'file') {
+      const file = this.normalizeFile(part.file as FileInput);
```
ishymko: nit: shouldn't be necessary it seems
```suggestion
      const file = this.normalizeFile(part.file);
```

src/server/express/index.ts
ishymko: Seems like it came back :)

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,476 @@
...
+                        schemes: pnConfig.authentication.schemes,
+                        credentials: pnConfig.authentication.credentials,
+                      }
+                    : undefined,
+                  id: pnConfig.id,
+                  url: pnConfig.url,
+                }
+              : undefined,
+          }
+        : undefined,
```
ishymko: Do we actually need to rebuild this object? We don't have any divergence in camel/snake case here.

src/server/transports/http_rest_transport_handler.ts
ishymko: 1. tests/optional/capabilities/test_streaming_methods.py::test_message_stream_basic

Agree on supporting both, as long as we don't export this conversion logic keeping it for 0.3 is reasonable as other SDKs support both as well.

2. tests/optional/capabilities/test_streaming_methods.py::test_sse_event_format_compliance

Yeah it's not fully clear for me as well. It seems like `test_message_stream_invalid_params` leads to this state. I've tried to run failing tests independently and they pass, while attaching `test_message_stream_invalid_params` leads to `Event loop is closed` error in subsequent streaming related test:

`uv run python -m pytest tests/optional/capabilities/test_streaming_methods.py::test_message_stream_invalid_params tests/optional/capabilities/test_streaming_methods.py::test_tasks_resubscribe  --test-scope all   --sut-url http://localhost:41241 --transports rest -s -v --log-cli-level=INFO` - fails `test_tasks_resubscribe` with `Event loop is closed`


`uv run python -m pytest tests/optional/capabilities/test_streaming_methods.py::test_tasks_resubscribe  --test-scope all   --sut-url http://localhost:41241 --transports rest -s -v --log-cli-level=INFO` - passes

Let's not block on it now, I see you created an issue in the TCK repo 👍

.github/workflows/run-tck.yaml
tchapacan: All right, It's removed, but I have added a little comment TODO referencing the issue in tck repo

src/server/express/http_rest_handler.ts
tchapacan: thanks for the reminder, forgot it, it's fixed

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: I have modify it to error 400 as per the spec, agreed it's fixed!

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: removed the optional, now it's required and updated it everywhere necessary in `http_rest_transport_handler.ts`, I was following the `jsonRpcTransportHandler` implementation, that's why. Fixed!

test/server/http_rest_handler.spec.ts
tchapacan: yes it's fixed!

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: Yes, I have stopped thinking about test in between waiting to converge on a solution, as we are and confirm we will use both format as input and camelCase as output I have now updated and reorganized a little the test case, LMK if it's ok

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: thanks it's fixed, agreed absolutly not necessary

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: nope not needed as now conversion involved, it can be shrink to only necessary, thanks for the precision, it's fixed now

tck/agent/index.ts
```diff
@@ -49,6 +54,8 @@ class SUTAgentExecutor implements AgentExecutor {
       `[SUTAgentExecutor] Processing message ${userMessage.messageId} for task ${taskId} (context: ${contextId})`
     );
 
+    console.log('fullMessage', JSON.stringify(userMessage, null, 2));
```
ishymko: nit: Debugging leftovers? Can be removed?

src/server/transports/rest/http_rest_transport_handler.ts
```diff
@@ -0,0 +1,463 @@
...
+          ? {
+              schemes: pnConfig.authentication.schemes,
+              credentials: pnConfig.authentication.credentials,
+            }
+          : undefined,
+        id: pnConfig.id,
+        url: pnConfig.url,
+      },
+      taskId,
+    };
```
ishymko: nit: it seems that the entire `pnConfig` can be used here?

```suggestion
      pushNotificationConfig: pnConfig,
      taskId,
    };
```

src/server/express/http_rest_handler.ts
```diff
@@ -0,0 +1,475 @@
...
+   * @param res - Express response object
+   * @param statusCode - HTTP status code
+   * @param body - Response body (omitted for 204 responses)
+   * @param context - ServerCallContext for setting extension headers
+   */
+  const sendResponse = (
+    res: Response,
+    statusCode: number,
+    body?: unknown,
+    context?: ServerCallContext
```
ishymko: nit: I believe now `context` can be made required (same for `sendStreamResponse`) as all public methods were updated in the same way.

tck/agent/index.ts
tchapacan: oups! yes forgot it, was using it to debug the tck messages! it's removed

src/server/transports/rest/http_rest_transport_handler.ts
tchapacan: yes this is true, thanks for spotted it! I have fixed it

src/server/express/http_rest_handler.ts
tchapacan: Yes, I have updated it for both `sendResponse` and `sendStreamResponse` and in the end the `handleError` wasn't using it, so simplified it a little.

ishymko:
@tchapacan thanks again, merging the PR! 🚀

As for the client support: correct, the same approach should be followed for REST. See reference PRs for JSON-RPC: #198, #169. Note that current JSON-RPC transport implementation has some compatibility hacks to be used from the old client, there is no need for it for HTTP+JSON. 

There is no TCK-like thing for client, but we have [e2e.spec.ts](https://github.com/a2aproject/a2a-js/blob/main/test/e2e.spec.ts) to test both client and server assembled like external users would do it (using `AgentExecutor` as an extension point with everything else not mocked). Currently those tests provide partial coverage around the most critical part which is message sending, we should be able to run current tests for both transports with HTTP+JSON client.

Could you please comment on #137 that you're planning to open a PR for client support so that I can "assign" this issue to you to avoid duplicated effort from the community? Thank you!

PR #162, opened 2025-11-03
refactor: extract request handlers from A2AExpressApp

ishymko:
# Description

As a preparation step for REST support proposed in #142, extract Express.js handlers from `A2AExpressApp` to allow assembling your own app using Express.js API directly for specifying individual paths and middlewares, rather than wrapping it in `A2AExpressApp`:

```ts
// const a2aRequestHandler = . . .

const app = express();

// Minimal example:
app.use(`/${AGENT_CARD_PATH}`, agentCardHandler({ agentCardProvider: a2aRequestHandler }));
app.use(jsonRpcHandler({ requestHandler: a2aRequestHandler }));

// Advanced example:
app.use(myCustomMiddleware());
app.use('/.well-known/custom-agent-card.json', myCustomAgentCardMiddleware, agentCardHandler({ agentCardProvider: a2aRequestHandler }));
app.use('/a2a/jsonrpc', myJsonRpcMiddleware, jsonRpcHandler({ requestHandler: a2aRequestHandler }));
app.use('/a2a/api', restHandler({ requestHandler: a2aRequestHandler }));
```

No logic changes to existing handlers, they are moved as is. Also `A2AExpressApp` is switched to use new handlers without behavior changes.

**Note:** new functions **are not** exported via `index.ts` yet, it will be done together with updating docs and examples once REST support is done.

Re #137

### DIFF
--- a/src/server/express/a2a_express_app.ts
+++ b/src/server/express/a2a_express_app.ts
@@ -31,6 +31,8 @@ export class A2AExpressApp {
 
         // Doing it here to maintain previous behaviour of invoking provided middlewares
         // after JSON body is parsed, jsonRpcHandler registers JSON parsing on the local router.
+        // body-parser used by express.json() ignores subsequent calls and is safe to be added twice:
+        // https://github.com/expressjs/body-parser/blob/168afff3470302aa28050a8ae6681fa1fdaf71a2/lib/read.js#L41.
         router.use(express.json(), jsonErrorHandler);
 
         if (middlewares && middlewares.length > 0) {


yarolegovich (approved):

src/server/express/a2a_express_app.ts
```diff
@@ -30,97 +30,19 @@ export class A2AExpressApp {
...
-                                id: null,
-                                error: a2aError.toJSONRPCError(),
-                            };
-                return res.status(400).json(errorResponse);
-            }
-            next(err);
-        });
+        // Doing it here to maintain previous behaviour of invoking provided middlewares
+        // after JSON body is parsed, jsonRpcHandler registers JSON parsing on the local router.
+        router.use(express.json(), jsonErrorHandler);
```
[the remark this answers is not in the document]
ishymko: This is temporary for compatibility.

src/server/express/json_rpc_handler.ts
```diff
@@ -0,0 +1,108 @@
...
+
+                res.setHeader('Content-Type', 'text/event-stream');
+                res.setHeader('Cache-Control', 'no-cache');
+                res.setHeader('Connection', 'keep-alive');
+                res.flushHeaders();
+
+                try {
+                    for await (const event of stream) {
+                        // Each event from the stream is already a JSONRPCResult
+                        res.write(`id: ${new Date().getTime()}\n`);
```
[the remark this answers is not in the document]
ishymko: Existing moved code.

src/server/express/agent_card_handler.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface AgentCardProvider {
+    getAgentCard(): Promise<AgentCard>;
+}
+
+/**
+ * Creates Express.js middleware to handle agent card requests.
+ * @example
+ * app.use('/.well-known/agent-card.json', agentCardHandler({ agentCardProvider: a2aRequestHandler }));
+ */
+export function agentCardHandler(options: AgentCardHandlerOptions): RequestHandler {
```
swapydapy, line 20: Any specific reason to have this wrapper interface AgentCardHandlerOptions? Can we just directly specify AgentCardProvider, since it is definitely needed.

Options pattern is used when there are optional params, and the API wants to enable specifying them in any order.

src/server/express/agent_card_handler.ts
```diff
@@ -0,0 +1,31 @@
+import express, { Request, RequestHandler, Response } from "express";
+import { AgentCard } from "../../types.js";
+
+export interface AgentCardHandlerOptions {
+    agentCardProvider: AgentCardProvider;
+}
+
+export interface AgentCardProvider {
```
swapydapy: This interface requires an object which supplies this getAgentCard(), What if the caller directly wants to pass a function.

Can we explore simplifying this to:
```
type AgentCardProvider = () => Promise<AgentCard>;
```

src/server/express/json_rpc_handler.ts
```diff
@@ -0,0 +1,103 @@
...
+ * app.use(jsonRpcHandler({ requestHandler: a2aRequestHandler }));
+ * // or
+ * app.use('/a2a/json-rpc', jsonRpcHandler({ requestHandler: a2aRequestHandler }));
+ */
+export function jsonRpcHandler(options: JsonRpcHandlerOptions): RequestHandler {
+    const jsonRpcTransportHandler = new JsonRpcTransportHandler(options.requestHandler);
+
+    const router = express.Router()
+
+    router.use(express.json(), jsonErrorHandler);
```
swapydapy, line 24: Is there a way to check if express.json() middleware is already added by the user.

src/server/express/json_rpc_handler.ts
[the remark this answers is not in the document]
swapydapy: This seems good point. Lets create a task for it to tackle in subsequent PR.

src/server/express/a2a_express_app.ts
```diff
@@ -30,97 +28,19 @@ export class A2AExpressApp {
...
-                                jsonrpc: '2.0',
-                                id: null,
-                                error: a2aError.toJSONRPCError(),
-                            };
-                return res.status(400).json(errorResponse);
-            }
-            next(err);
-        });
+        // Doing it here to maintain previous behaviour of invoking provided middlewares
+        // after JSON body is parsed, jsonRpcHandler registers JSON parsing on the local router.
```
swapydapy, line 33: What happens if same JSON parsing middleware is specified twice.

Since it is being registered here as well as inside jsonRpcHandler.

src/server/express/json_rpc_handler.ts
```diff
@@ -0,0 +1,103 @@
+import express, { Router, Request, Response, ErrorRequestHandler, NextFunction, RequestHandler } from "express";
+import { JSONRPCErrorResponse, JSONRPCSuccessResponse, JSONRPCResponse } from "../../types.js";
+import { A2AError } from "../error.js";
+import { A2ARequestHandler } from "../request_handler/a2a_request_handler.js";
+import { JsonRpcTransportHandler } from "../transports/jsonrpc_transport_handler.js";
+
+export interface JsonRpcHandlerOptions {
```
swapydapy, line 7: Same comment about probably removing this options wrapper interface.

src/server/express/json_rpc_handler.ts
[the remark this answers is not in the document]
ishymko: Created #174.

src/server/express/agent_card_handler.ts
ishymko: Mainly to leave a room for new parameters and make it similar to other libraries, as this is going to be a public interface, it will to be hard to change later. Having options object also gives explicit names to parameters which helps when you have parameters like functions and literals, it's not the case now, but my idea was to establish it from the start for a public interface.

src/server/express/agent_card_handler.ts
ishymko: The interface allows passing `A2ARequestHandler` instance which I think is going to be a common usage. Having this declaration would require wrapping `A2ARequestHandler` into a lambda.

However I agree that providing a lambda is handy, supported both.

src/server/express/json_rpc_handler.ts
ishymko: Nothing documented/recommended as far as I know, but I don't think we should worry about it, body-parser used by `express.json()` won't parse things twice: https://github.com/expressjs/body-parser/blob/master/lib/read.js#L41.

Adding this middleware at router level seems to be a recommended way: https://expressjs.com/en/resources/middleware/body-parser.html#express-route-specific.

src/server/express/a2a_express_app.ts
ishymko: Second call is no-op: https://github.com/expressjs/body-parser/blob/master/lib/read.js#L41, updated the comment.

For modular apps this should a common thing to have it installed at router level (also recommended [here](https://expressjs.com/en/resources/middleware/body-parser.html#express-route-specific)), so I think it's safe to rely on proper handling of it being invoked multiple times for one request.

swapydapy (approved):

PR #169, opened 2025-11-07
refactor: extract JSON-RPC specific logic from client

ishymko:
# Description

Currently `A2AClient` defined in `client.ts` uses JSON-RPC types in its interface, i.e. `sendMessage` uses [`SendMessageResponse`](https://github.com/a2aproject/a2a-js/blob/e7e8f35d5d5356d30a13131dea0caff97be8cd53/src/client/client.ts#L207) which contains [JSON-RPC fields like `id` and `jsonrpc` version](https://github.com/a2aproject/a2a-js/blob/e7e8f35d5d5356d30a13131dea0caff97be8cd53/src/types.ts#L2177-L2190). It was also the case for Python SDK which required introducing a new [client.py](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/client.py) keeping the old one as [legacy.py](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/legacy.py) for backward compatibility (see [#348](https://github.com/a2aproject/a2a-python/pull/348)).

As a first step of introducing a transport-agnostic client, extract JSON-RPC specific logic into a transport abstraction and switch existing `A2AClient` to it in a backward compatible way to take advantage of existing tests.

**Note:** new types are not exported from `index.ts`, this will be done once new client is ready.

# Changes

1. Define `A2ATransport` abstraction without using JSON-RPC types.
2. Implement `JsonRpcTransport` from existing `A2AClient`.
3. Use new `JsonRpcTransport` from `A2AClient` - new transports are **not going** to be added into this client as it uses JSON-RPC types. This is done in a backward compatible way, existing behavior is preserved: returning JSON-RPC errors as objects, populating JSON-RPC specific fields.
4. Define shared transport-agnostic errors for A2A specific errors defined in [the spec](https://a2a-protocol.org/latest/specification/#82-a2a-specific-errors).

Re #137, #142

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -18,7 +18,8 @@ import {
   Task,
   TaskArtifactUpdateEvent,
   TaskStatusUpdateEvent,
-  A2ARequest} from '../types.js'; // Assuming schema.ts is in the same directory or appropriately pathed
+  A2ARequest,
+} from '../types.js'; // Assuming schema.ts is in the same directory or appropriately pathed
 import { AGENT_CARD_PATH } from '../constants.js';
 import { JsonRpcTransport } from './transports/json_rpc_transport.js';
 
@@ -39,7 +40,7 @@ export class A2AClient {
   private readonly agentCardPromise: Promise<AgentCard>;
   private readonly customFetchImpl?: typeof fetch;
   private serviceEndpointUrl?: string; // To be populated from AgentCard after fetchin
-  
+
   // A2AClient is built around JSON-RPC types, so it will only support JSON-RPC transport, new client with transport agnostic interface is going to be created for multi-transport.
   // New transport abstraction isn't going to expose individual transport specific fields, so to keep returning JSON-RPC IDs here for compatibility,
   // keep counter here and pass it to JsonRpcTransport via an optional idOverride parameter (which is not visible via transport-agnostic A2ATransport interface).
@@ -147,7 +148,10 @@ export class A2AClient {
    * @returns A Promise resolving to SendMessageResponse, which can be a Message, Task, or an error.
    */
   public async sendMessage(params: MessageSendParams): Promise<SendMessageResponse> {
-    return await this.invokeJsonRpc<MessageSendParams, SendMessageResponse>((t, p, id) => t.sendMessage(p, id), params)
+    return await this.invokeJsonRpc<MessageSendParams, SendMessageResponse>(
+      (t, p, id) => t.sendMessage(p, id),
+      params
+    );
   }
 
   /**
@@ -188,7 +192,10 @@ export class A2AClient {
         'Agent does not support push notifications (AgentCard.capabilities.pushNotifications is not true).'
       );
     }
-    return await this.invokeJsonRpc<TaskPushNotificationConfig, SetTaskPushNotificationConfigResponse>((t, p, id) => t.setTaskPushNotificationConfig(p, id), params)
+    return await this.invokeJsonRpc<
+      TaskPushNotificationConfig,
+      SetTaskPushNotificationConfigResponse
+    >((t, p, id) => t.setTaskPushNotificationConfig(p, id), params);
   }
 
   /**
@@ -196,8 +203,13 @@ export class A2AClient {
    * @param params Parameters containing the taskId.
    * @returns A Promise resolving to GetTaskPushNotificationConfigResponse.
    */
-  public async getTaskPushNotificationConfig(params: TaskIdParams): Promise<GetTaskPushNotificationConfigResponse> {
-    return await this.invokeJsonRpc<TaskIdParams, GetTaskPushNotificationConfigResponse>((t, p, id) => t.getTaskPushNotificationConfig(p, id), params)
+  public async getTaskPushNotificationConfig(
+    params: TaskIdParams
+  ): Promise<GetTaskPushNotificationConfigResponse> {
+    return await this.invokeJsonRpc<TaskIdParams, GetTaskPushNotificationConfigResponse>(
+      (t, p, id) => t.getTaskPushNotificationConfig(p, id),
+      params
+    );
   }
 
   /**
@@ -205,8 +217,13 @@ export class A2AClient {
    * @param params Parameters containing the taskId.
    * @returns A Promise resolving to ListTaskPushNotificationConfigResponse.
    */
-  public async listTaskPushNotificationConfig(params: ListTaskPushNotificationConfigParams): Promise<ListTaskPushNotificationConfigResponse> {
-    return await this.invokeJsonRpc<ListTaskPushNotificationConfigParams, ListTaskPushNotificationConfigResponse>((t, p, id) => t.listTaskPushNotificationConfig(p, id), params)
+  public async listTaskPushNotificationConfig(
+    params: ListTaskPushNotificationConfigParams
+  ): Promise<ListTaskPushNotificationConfigResponse> {
+    return await this.invokeJsonRpc<
+      ListTaskPushNotificationConfigParams,
+      ListTaskPushNotificationConfigResponse
+    >((t, p, id) => t.listTaskPushNotificationConfig(p, id), params);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+import { A2ATransport } from './transport.js';
+
+export interface JsonRpcTransportOptions {
+  endpoint: string;
+  fetchImpl?: typeof fetch;
+}
+
+export class JsonRpcTransport implements A2ATransport {
+  private requestIdCounter: number = 1;
+  private customFetchImpl?: typeof fetch;
```
yarolegovich: nit: readonly for fields which are assigned once in constructor

src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+    method: string,
+    params: TParams,
+    idOverride: number | undefined,
+  ): Promise<TResponse> {
+    const requestId = idOverride ?? this.requestIdCounter++;
+    
+    const rpcRequest: JSONRPCRequest = {
+      jsonrpc: "2.0",
+      method,
+      params: params as { [key: string]: any; },
```
yarolegovich: nit: move to parameter constraint  `TParams extends { [key: string]: any }`

src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+      id: requestId,
+    };
+
+    const httpResponse = await this._fetchRpc(rpcRequest);
+
+    if (!httpResponse.ok) {
```
yarolegovich, line 169: httpResponse is going to be ok for error responses. you should be parsing json and checking for error presence in the body

src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+        throw new Error(`RPC error for ${method}: ${errorJson.error.message} (Code: ${errorJson.error.code}, HTTP Status: ${httpResponse.status}) Data: ${JSON.stringify(errorJson.error.data || {})}`);
+      } else {
+        throw new Error(`HTTP error for ${method}! Status: ${httpResponse.status} ${httpResponse.statusText}. Response: ${errorBodyText}`);
+      }
+    }
+
+    const rpcResponse: JSONRPCSuccessResponse = await httpResponse.json();
+
+    if (rpcResponse.id !== requestId) {
+      console.error(`CRITICAL: RPC response ID mismatch for method ${method}. Expected ${requestId}, got ${rpcResponse.id}.`);
```
yarolegovich: nit: Go SDK doesn't perform any validation, but I'd probably be throwing in case of a mismatch

src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+  ): AsyncGenerator<TStreamItem, void, undefined> {
+    if (!response.body) {
+      throw new Error("SSE response body is undefined. Cannot read stream.");
+    }
+    const reader = response.body.pipeThrough(new TextDecoderStream()).getReader();
+    let buffer = "";
+    let eventDataBuffer = "";
+
+    try {
+      while (true) {
```
yarolegovich, line 275: might be cleaner to use `for await (const value of response.body.pipeThrough(new TextDecoderStream())) {`

src/client/transports/json_rpc_transport.ts
```diff
@@ -0,0 +1,341 @@
...
+          buffer = buffer.substring(lineEndIndex + 1);
+
+          if (line === "") {
+            if (eventDataBuffer) {
+              const result = this._processSseEventData<TStreamItem>(eventDataBuffer, originalRequestId);
+              yield result;
+              eventDataBuffer = "";
+            }
+          } else if (line.startsWith("data:")) {
+            eventDataBuffer += line.substring(5).trimStart() + "\n";
```
yarolegovich: nit: `line.substring("data:".length)` less error-prone and more clear

yarolegovich: A2A [spec](https://a2a-protocol.org/latest/specification/#331-json-rpc-20-streaming) makes parsing simpler:
> Each SSE data field contains a complete JSON-RPC 2.0 Response object (specifically, a [SendStreamingMessageResponse](https://a2a-protocol.org/latest/specification/#721-sendstreamingmessageresponse-object)).

So when you slice a line:
```js
 while ((lineEndIndex = buffer.indexOf('\n')) >= 0) {
  const line = buffer.substring(0, lineEndIndex).trim();
```
And it `startsWith("data:")` you can just parse it expecting a well-formed jsonrpc response.  No need to handle consecutive data-line concatenation.

src/client/transports/json_rpc_transport.ts
ishymko: Was moved as is, will refactor this during extracting SSE code to share with REST implementation. Created #175 and referenced comments regarding this implementation.

src/client/transports/json_rpc_transport.ts
ishymko: Same as https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517957983.

ishymko: Same as https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517957983.

src/client/transports/json_rpc_transport.ts
ishymko: Good point, kept error parsing for non-OK responses to throw domain errors if error response was returned with non-OK code, but added `error` check for OK HTTP code as well.

src/client/transports/json_rpc_transport.ts
ishymko: Fair enough, but don't want to change it in this PR, created #176.

src/client/client.ts
```diff
@@ -49,10 +35,14 @@ export interface A2AClientOptions {
  * A2AClient is a TypeScript HTTP client for interacting with A2A-compliant agents.
  */
 export class A2AClient {
-  private agentCardPromise: Promise<AgentCard>;
+  private readonly agentCardPromise: Promise<AgentCard>;
+  private readonly customFetchImpl?: typeof fetch;
+  private serviceEndpointUrl?: string; // To be populated from AgentCard after fetchin
+  private transport?: JsonRpcTransport; // A2AClient is built around JSON-RPC types, so it will only support JSON-RPC transport.
```
swapydapy: We should use A2ATransport here.

src/client/client.ts
```diff
@@ -622,4 +356,34 @@ export class A2AClient {
     }
     return this.serviceEndpointUrl;
   }
+
+  private async invokeJsonRpc<TParams extends JsonRpcParams, TResponse extends JSONRPCResponse>(caller: JsonRpcCaller<TParams, TResponse>, params?: TParams): Promise<TResponse> {
```
swapydapy: This seems redundant since json rpc transport should take care of sending back JSON RPC response.

swapydapy (changes requested):

src/client/client.ts
ishymko: This is by design, as this client uses JSON-RPC types, changing its interface to something else is going to be a breaking change, so it will only support JSON-RPC, for multi-transport support new client is going to be created. Python followed the same approach (see [here](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/legacy.py)) to avoid breaking changes. This one can be deleted for 1.0 release.

In order to maintain the same behavior here I use JSON-RPC transport directly to take advantage of JSON-RPC specific things (like request ID). This is not going to be needed for a new client which will use transport-agnostic types. Also this client isn't going to support other transports (to use `A2ATransport` interface) as it uses JSON-RPC specific types in its public interface.

I rearranged comments a bit to make it more explicit.

src/client/client.ts
ishymko: Same as https://github.com/a2aproject/a2a-js/pull/169#discussion_r2522851919.

src/client/client.ts
swapydapy: Got it.

swapydapy (approved):

PR #185, opened 2025-11-17
ci: integrate prettier check in github workflow

### LINKED ISSUE #177: [Feat]: Lint code on PRs
I.e. [typescript-eslint](https://typescript-eslint.io/) with [Prettier](https://typescript-eslint.io/users/what-about-formatting/#suggested-usage---prettier).

guglielmo-san:
# Description

This PR integrates the prettier check in the github workflow to ensure the code respects the correct formatting standards

Fixes #177  🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "pretest": "npm run build",
     "test": "mocha test/**/*.spec.ts",
     "lint": "npx eslint .",
-    "format": "prettier --write ./README.md",
+    "format:readme": "prettier --write ./README.md",
     "lint:fix": "npx eslint . --fix",
     "coverage": "c8 npm run test",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",


src/types.ts
ishymko, line 1: We need to exempt this file, otherwise it's going to be changed on each regeneration. Or it should run formatting as a part regeneration script.

.github/workflows/eslint.yml
ishymko, line 1: Let's also reformat code snippets from README.md according to the new rules. Not sure how easy it'd be to do automatically, but at least let's make it consistent now.

package.json
```diff
@@ -71,6 +72,7 @@
     "pretest": "npm run build",
     "test": "mocha test/**/*.spec.ts",
     "lint": "npx eslint .",
+    "format": "prettier --write ./README.md",
```
ishymko: How about
```suggestion
    "format:readme": "prettier --write ./README.md",
```

Also it seems like script names use camelCase, see `prepublishOnly` in the [official docs](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts), so maybe we should switch to this convention later.

Issue #179, opened 2025-11-14
[Cleanup]: 1.0 breaking changes


ishymko:
Umbrella issue to collect all scheduled API breaking changes which can be done with 1.0 release

- [ ] `A2AClient` - this client is built around JSON-RPC types and can't be used in transport-agnostic way. New transport agnostic client can be renamed to `A2AClient` or this one can just be removed.
- [ ] `JsonRpcTransport` - remove all the code created for compatibility with the old `A2AClient` client.
- [ ] `ServerCallContext` - the parameter should be made mandatory in all extension points.
- [ ] `A2AExpressApp` - should be removed in favor of using middlewares directly.

PR #190, opened 2025-11-19
build: enable no-explicit-any in eslint

### LINKED ISSUE #184: [Feat]: Remove exemption for 'any' keyword in the eslint.config

_No response_

rules: {
      '@typescript-eslint/no-explicit-any': 'off', 
}

Allow this rule defined in the eslint config only for the test folder

_No response_

### Additional context

_No response_

### Code of Conduct

guglielmo-san:
# Description

The purpose of this PR is removing all the references to keyword 'any' present in the repository, as it is considered a bad practice. Future adoptions will be triggered by the linter automatic checks

Fixes #184  🦕

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -360,7 +360,7 @@ export class JsonRpcTransport implements A2ATransport {
         e
       );
       throw new Error(
-        `Failed to parse SSE event data: "${jsonData.substring(0, 100)}...". Original error: ${(e instanceof SyntaxError && e.message) || 'Unknown error'}`
+        `Failed to parse SSE event data: "${jsonData.substring(0, 100)}...". Original error: ${(e instanceof Error && e.message) || 'Unknown error'}`
       );
     }
   }


src/client/transports/json_rpc_transport.ts
```diff
@@ -354,9 +359,13 @@ export class JsonRpcTransport implements A2ATransport {
...
-      throw new Error(
-        `Failed to parse SSE event data: "${jsonData.substring(0, 100)}...". Original error: ${e.message}`
-      );
+      if (e instanceof SyntaxError) {
+        throw new Error(
+          `Failed to parse SSE event data: "${jsonData.substring(0, 100)}...". Original error: ${e.message}`
+        );
+      } else {
+        throw e;
+      }
```
ishymko: Feels like a behavior change, previously all errors were wrapped here. I think this can be refactored to avoid `startsWith` matching above similar to #167 which should also make this block cleaner. Can we avoid checking for `SyntaxError` specifically for now and refactor it in a dedicated PR?

src/client/client.ts
```diff
@@ -442,24 +444,34 @@ export class A2AClient {
...
     'errorResponse' in error &&
-    error.errorResponse &&
+    typeof error.errorResponse === 'object' &&
+    error.errorResponse !== null &&
+    'jsonrpc' in error.errorResponse &&
     error.errorResponse.jsonrpc === '2.0' &&
-    error.errorResponse.error
-  );
+    'error' in error.errorResponse &&
+    error.errorResponse.error !== null
```
ishymko: nit: I think it can be simplified a bit by replacing `typeof` and `null` check with `instanceof Object`.

```suggestion
    error instanceof Object &&
'errorResponse' in error &&
    error.errorResponse instanceof Object &&
    'jsonrpc' in error.errorResponse &&
error.errorResponse.jsonrpc === '2.0' &&
    'error' in error.errorResponse &&
    error.errorResponse.error !== null
```

src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -25,7 +25,7 @@ export class JsonRpcTransportHandler {
    * For non-streaming methods, it returns a Promise of a single JSONRPCMessage (Result or ErrorResponse).
    */
   public async handle(
-    requestBody: any,
+    requestBody: string | A2ARequest,
```
ishymko: Hmm, this is going to be a breaking change, this type is exported from `index.ts`. Let's suppress it here for now and add a TODO linking with #179 to fix later.

PR #195, opened 2025-11-21
feat: support authentication on server side

guglielmo-san:
# Description

This PR integrates a support for authentication on the server side.
A new `User` interface is created to represent the Authenticated and UnAuthenticated user in the context, with 2 required  function:
1. isAuthenticathed()
2. userName()

The callback `authenticathedUserExtractor` provided to the `A2AExpressApp` is responsible for extracting the user object from the Request (if set up by the authentication middleware), and returning an instance of the `User` interface. The returned instance will be added to the ServerCallContext.
If no authentication middleware or no `authenticathedUserExtractor` is provided to the Server, the user in the ServerCallContext will be of type UnAuthenticated.

Fixes #194  🦕

Release-As: 0.3.6

PR #171, opened 2025-11-07
feat: server side support for extensions

guglielmo-san:
# Description

The code submitted in the PR introduces the support for extensions on the server side.

Thank you for opening a Pull Request!
Before submitting your PR, there are a few things you can do to make sure it goes smoothly:

- [x] Follow the [`CONTRIBUTING` Guide](https://github.com/google-a2a/a2a-js/blob/main/CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
  - Important Prefixes for [release-please](https://github.com/googleapis/release-please):
    - `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
    - `feat:` represents a new feature, and correlates to a SemVer minor.
    - `feat!:`, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the `!`) and will result in a SemVer major.
- [x] Ensure the tests and linter pass
- [x] Appropriate docs were updated (if necessary)

BEGIN_COMMIT_OVERRIDE
feat: server side support for extensions
END_COMMIT_OVERRIDE

Fixes #170 🦕

PR #200, opened 2025-11-26
fix: Populate the ServerCallContext user param when a new ServerCallContext is created

### LINKED ISSUE #199: [Bug]: The user set by authentication is removed from the server call context
### What happened?

The server call context does not contain the user create and authenticated

### Relevant log output

```shell

```

### Code of Conduct

guglielmo-san:
# Description

This PR is a fix for the bug caused by the user being deleted when the request is containing requested extensions, as a new ServerCallContext was created without copying the `context.user`.

Fixes #199 🦕
PRs #171, #195 

BEGIN_COMMIT_OVERRIDE
refactor: Populate the ServerCallContext user param when a new ServerCallContext is created
END_COMMIT_OVERRIDE

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -150,14 +150,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
       contextId,
       taskId,
     };
-    return new RequestContext(
-      messageForContext,
-      taskId,
-      contextId,
-      task,
-      referenceTasks,
-      context
-    );
+    return new RequestContext(messageForContext, taskId, contextId, task, referenceTasks, context);
   }
 
   private async _processEvents(


src/server/request_handler/default_request_handler.ts
ishymko: nit: can we initialize with `context` here instead and omit `?? context` below? Or maybe even override parameter itself and do not create `newServerCallContext`.

src/server/request_handler/default_request_handler.ts
```diff
@@ -143,7 +135,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
           exposedExtensions.has(extension)
         )
       );
-      newServerCallContext = new ServerCallContext(validExtensions);
+      newServerCallContext = new ServerCallContext(validExtensions, context.user);
```
ishymko: Should we make `User` a mandatory parameter in `ServerCallContext` to avoid such things? This type wasn't released yet, so it's not going to be a breaking change.

src/server/index.ts
```diff
@@ -29,4 +28,5 @@ export type { DefaultPushNotificationSenderOptions } from './push_notification/d
 export type { PushNotificationStore } from './push_notification/push_notification_store.js';
 export { InMemoryPushNotificationStore } from './push_notification/push_notification_store.js';
 
-export type { User, UnauthenticatedUser } from './authentication/user.js';
+export type { User } from './authentication/user.js';
+export { UnauthenticatedUser } from './authentication/user.js';
```
ishymko, line 33: Is it intentional?

src/server/request_handler/default_request_handler.ts
ishymko, line 1: Can we link PR(s) which introduced the original code in the description? 

Also let's use `BEGIN_COMMIT_OVERRIDE` to avoid including this into the changelog. The code under fix wasn't released yet, so from the end user perspective it's not a fix from the previous version, just a new functionality. PR title makes sense though. I.e.

```
BEGIN_COMMIT_OVERRIDE
refactor: Populate the ServerCallContext user param when a new ServerCallContext is created
END_COMMIT_OVERRIDE
```

src/server/request_handler/default_request_handler.ts
guglielmo-san: Uhm I am not fully convinced on this, but in any case i would do it in a different PR

src/server/index.ts
guglielmo-san: Yes, the Unauthenticated user was exported as a type, but actually it is a class. It is a small item, and i noticed it writing the test for this PR, i don't think it is necessary to separate in an ad hoc PR

ishymko (approved):

PR #211, opened 2025-11-27
docs: create a sample for authentication feature

guglielmo-san:
# Description

This PR introduces a sample for the authentication feature.

Re #195

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -84,7 +84,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
     if (typeof this.extendedAgentCardProvider === 'function') {
       return this.extendedAgentCardProvider(context);
     }
-    if (context?.user?.isAuthenticated()) {
+    if (context?.user?.isAuthenticated) {
       return this.extendedAgentCardProvider;
     }
     return this.agentCard;


src/samples/authentication/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  // 3. Create DefaultRequestHandler
+  const requestHandler = new DefaultRequestHandler(
+    authenticationAgentCard,
+    taskStore,
+    agentExecutor
+  );
+
+  // 4. Create and setup A2AExpressApp, with user builder and authentication middleware
+  const appBuilder = new A2AExpressApp(requestHandler, userBuilder);
+  const expressApp = appBuilder.setupRoutes(express(), '', [authenticationHandler]);
```
ishymko: Let's use new API added in #162. Either way this sample can't be used with currently published version and the new version should contain all changes.

src/samples/authentication/user_builder.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  public email(): string {
+    return this._email;
+  }
+
+  public role(): string {
+    return this._role;
+  }
+}
```
ishymko, line 12: I'd simplify the declaration and put the content of this file together with authentication middleware to make it clear how things work together and minimize fragmentation.

```suggestion
export class CustomUser implements User {
  constructor(
    public userName: string,
    public email: string,
    public role: string
  ) {}

  public isAuthenticated(): boolean {
    return true;
  }
}
```

src/samples/authentication/agent_executor.ts
ishymko, line 1: I'd consider bundling everything together in `index.ts` for simplicity.

PR #214, opened 2025-11-28
build: run tsc --noEmit on lint

guglielmo-san:
# Description

`tsup` does not perform strict type checking before the generation of the `d.ts` files. Adding the `tsc --noEmit` step checks for types violation.


src/samples/agents/movie-agent/index.ts
```diff
@@ -253,7 +253,8 @@ const movieAgentCard: AgentCard = {
     organization: 'A2A Samples',
     url: 'https://example.com/a2a-samples', // Added provider URL
   },
-  version: '0.0.2', // Incremented version
+  version: '0.0.2', // Incremented version,
```
ishymko: nit:
```suggestion
  version: '0.0.2', // Incremented version
```

PR #216, opened 2025-11-28
refactor: create a dedicated package.json for samples

guglielmo-san:
# Description

This PR creates a new `package.json` in the `samples` folder, not to pollute the main `package.json` with libraries specific to the samples.

### DIFF
--- a/src/samples/package.json
+++ b/src/samples/package.json
@@ -1,7 +1,4 @@
 {
-  "name": "@a2a-js/samples",
-  "version": "0.3.5",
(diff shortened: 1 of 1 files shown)


src/samples/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@a2a-js/samples",
+  "version": "0.3.5",
+  "description": "Samples for @a2a-js/sdk",
```
ishymko: Let's remove name, version and description as we're not going to publish examples to npm.

PR #219, opened 2025-11-30
feat: remove EventEmitter dependency to support Edge Runtime

### LINKED ISSUE #218: [Feat]: Edge Runtime Support

PR #71 made Express optional, but the SDK still can't run on edge runtimes 
(Cloudflare Workers, Deno, Bun) because:

1. `DefaultExecutionEventBus` uses Node.js `EventEmitter`
2. Extension utilities (`HTTP_EXTENSION_HEADER`, `getRequestedExtensions`) aren't exported

This forces anyone wanting Hono/Fastify support to maintain a full fork.

### 1. Replace EventEmitter with EventTarget
- Web-standard API available in all modern runtimes
- Node.js 15+ has native EventTarget support
- Maintains same public API

### 2. Export extension utilities
// In src/server/index.ts
export { HTTP_EXTENSION_HEADER } from '../constants.js';
export { getRequestedExtensions } from './utils.js';

## Benefits

- ✅ Native Cloudflare Workers support (no `nodejs_compat`)
- ✅ Native Deno/Bun support
- ✅ Browser compatibility
- ✅ Enables community framework adapters
- ✅ Smaller bundle for edge deployments

I've implemented this in my fork: https://github.com/drew-foxall/a2a-js-sdk

### Additional context

I've also used my fork to create working implementations using hono to create alternate versions of the a2a-examples https://github.com/drew-foxall/a2a-js-sdk-examples

### Code of Conduct

drew-foxall:
## Summary

This PR addresses [issue #218](https://github.com/a2aproject/a2a-js/issues/218) by replacing the Node.js-specific `EventEmitter` with the web-standard `EventTarget` API, enabling the SDK to run natively on edge runtimes (Cloudflare Workers, Deno, Bun) without requiring Node.js compatibility modes.

### Replace EventEmitter with EventTarget

**File:** `src/server/events/execution_event_bus.ts`

- Removes Node.js `events` module dependency
- Uses web-standard `EventTarget` API available in all modern runtimes
- Maintains backward-compatible API (`on`, `off`, `once`, `removeAllListeners`)
- Adds `ExecutionEventName` type alias for `'event' | 'finished'`
- Includes `CustomEvent` polyfill for Node.js 15-18 (built-in from Node.js 19+)

### EventEmitter Behavior Compatibility

The implementation faithfully replicates EventEmitter semantics:

- **FIFO dispatch order** - Listeners fire in the order they were registered
- **`this` context** - Listeners are invoked with `this` bound to the event bus instance
- **Memory semantics** - Listeners are held via strong references until explicitly removed (identical to EventEmitter)

### Type Safety

- Uses type guards (`isAgentExecutionCustomEvent`) for safe runtime narrowing - no unsafe casts
- `'finished'` event listeners are invoked with no arguments (using `Function.prototype.call`) matching EventEmitter behavior
- Internal listener tracking uses `readonly` Maps

## Benefits

- ✅ **Native Cloudflare Workers support** - No `nodejs_compat` flag needed
- ✅ **Native Deno support** - No `npm:` shims required  
- ✅ **Native Bun support** - Full web API compatibility
- ✅ **Browser compatibility** - True universal JavaScript
- ✅ **Node.js 15+** - EventTarget is built-in

## Breaking Changes

- Requires **Node.js 15+** (EventTarget support)
- Node.js 14 no longer supported (EOL anyway)

## Testing

- All **146 existing tests pass** with these changes
- Added **27 new tests** for `DefaultExecutionEventBus` covering:
  - Basic `on`/`off`/`once` functionality
  - Multiple registrations of same listener
  - `once()` auto-cleanup behavior
  - `removeAllListeners()` for specific events and all events
  - Event firing order (FIFO)
  - Method chaining
  - `this` context binding

## Related Work

I've implemented and tested these changes in my fork:
- Fork: https://github.com/drew-foxall/a2a-js-sdk
- Working Hono implementation: https://github.com/drew-foxall/a2a-js-sdk-examples

Closes #218

Release-As: 0.3.7

### DIFF
--- a/test/server/execution_event_bus.spec.ts
+++ b/test/server/execution_event_bus.spec.ts
@@ -377,4 +377,64 @@ describe('DefaultExecutionEventBus', () => {
       expect(thisCapture.value).to.equal(eventBus);
     });
   });
+
+  describe('listener errors', () => {
+    async function captureUncaughtException(run: () => void): Promise<void> {
+      // Node's EventTarget reports listener errors as an uncaught exception on nextTick.
+      // We attach a temporary handler to prevent vitest from failing the run.
+      let handlerCalled = false;
+      const handler = () => {
+        handlerCalled = true;
+      };
+
+      // Use the real Node global when available (no casts); `typeof` is safe even if `process` is missing.
+      const nodeProcess = typeof process !== 'undefined' ? process : undefined;
+      if (nodeProcess) {
+        nodeProcess.once('uncaughtException', handler);
+      }
+
+      run();
+
+      // Give Node a tick to surface the uncaught exception (if any), then clean up.
+      await new Promise((resolve) => setTimeout(resolve, 0));
+
+      if (!handlerCalled && nodeProcess) {
+        nodeProcess.removeListener('uncaughtException', handler);
+      }
+    }
+
+    it('should continue calling other event listeners when one throws', async () => {
+      let secondCalled = false;
+
+      await captureUncaughtException(() => {
+        eventBus.on('event', () => {
+          throw new Error('listener error');
+        });
+        eventBus.on('event', () => {
+          secondCalled = true;
+        });
+
+        eventBus.publish(createMessage('test'));
+      });
+
+      expect(secondCalled).to.equal(true);
+    });
+
+    it('should continue calling other finished listeners when one throws', async () => {
+      let secondCalled = false;
+
+      await captureUncaughtException(() => {
+        eventBus.on('finished', () => {
+          throw new Error('listener error');
+        });
+        eventBus.on('finished', () => {
+          secondCalled = true;
+        });
+
+        eventBus.finished();
+      });
+
+      expect(secondCalled).to.equal(true);
+    });
+  });
 });
--- a/vitest.edge.config.ts
+++ b/vitest.edge.config.ts
@@ -1,20 +1,9 @@
-import { defineConfig } from 'vitest/config';
+import { mergeConfig } from 'vitest/config';
+import defaultConfig from './vitest.config';
 
-/**
- * Vitest configuration for Edge runtime testing.
- *
- * This config runs all tests EXCEPT the express subfolder, which contains
- * Node.js-specific code (Express framework) that is not compatible with
- * Edge runtime environments.
- *
- * The Edge runtime is provided by @edge-runtime/vm, which emulates
- * Cloudflare Workers, Vercel Edge Functions, and similar environments.
- */
-export default defineConfig({
+export default mergeConfig(defaultConfig, {
   test: {
-    globals: false,
     environment: 'edge-runtime',
-    include: ['test/**/*.spec.ts'],
     exclude: [
       // Express tests require Node.js-specific APIs (http, Express framework)
       'test/server/a2a_express_app.spec.ts',


src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,113 @@
...
+    const wrappedListener: WrappedListener = (e: Event) => {
+      // Clean up tracking
+      this.listenerMap.delete(listener);
+      this.allListeners.get(eventName)?.delete(listener);
+
+      if (e instanceof CustomEvent) {
+        listener(e.detail);
+      } else {
+        listener(undefined as unknown as AgentExecutionEvent);
+      }
```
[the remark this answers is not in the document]
drew-foxall: Suggestion applied 🙇🏻‍♂️

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,113 @@
...
+   * EventEmitter-compatible 'on' method.
+   * Wraps the listener to extract event detail from CustomEvent.
+   */
+  on(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this {
+    const wrappedListener: WrappedListener = (e: Event) => {
+      if (e instanceof CustomEvent) {
+        listener(e.detail);
+      } else {
+        listener(undefined as unknown as AgentExecutionEvent);
+      }
```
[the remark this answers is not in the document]
drew-foxall: Suggestion applied 🙇🏻‍♂️

src/server/events/execution_event_bus.ts
```diff
@@ -1,28 +1,192 @@
...
   publish(event: AgentExecutionEvent): void;
-  on(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
-  off(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
-  once(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
+  on(eventName: 'event', listener: EventListener): this;
+  on(eventName: 'finished', listener: FinishedListener): this;
+  off(eventName: 'event', listener: EventListener): this;
+  off(eventName: 'finished', listener: FinishedListener): this;
+  once(eventName: 'event', listener: EventListener): this;
+  once(eventName: 'finished', listener: FinishedListener): this;
```
ishymko: This is going to be at least build-time breaking change for users implemented this interface (or inherited from `DefaultExecutionEventBus`), as both are exported via `server/index.ts`.

I.e. rolling back changes your changes in a sample prevents it from compiling.

Let's try to stick with current public API so that we can benefit from this change before major version release (no exact plans at the moment).

`extensions.ts` can be used as a "compatibility test" and shouldn't require any changes.

drew-foxall: Good point. I'll go back and line things up with the existing API

drew-foxall: Thanks for the feedback and yes of course, I've reverted this change.

src/server/events/execution_event_bus.ts
ishymko: I think extracting `'event' | 'finished'` into a type alias like `ExecutionEventName` makes sense considering increasing amount of duplicates. It shouldn't be a public API breaking change as it's just a type alias (we need to export it from `src/server/index.ts` though).

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+type Listener = (event: AgentExecutionEvent) => void;
+
+/**
+ * Web-compatible ExecutionEventBus using EventTarget.
+ * Works across all modern runtimes: Node.js 15+, browsers, Cloudflare Workers, Deno, Bun.
+ *
+ * This is a drop-in replacement for Node.js EventEmitter with identical API and
+ * memory semantics. Listeners are held until explicitly removed (via `off()` or
+ * `removeAllListeners()`) or until the instance is garbage collected - exactly
+ * like EventEmitter. No additional cleanup is required beyond standard practices.
```
ishymko: Seems it's not fully accurate:

1. > with identical API

[Node 18 docs](https://nodejs.org/docs/latest-v18.x/api/events.html#class-eventemitter) (we use it in CI now) list more methods which were previously here for users extending from `DefaultExecutionEventBus` (as it used to `extends EventEmitter`). As `DefaultExecutionEventBus` is exported, in theory it can break users who extended from this class and used other `EventEmitter` methods. I think it's very unlikely and I don't think that we should implement `EventEmitter` methods not used here originally, however it'd be good to rephrase the comment.

2. > or until the instance is garbage collected

Not sure it's the case, I don't see any weak references being used and also can't find same mention in Node docs.

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+ */
+export class DefaultExecutionEventBus extends EventTarget implements ExecutionEventBus {
+  // Track original listeners to their wrapped versions for proper removal.
+  // Structure: eventName -> listener -> array of wrapped listeners (to handle multiple registrations)
+  private listenerMap: Map<'event' | 'finished', Map<Listener, WrappedListener[]>> = new Map();
+
   constructor() {
     super();
+    this.listenerMap.set('event', new Map());
+    this.listenerMap.set('finished', new Map());
```
ishymko: nit: maybe let's move complete `listenerMap` init logic into one place? (doing the same in constructor also works)

```suggestion
  private listenerMap: Map<'event' | 'finished', Map<Listener, WrappedListener[]>> = new Map([
    ['event', new Map()],
    ['finished', new Map()],
  ]);

  constructor() {
    super();
  }
```

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+ *
+ * This is a drop-in replacement for Node.js EventEmitter with identical API and
+ * memory semantics. Listeners are held until explicitly removed (via `off()` or
+ * `removeAllListeners()`) or until the instance is garbage collected - exactly
+ * like EventEmitter. No additional cleanup is required beyond standard practices.
+ */
+export class DefaultExecutionEventBus extends EventTarget implements ExecutionEventBus {
+  // Track original listeners to their wrapped versions for proper removal.
+  // Structure: eventName -> listener -> array of wrapped listeners (to handle multiple registrations)
+  private listenerMap: Map<'event' | 'finished', Map<Listener, WrappedListener[]>> = new Map();
```
ishymko: nit: 
```suggestion
  private readonly listenerMap: Map<'event' | 'finished', Map<Listener, WrappedListener[]>> = new Map();
```

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+   * Supports multiple registrations of the same listener (like EventEmitter).
+   */
+  on(eventName: 'event' | 'finished', listener: Listener): this {
+    const wrappedListener: WrappedListener = (e: Event) => {
+      if (e.type === 'event') {
+        listener((e as CustomEvent<AgentExecutionEvent>).detail);
+      } else {
+        // 'finished' event - call with undefined (matches EventEmitter behavior)
+        listener(undefined as unknown as AgentExecutionEvent);
+      }
```
ishymko: How about extracting this and the same block in `once` into a helper function?

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+    const wrappedListeners = eventListeners.get(listener);
+    if (wrappedListeners && wrappedListeners.length > 0) {
+      // Remove the most recently added listener (LIFO)
+      const wrappedListener = wrappedListeners.pop()!;
+      this.removeEventListener(eventName, wrappedListener);
+      // Clean up the map entry if no more wrapped listeners
+      if (wrappedListeners.length === 0) {
+        eventListeners.delete(listener);
+      }
+    }
```
ishymko: nit: should `this.removeEventListener(eventName, wrappedListener);` be the last statement here to be consistent with other methods were underlying `EventTarget` methods are invoked last? There should be no functional difference but it'd be more clear IMO to have the same style everywhere.

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+   * Listener is automatically removed after first invocation.
+   * Supports multiple registrations of the same listener (like EventEmitter).
+   */
+  once(eventName: 'event' | 'finished', listener: Listener): this {
+    const eventListeners = this.listenerMap.get(eventName)!;
+
+    const wrappedListener: WrappedListener = (e: Event) => {
+      // Clean up tracking for this specific wrapped listener
+      const wrappedListeners = eventListeners.get(listener);
+      if (wrappedListeners) {
```
ishymko: nit: (to be the same as in `off`)
```suggestion
      if (wrappedListeners && wrappedListeners.length > 0) {
```

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+        const index = wrappedListeners.indexOf(wrappedListener);
+        if (index !== -1) {
+          wrappedListeners.splice(index, 1);
+        }
+        if (wrappedListeners.length === 0) {
+          eventListeners.delete(listener);
+        }
```
ishymko: Can this be moved under the above `if (index !== -1) {`?

src/server/events/execution_event_bus.ts
```diff
@@ -13,16 +11,165 @@ export interface ExecutionEventBus {
...
+      }
+    };
+
+    const eventListeners = this.listenerMap.get(eventName)!;
+    const wrappedListeners = eventListeners.get(listener);
+    if (wrappedListeners) {
+      wrappedListeners.push(wrappedListener);
+    } else {
+      eventListeners.set(listener, [wrappedListener]);
+    }
```
ishymko: Same as [here](https://github.com/a2aproject/a2a-js/pull/219/files#r2611048468), a helper function maybe?

src/server/events/execution_event_bus.ts
drew-foxall: I've gone ahead and removed this. I think the comment ended a little muddled as I removed weak map in favour of map.

src/server/events/execution_event_bus.ts
drew-foxall: Gone ahead and added this ^-^

src/server/events/execution_event_bus.ts
drew-foxall: Implemented

src/server/events/execution_event_bus.ts
```diff
@@ -1,28 +1,284 @@
...
-  on(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
-  off(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
-  once(eventName: 'event' | 'finished', listener: (event: AgentExecutionEvent) => void): this;
-  removeAllListeners(eventName?: 'event' | 'finished'): this;
+  on(eventName: ExecutionEventName, listener: (event: AgentExecutionEvent) => void): this;
+  off(eventName: ExecutionEventName, listener: (event: AgentExecutionEvent) => void): this;
+  once(eventName: ExecutionEventName, listener: (event: AgentExecutionEvent) => void): this;
+  removeAllListeners(eventName?: ExecutionEventName): this;
   finished(): void;
 }
```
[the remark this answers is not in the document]
drew-foxall: The existing codebase (execution_event_queue.ts) already uses () => void listeners for 'finished' events, which TypeScript allows because functions with fewer parameters are assignable to function types expecting more. Changing this would be a breaking API change and is likely out of scope.

src/server/events/execution_event_bus.ts
```diff
@@ -1,28 +1,284 @@
...
+   * @param listener The callback function to invoke when the event is emitted.
+   * @returns This instance for method chaining.
+   */
+  on(eventName: ExecutionEventName, listener: (event: AgentExecutionEvent) => void): this {
+    if (eventName === 'event') {
+      this.addEventListenerInternal(listener);
+    } else {
+      // For 'finished' events, the listener is called with no arguments.
+      // The interface types it as receiving AgentExecutionEvent for API simplicity,
+      // but the actual runtime behavior (matching EventEmitter) passes no args.
```
ishymko: Let's remove this comment here as it's explained in `addFinishedListenerInternal` comment (and there it has better context).

src/server/events/execution_event_bus.ts
```diff
@@ -1,28 +1,284 @@
...
+
+  /**
+   * Removes a wrapped listener from the tracking map (for once cleanup).
+   * Returns true if the listener was found and removed.
+   */
+  private untrackWrappedListener(
+    listenerMap: Map<Listener, WrappedListener[]>,
+    listener: Listener,
+    wrapped: WrappedListener
+  ): boolean {
```
ishymko: Returned value is never used, let's remove it.

test/server/execution_event_bus.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+
+      eventBus.on('event', listener);
+      eventBus.off('event', listener);
+
+      eventBus.publish(createMessage('test'));
+
+      expect(callCount).to.equal(0);
+    });
+
+    it('should remove listeners in LIFO order when same listener registered multiple times', () => {
```
ishymko: This test still passes if `pop()` is replaced with `shift()` in the code, so it doesn't ensure LIFO order.

However this LIFO behavior isn't mentioned in [Node docs for `EventEmitter`](https://nodejs.org/docs/latest-v18.x/api/events.html#emitterremovelistenereventname-listener):

> `removeListener()` will remove, at most, one instance of a listener from the listener array

I believe we should remove mentions of LIFO order, also I doubt we can distinguish it from the outside for same `listener` instance.

test/server/execution_event_bus.spec.ts
ishymko, line 1: `EventTarget` does this for us, but let's still add tests to ensure that errors from listeners don't propagate outside to ensure this behavior stays even if we change internals.

src/server/events/execution_event_bus.ts
drew-foxall: Made this change 👍

src/server/events/execution_event_bus.ts
drew-foxall: Made this change 👍

test/server/execution_event_bus.spec.ts
drew-foxall: Made this change 👍

test/server/execution_event_bus.spec.ts
drew-foxall: I've added the edge tests but didn't add tests here just because in Node EventListener EventTarget errors aren’t swallowed in Node; they’re surfaced as uncaught exceptions (async), while other listeners still run.

ishymko: > I've added the edge tests

Nice! 

UPD: I actually experimented around it locally and tests pass with Node.js APIs imported :( Seems like this emulation rather targets adding Edge APIs not supressing Node ones (see https://github.com/vitest-dev/vitest/issues/7699). I'll have another look separately on how this testing can be improved further, thanks for adding those either way and let's keep them!

> they’re surfaced as uncaught exceptions (async), while other listeners still run

Yeah sorry that's what I meant, a test to ensure other listeners still run if one throws. Just in case we replace `EventTarget` with something else (or a manual implementation), to make sure there is a test validating this behaviour.

vitest.edge.config.ts
```diff
@@ -0,0 +1,25 @@
...
+    environment: 'edge-runtime',
+    include: ['test/**/*.spec.ts'],
+    exclude: [
+      // Express tests require Node.js-specific APIs (http, Express framework)
+      'test/server/a2a_express_app.spec.ts',
+      // Node modules should always be excluded
+      '**/node_modules/**',
+    ],
+  },
+});
```
ishymko: nit: How about

```suggestion
export default mergeConfig(defaultConfig, {
  test: {
    environment: 'edge-runtime',
    exclude: [
      // Express tests require Node.js-specific APIs (http, Express framework)
      'test/server/a2a_express_app.spec.ts',
      // Node modules should always be excluded
      '**/node_modules/**',
    ],
  },
});
```

and also imports are going to be

```ts
import { mergeConfig } from 'vitest/config';
import defaultConfig from './vitest.config';
```

Currently we don't have a lot to inherit but just to avoid changing two places in the future if we have more.

ishymko (approved):
Last minors and should be good to go 🚀 , thank you for your contribution!

vitest.edge.config.ts
drew-foxall: Ah I complete forgot about mergeConfig. Added 👍

drew-foxall: And I've added a test that should cover the same error handling behaviour 🤞 

Looking forward to building on top of this all ^-^

ishymko (approved):

PR #223, opened 2025-12-02
feat: implement client interceptors

ishymko:
# Description

Implement client interceptors and export multi-transport client types to prepare for upcoming release. 

Interceptors support "early return" both before and after transport invocation. For early return before transport invocation  client invokes appropriate "after" methods of executed interceptors.

Release-As: 0.3.6

### DIFF
--- a/src/client/interceptors.ts
+++ b/src/client/interceptors.ts
@@ -17,6 +17,7 @@ export interface CallInterceptor {
 export interface BeforeArgs<K extends keyof Client = keyof Client> {
   /**
    * Identifies the client method invoked and its payload.
+   * Payload inside the input object can be modified.
    */
   readonly input: ClientCallInput<K>;
 
@@ -35,6 +36,7 @@ export interface BeforeArgs<K extends keyof Client = keyof Client> {
 export interface AfterArgs<K extends keyof Client = keyof Client> {
   /**
    * Identifies the client method invoked and its result.
+   * Payload inside the result object can be modified.
    */
   readonly result: ClientCallResult<K>;
 


src/client/interceptors.ts
```diff
@@ -0,0 +1,119 @@
...
+// Types below are helper types and are not exported to allow simplifying it without affecting
+// public API if necessary. They are exported via type aliases ClientXxx which can be replaced with explicit union if necessary.
+
+/**
+ * For
+ *
+ * interface Foo {
+ *   f1(): Promise<Result1>;
+ *   f2(): Promise<Result2>;
+ * }
```
guglielmo-san, line 71: parameters are missing in this example
```suggestion
 * interface Foo {
 *   f1(arg: string): Promise<Result1>;
 *   f2(arg: number): Promise<Result2>;
 * }
```

src/client/interceptors.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface CallInterceptor {
+  /**
+   * Invoked before transport method.
+   */
+  before(options: BeforeArgs): Promise<void>;
+
+  /**
+   * Invoked after transport method.
+   */
+  after(options: AfterArgs): Promise<void>;
```
guglielmo-san: why not

```suggestion
  before(options: BeforeArgs): Promise<BeforeArgs>;

  /**
   * Invoked after transport method.
   */
  after(options: AfterArgs): Promise<AfterArgs>;
```
In this way you have the flexibility to modify the input and output payload in the interceptor but retunrning it as a new object

ishymko: I was thinking about it, but not all fields in `Args` should be changed, i.e. method name inside `ClientCallXxx`, so allowing creating a brand new object gives a wrong message about what can be changed and what can not. Currently it is defined via `readonly` on this object passed by reference.

You still can modify payload in both cases completely, as `value` in `ClientCallXxx` is not readonly.

src/client/interceptors.ts
```diff
@@ -0,0 +1,121 @@
...
+   * Identifies the client method invoked and its payload.
+   * Payload inside the input object can be modified.
+   */
+  readonly input: ClientCallInput<K>;
+
+  /**
+   * If set by the interceptor, stops execution, invokes "after"
+   * for executed interceptors and returns the result. Transport is not called.
+   */
+  earlyReturn?: ClientCallResult<K>;
```
yarolegovich: why not return it as a before / after result?

PR #236, opened 2025-12-05
feat: add AgentCard to interceptors parameters

guglielmo-san:
# Description

This PR add a feature to PR #223 adding the agentCard to the interceptors' parameters. Some values of the agentCard may be necessary at the interceptors level (e.g. security schemes).

BEGIN_COMMIT_OVERRIDE
refactor: add AgentCard to interceptors parameters
END_COMMIT_OVERRIDE

### DIFF
--- a/src/server/authentication/user.ts
+++ b/src/server/authentication/user.ts
@@ -1,8 +1,21 @@
+/**
+ * Represents a user accessing A2A server.
+ */
 export interface User {
+  /**
+   * Indicates whether the user is authenticated.
+   */
   get isAuthenticated(): boolean;
+
+  /**
+   * A unique name (identifier) for the user.
+   */
   get userName(): string;
 }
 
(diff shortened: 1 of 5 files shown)


src/client/interceptors.ts
```diff
@@ -21,6 +22,11 @@ export interface BeforeArgs<K extends keyof Client = keyof Client> {
    */
   readonly input: ClientCallInput<K>;
 
+  /**
+   * Identifies the agent card cached on the client
+   */
+  readonly agentCard?: AgentCard;
```
ishymko: Can it be a required field?

```suggestion
  readonly agentCard: AgentCard;
```

guglielmo-san: Yes it could, but why making it required?
I don't think it is strictly needed for an interceptor

PR #244, opened 2025-12-09
refactor: reverse execution order of interceptors in the after interceptor

guglielmo-san:
# Description

This PR reverses the execution order of interceptors in the interceptAfter method. It ensures symmetry with interceptBefore, meaning the last interceptor called during the 'before' phase is the first one called during the 'after' phase.

### DIFF
--- a/.github/workflows/run-tck.yaml
+++ b/.github/workflows/run-tck.yaml
@@ -10,11 +10,14 @@ on:
 env:
   # TODO once we have the TCK for 0.4.0 we will need to look at the branch to decide which TCK version to run.
   # Tag of the TCK
-  TCK_VERSION: 0.3.0.beta3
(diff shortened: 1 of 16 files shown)


src/client/multitransport-client.ts
```diff
@@ -382,10 +382,8 @@ export class Client {
     args: AfterArgs<K>,
     interceptors?: CallInterceptor[]
   ): Promise<void> {
-    if (!this.config?.interceptors || this.config.interceptors.length === 0) {
-      return;
-    }
-    for (const interceptor of interceptors ?? this.config.interceptors) {
+    const reversedInterceptors = [...(interceptors || this.config?.interceptors || [])].reverse();
```
ishymko: nit: I believe it's functionally equivalent here, but maybe `??` feels more common in this context?

```suggestion
    const reversedInterceptors = [...(interceptors ?? this.config?.interceptors ?? [])].reverse();
```

PR #258, opened 2025-12-15
feat: add rest client

### LINKED ISSUE #137: [Feat]: Support HTTP+JSON/REST

### LINKED ISSUE #175: [Task]: Extract SSE logic from JSON-RPC client transport to share with REST implementation
See also:
- https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517661823
- https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517667781
- https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517701417

tchapacan:
# Description

Following continuation of this PR https://github.com/a2aproject/a2a-js/pull/142, this one is related to the implementation of the REST mode support for the client part of **a2a-js** sdk. Following the same pattern as other sdk (agnostic transport).

----

Fixes #137, fixes #175 🦕

Release-As: 0.3.7


src/client/transports/rest_transport.ts
```diff
@@ -0,0 +1,339 @@
...
+    params: GetTaskPushNotificationConfigParams,
+    options?: RequestOptions
+  ): Promise<TaskPushNotificationConfig> {
+    const configId = params.pushNotificationConfigId ?? '';
+    return this._sendRequest<TaskPushNotificationConfig>(
+      'GET',
+      `/v1/tasks/${encodeURIComponent(params.id)}/pushNotificationConfigs/${encodeURIComponent(configId)}`,
+      undefined,
+      options
+    );
```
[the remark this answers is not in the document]
tchapacan: fixed in latest push, this is a good catch gemini 😄

src/client/transports/rest_transport.ts
```diff
@@ -0,0 +1,339 @@
...
+  async setTaskPushNotificationConfig(
+    params: TaskPushNotificationConfig,
+    options?: RequestOptions
+  ): Promise<TaskPushNotificationConfig> {
+    return this._sendRequest<TaskPushNotificationConfig>(
+      'POST',
+      `/v1/tasks/${encodeURIComponent(params.taskId)}/pushNotificationConfigs`,
+      {
+        pushNotificationConfig: params.pushNotificationConfig,
+        push_notification_config: params.pushNotificationConfig,
```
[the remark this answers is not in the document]
tchapacan: In the end I did the opposite and removed the snake-case. I know that the server support both in REST mode from this [PR](https://github.com/a2aproject/a2a-js/pull/142) and our discussion, but according to [spec](https://a2a-protocol.org/latest/specification/#55-json-field-naming-convention) this will evolve to only camelCase. So my proposition is to keep only camelCase from a client point of view.

src/client/transports/sse_parser.ts
ishymko, line 1: Let's put it into [sse_utils.ts](https://github.com/a2aproject/a2a-js/blob/main/src/sse_utils.ts) created in the server PR and create a "symmetry" test that parser is able to understand what's generated by formatter.

src/client/transports/rest_transport.ts
ishymko, line 1: Let's have dedicated unit tests for a transport similar to the one done for the old client ([client.spec.ts](https://github.com/a2aproject/a2a-js/blob/main/test/client/client.spec.ts)). Testing all methods via e2e test is going to be problematic (should be possible though, I'll investigate it separately) but would be good to have some coverage.

We don't have one for JSON-RPC transport unfortunately as I relied on [client.spec.ts](https://github.com/a2aproject/a2a-js/blob/main/test/client/client.spec.ts) as old client was switched to use the transport as well, but I'll do it to properly deprecated old client.

src/client/transports/rest_transport.ts
```diff
@@ -0,0 +1,343 @@
...
+  INVALID_PARAMS: -32602,
+  INTERNAL_ERROR: -32603,
+  TASK_NOT_FOUND: -32001,
+  TASK_NOT_CANCELABLE: -32002,
+  PUSH_NOTIFICATION_NOT_SUPPORTED: -32003,
+  UNSUPPORTED_OPERATION: -32004,
+  CONTENT_TYPE_NOT_SUPPORTED: -32005,
+  INVALID_AGENT_RESPONSE: -32006,
+  AUTHENTICATED_EXTENDED_CARD_NOT_CONFIGURED: -32007,
+} as const;
```
ishymko: We have the same [in server](https://github.com/a2aproject/a2a-js/blob/6c76fef26473d8b15d729040b6c474de2794c7fe/src/server/transports/rest/rest_transport_handler.ts#L59-L69), let's extract into some common place. It's kind of a tech debt as we shouldn't use JSON-RPC codes for REST, but 0.3 doesn't define errors for REST properly so there is no other choice. Let's move it to [errors.ts](https://github.com/a2aproject/a2a-js/blob/main/src/errors.ts) for now without exporting for users.

test/e2e.spec.ts
```diff
@@ -30,112 +33,152 @@ class TestAgentExecutor implements AgentExecutor {
...
+      }),
+  },
+  {
+    name: 'REST',
+    preferredTransport: 'HTTP+JSON',
+    serverPath: '/a2a/rest',
+    clientFactory: () =>
+      new ClientFactory({
+        transports: [new RestTransportFactory()],
+      }),
```
ishymko: nit: can we have a single `clientFactory` here with both transports, maybe even the default instance we provide? I believe `preferredTransport` set in the agent card should work, so it'll be a bit closer to real usage.

src/client/transports/sse_parser.ts
```diff
@@ -0,0 +1,52 @@
...
+        buffer = buffer.substring(lineEndIndex + 1);
+
+        if (line === '') {
+          if (eventDataBuffer) {
+            yield { type: eventType, data: eventDataBuffer };
+            eventDataBuffer = '';
+            eventType = 'message';
+          }
+        } else if (line.startsWith('event:')) {
+          eventType = line.substring(6).trim();
```
ishymko: nit:
```suggestion
          eventType = line.substring('event:'.length).trim();
```

src/client/transports/sse_parser.ts
```diff
@@ -0,0 +1,52 @@
...
+        } else if (line.startsWith('data:')) {
+          eventDataBuffer += line.substring(5).trimStart() + '\n';
```
ishymko: nit:
```suggestion
          eventDataBuffer += line.substring('data:'.length).trimStart() + '\n';
```

src/client/transports/json_rpc_transport.ts
```diff
@@ -303,62 +304,8 @@ export class JsonRpcTransport implements Transport {
...
-    } catch (error) {
-      console.error(
-        'Error reading or parsing SSE stream:',
-        (error instanceof Error && error.message) || 'Error unknown'
-      );
-      throw error;
-    } finally {
-      reader.releaseLock();
+    for await (const event of parseSseStream(response)) {
+      yield this._processSseEventData<A2AStreamEventData>(event.data, clientRequestId);
```
ishymko, line 308: Hmm, seems like the old code here didn't handle errors (`event: error`). Could you please add the same thing you did for REST? (yeah out of scope a bit but with this refactoring the problem became obvious 😄 )

src/client/transports/sse_parser.ts
ishymko, line 1: Could you please also take a look at #175? Especially at the last one, seems like parsing can be simplified significantly.

src/client/transports/rest_transport.ts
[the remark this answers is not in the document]
ishymko: Agree, let's keep camelCase in the client.

src/client/transports/sse_parser.ts
tchapacan: I have moved the `sse_parser.ts` into the shared one at the root of the project in `src/sse_utils.ts`, agree it's way much better like this, added some tests as well

src/client/transports/rest_transport.ts
tchapacan: I have created it in `test/client/transports/rest_transport.test.ts` next to the jsonrpc one (that i saw is only testing send message currently) and i had to update the config of vitest to include `'test/**/*.test.ts'` so that it's taken into account. I took the assumption that it might look like this in the future, but as I have modified config of your tooling I can still revert it and create a spec file specific for the rest. LMK if that's ok or if you see further test I could add, I'm not sure I have covered all unhappy path, but some of it seems already done, and kind of generic in `client.spec.ts` versus the one dedicated to rest or jsonrpc (route/method etc).

test/e2e.spec.ts
tchapacan: Yes agreed! I have modified it this way, LMK if thats good. One client factory using the default config (so taking both), and we iterate other the agent card `preferredTransport`, `serverPath` etc for the e2e test

src/client/transports/sse_parser.ts
tchapacan: thanks it's fixed

src/client/transports/sse_parser.ts
tchapacan: thanks it's fixed

src/client/transports/json_rpc_transport.ts
tchapacan: oups yes I saw it and completly forgot to add it for jsonrpc lol! yes it's added and both error handling are symmetric now, thanks !

src/client/transports/sse_parser.ts
tchapacan: I took a look and tried to do a cocktail between your suggestion upper and the proposition in https://github.com/a2aproject/a2a-js/issues/175, it seems easier to understand after refacto but i'd like a double check from you to make sure it's good like this

src/client/transports/rest_transport.ts
tchapacan: I have extracted it in `errors.ts` as proposed at the root in `src/` and it seems clearer like that. I noticed something a little odd in the server part here we might have missed: https://github.com/a2aproject/a2a-js/pull/258/files#diff-09d4639007902e58571c5cb95344d3a76510f8f2c501f03bfbd3fa8cb8c34076L68 and https://github.com/a2aproject/a2a-js/pull/258/files#diff-55db69e02ff5714d444d8081ec6ecac5d9833fb29fda64d1e829e5766434fdc0R13-R14

I don't find where it was but in the spec for jsonrpc I'm not sure there is an error code for unauthenticated http 401 to map too. `CONTENT_TYPE_NOT_SUPPORTED` and `UNAUTHORIZED` seems semantically far away from each other, so I'm not sure they are sharing the same a2a error code or maybe i'm missing something.

tchapacan:
Helo @ishymko ,

2 open questions crossed my mind while re-reading the PR.

- Do we plan to add a sort of dynamic way to escape a potential `/` suffix (using a regex for example) in the endpoint in case someone add it in the agentCard that could break the full path url construction? I don't like adding too much regex, it can be error prone, but well depending of the case sometimes can be helpful. Maybe it's enforced by the spec and not a subject, but just asking as I already saw this.
- Again the spec is quite obvious about the fact that in the agentCard it's `HTTP+JSON` in uppercase and respectively for other transports. But unfortunately I have already seen custom implementation that leave it lowercase. Does it make sense to always uppercase it at runtime to ensure compatibility?

What's your opinion regarding those 2?

src/client/transports/json_rpc_transport.ts
ishymko: Oh now I understand, it used to work correctly before, because `_processSseEventData` parses the payload and tries to look for `error` or `result` there (and throws if can't find any of those). 

So it works well for both event types. My bad, this can be reverted.

src/client/transports/rest_transport.ts
ishymko: Ah great catch on `*.test.ts`, it wasn't intentional, let's rename both to `spec` and keep current pattern in vitest config.

test/client/transports/rest_transport.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    it('should throw TaskNotFoundError on -32001', async () => {
+      const messageParams = createMessageParams();
+      mockFetch.resolves(createRestErrorResponse(-32001, 'Task not found', 404));
+
+      await expect(transport.sendMessage(messageParams)).rejects.toThrow(TaskNotFoundError);
+    });
+  });
+
+  describe('getTask', () => {
```
ishymko, line 115: Let's add one with `historyLength` as well as it's passed via query.

src/sse_utils.ts
```diff
@@ -59,3 +71,69 @@ export function formatSSEEvent(event: unknown): string {
...
+        if (eventData) {
+          yield { type: eventType, data: eventData };
+          eventData = '';
+          eventType = 'message';
+        }
+      } else if (line.startsWith('event:')) {
+        eventType = line.substring('event:'.length).trim();
+      } else if (line.startsWith('data:')) {
+        // Expect well-formed JSON on a single data line
+        eventData = line.substring('data:'.length).trimStart();
```
ishymko: Can we have 

```suggestion
        eventData = line.substring('data:'.length).trim();
```

Seems like original implementation did the same, but I am not sure why, should be fine trimming both ends.

test/sse_utils.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+ * Used to test that the parser can understand what the formatter produces.
+ */
+function createMockResponse(sseData: string): Response {
+  const encoder = new TextEncoder();
+  const stream = new ReadableStream({
+    start(controller) {
+      controller.enqueue(encoder.encode(sseData));
+      controller.close();
+    },
+  });
```
ishymko, line 26: nit: I'd split `sseData` into some small 1-2 chars chunks to better "stress test" buffering and concatenation.

src/client/transports/rest_transport.ts
ishymko: I believe it's a typo from [here](https://github.com/a2aproject/a2a-js/pull/142/changes#diff-e33da4df125271ea190143c721b702b6b17c04af0bd55f03d906bc0a3380691aR68), there is no unauthorized code in [the spec](https://a2a-protocol.org/v0.3.0/specification/#82-a2a-specific-errors). `-32005` is `ContentTypeNotSupportedError`, unauthorized should be removed.

src/sse_utils.ts
tchapacan: thanks it's fixed !

test/sse_utils.spec.ts
tchapacan: in the function that generate the mock i have split the data by 2 char, lmk if it's ok like this

test/client/transports/rest_transport.spec.ts
tchapacan: thanks it's fixed, test added

src/client/transports/json_rpc_transport.ts
tchapacan: yes that's true! i have reverted it, as the naming is the same i was expecting the behaviour too

src/client/transports/rest_transport.ts
tchapacan: Thanks for the confirmation, yes I have removed it completly, and it's a proper and clean http 401 that will be return without mapping

src/client/transports/rest_transport.ts
tchapacan: renamed both and reverted the config of vitest

PR #279, opened 2025-12-18
feat: implement gRPC server

### LINKED ISSUE #275: [Feat]: Add support for gRpc

_No response_

## **Background**

The A2A TypeScript SDK requires support for the gRPC transport protocol. This entails a Node.js server and a client implementation that is compatible with both Node.js and browser environments. This document evaluates available architectural approaches and proposes a recommended implementation strategy.

## **gRPC Transport Strategy**

To effectively utilize generated types within the gRPC protocol, a specific transport runtime library is required. The following libraries are the industry-standard options for our target environments:

* **Node.js Environment:** The recommended choice is [**`grpc-node`](https://github.com/grpc/grpc-node)** (specifically the [`@grpc/grpc-js`](https://github.com/grpc/grpc-node/tree/master/packages/grpc-js) package). This provides a pure JavaScript implementation of gRPC for Node.js.  
* **Browser Environment:** The recommended choice is [**`grpc-web`](https://github.com/grpc/grpc-web)**. Because browsers cannot handle the HTTP/2 trailers and raw TCP sockets required by standard gRPC, `grpc-web` allows the browser client to communicate with services via a proxy (typically **Envoy**) that translates the traffic.

## **Code Generation**

To implement Protobuf support, the first step requires generating .ts files from the .proto.

### **Static vs Dynamic Generation**

An architectural decision for gRPC in TypeScript is the choice between static and dynamic code generation:

* Dynamic Generation: loads `.proto` files at runtime. While flexible, it lacks strict type definitions, effectively negating the benefits of using TypeScript.  
* Static Generation: generates client and server code ahead of time.

For our use case, static code generation is preferable. By generating strict TypeScript definitions from our `.proto` files, we ensure robust compile-time type checking. This approach improves the developer experience by providing IntelliSense/autocompletion and catching errors during development rather than at runtime.

### **Plugins for Code Generation**

Different plugins have been evaluated to generate the typescript code for Message and Service from the .proto files.

#### **1\. [ts-proto](https://github.com/stephenh/ts-proto)**

**Pros:**

* **Idiomatic TypeScript:** It generates Plain Old JavaScript Objects (POJOs) and Interfaces.  
* **Smallest Bundle Size:** Relies on minimal runtime code. It avoids importing massive libraries to parse messages and often generates a single, concise `.ts` file.  
* **Simple gRPC Support:** It supports gRPC on Node environments natively.  
* **Customization**: Offers extensive options to customize output.

**Cons:**

* **gRPC Web Integration:** Native browser support requires an additional third-party library ([`nice-grpc-web`](https://github.com/deeplay-io/nice-grpc)).  
* **Conformance Test:** Does not perform well on Protobuf Conformance test compared to other options

**Dependencies:** 

* ts-proto: for Messages and Services generation  
* @grpc/grpc-js: for Server side implementation  
* nice-grpc-web: for support of browser gRPC

#### **2\. [protobuf-ts](https://github.com/timostamm/protobuf-ts)**

**Pros:**

* **Excellent gRPC-Web Transports:** It ships with its own robust `GrpcWebTransport (`[@protobuf-ts/grpcweb-transport](https://github.com/timostamm/protobuf-ts/blob/main/packages/example-browser-grpcweb-transport-client/README.md)). It handles interceptors (middleware), abort signals (canceling requests), and metadata handling very well out of the box.  
* **Idiomatic TypeScript:** Like `ts-proto`, it generates clean POJOs and Interfaces.  
* **Standards Compliant:** It passes the official Protocol Buffer conformance tests.

**Cons:**

* **Larger Bundle than ts-proto:** Because it includes a robust runtime (for handling the transports and types), it adds more kilobytes to your final application bundle than `ts-proto`.  
* **Naming convention:** uses C\# naming convention, appending an \`I\` at the beginning of Interfaces names.  
* **Import format**: this library does not support ESM import format in the generated files. This would require using another library to modify the generated imports.

**Dependencies:**

* protobuf-ts: for Messages and Services generation  
* @grpc/grpc-js: for Server side implementation  
* @protobuf-ts/grpcweb-transport: for support of browser gRPC

#### **3\. [ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen)**

**Pros:**

* **Plain Integration:** it supports integration with gRPC and gRPC-web.

**Cons:**

* **Difficult Developer Experience:** It generates "Java-style" code for Messages, with getters/setters.  
* **Messy Output:** This plugin generates both the *types* (`.d.ts`) and the *JavaScript* (`.js`)  
* **Bloated Dependencies:** It requires the `google-protobuf` runtime library, which is very large and does not tree-shake well.

**Dependencies:**

* protoc-gen-ts: for Messages and Services generation  
* @grpc/grpc-js: for Server side implementation  
* Google-protobuf

#### **4\. p[rotoc-gen-grpc-web](https://github.com/grpc/grpc-web#code-generator-plugin)**

**Pros:**

* **Official library:** this is the official plugin library developed for Typescript by grpc-Web

**Cons:**

* **Experimental:** it is still defined “experimental” support in the documentation  
* **Only gRpr-web**: this plugin can generate code that will work only with gRpc-web on Browser 

## **Alternative Approach: Connect Protocol**

Instead of relying on the legacy `@grpc/grpc-js library` and `grpc-web` stack, a modern alternative is to adopt the **Connect** ecosystem (by Buf), using [**Protobuf-ES**](https://github.com/bufbuild/protobuf-es) for code generation and [**Connect-ES**](https://github.com/connectrpc/connect-es) as the transport library. Connect-ES does not use `@grpc/grpc-js`. Instead, it runs on top of the standard, built-in Node.js `http2` modules.

**Pros**:

* **No Envoy Proxy Required:** The "Connect Protocol" allows the browser client to talk directly to the Node.js server. The server (implemented with [`@connectgrpc/connect-node`](https://github.com/connectrpc/connect-es/tree/main/packages/connect-node)) automatically handles the translation of incoming gRPC-Web or Connect requests. This removes the need for a complex Envoy proxy to translate protocols.  
* [**Standards Compliant**](https://buf.build/blog/protobuf-conformance)**:** Protobuf-ES generated code passes the official Protocol Buffer conformance tests.  
* **Universal Handlers:** The ConnectRPC server can work as a simple route handler in standard Node frameworks.  
* **Browser Native:** The client uses the standard `fetch` API. It is extremely lightweight and creates standard network requests that are visible and debuggable in the Chrome/Firefox DevTools "Network" tab.  
* **Multi-Protocol Support:** A single server definition automatically supports three protocols simultaneously:  
  1. **gRPC** (for backend-to-backend).  
  2. **gRPC-Web** (for legacy browser clients).  
  3. **Connect** (for modern browser clients and easy debugging).

**Cons**:

* **Newer Ecosystem:** As it represents a newer ecosystem, it has fewer use cases.  
* **"Pure" gRPC Distinction:** Browser communication is technically strictly HTTP/1.1 or HTTP/2 POST requests rather than raw TCP gRPC.

## **Summary**

| Generator plugin | Node.js Runtime (Server/Client) | Browser Runtime (gRPC-Web) | Proxy Required? |
| :---- | :---- | :---- | :---- |
| ts-proto | @grpc/grpc-js | nice-grpc-web | Yes (Envoy) |
| protobuf-ts | @grpc/grpc-js | @protobuf-ts/grpcweb-transport  | Yes (Envoy) |
| ts-protoc-gen | @grpc/grpc-js | @improbable-eng/grpc-web | Yes (Envoy) |
| protoc-gen-grpc-web | n/a | grpc-web | Yes (Envoy) |
| protoc-gen-es | @connectrpc/connect-node | @connectrpc/connect-web | No (if both client and server use Connect) |

#### **1\. Node.js Environment**

The Node.js implementation will utilize **`ts-proto`** in conjunction with the **`@grpc/grpc-js`** runtime library. This choice is due to its high customization (generates import with .js suffix) and its ability to generate idiomatic, type-safe TypeScript interfaces.

**2\. Browser Environment: gRPC-Web Integration**

To extend SDK functionality to the browser, the project will incorporate the official **`protoc-gen-grpc-web`** generator. This ensures the SDK adheres to the formal gRPC-Web specification, maintaining strict alignment with Google’s infrastructure standards. While this necessitates a proxy layer (e.g., Envoy) for browser-to-server communication, it preserves the protocol’s integrity.

#### **3\. Justification**

**Connect** protocol was ultimately dismissed to maintain a "pure gRPC" ecosystem. While it offers simplified browser transport, it deviates from the standard gRPC-Web framing.  
While **protobuf-ts** is a capable alternative, it required post-processing of generated files to ensure compliance with ESM (ECMAScript Modules) import standards. Avoiding this additional complexity reduces the maintenance work of the SDK.  
**ts-protoc-gen** was excluded due to its reliance on legacy code patterns. The tool produces a verbose output utilizing manual getters and setters, which results in a suboptimal TypeScript experience.  

_No response_

### Additional context

_No response_

### Code of Conduct

guglielmo-san:
# Description

This Pull Request implements server-side support for the gRPC transport protocol.

**Schema and Generation**
The implementation is based on [A2A Proto Specification v3.0.0](https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/grpc/a2a.proto).

**Tooling**
ts-proto was adopted for generating TypeScript service definitions (with buf), which will reuse the message types previously generated (this requires a manual modification of the generated file).

**Limitations**
Due to the current dependency on Proto v3.0.0, the gRPC implementation is missing some APIs already available for other transports, for example the `getExtendedAuthenticatedAgentCard` method.
These missing endpoints will be implemented with future official releases of proto specification.

**Entry Points**
A new dedicated entry point has been introduced to export the gRPC server logic and  @grpc/grpc-js and its related transport libraries have been added as peer dependencies.

Fixes #275  🦕

BEGIN_COMMIT_OVERRIDE
feat: implement gRPC server (#279)

Release-As: 0.3.9
END_COMMIT_OVERRIDE

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,9 @@
       "server/express": [
         "./dist/server/express/index.d.ts"
       ],
+      "server/grpc": [
+        "./dist/server/grpc/index.d.ts"
+      ],
       "client": [
         "./dist/client/index.d.ts"
       ]


src/grpc/utils/to_proto.ts
```diff
@@ -0,0 +1,615 @@
...
+    };
+  }
+
+  static listTaskPushNotificationConfigs(
+    config: types.TaskPushNotificationConfig[]
+  ): ListTaskPushNotificationConfigResponse {
+    return {
+      configs: config.map((c) => this.taskPushNotificationConfig(c)),
+      nextPageToken: '',
+    };
```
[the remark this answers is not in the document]
guglielmo-san: we should update types for 1.0

package.json
```diff
@@ -71,10 +73,12 @@
     "format:readme": "prettier --write ./README.md",
     "lint:fix": "npx eslint . --fix",
     "coverage": "vitest run --coverage",
+    "ts-proto": "curl -o ./src/grpc/a2a.proto https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/grpc/a2a.proto && protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_out=./src/grpc --ts_proto_opt=oneof=unions-value,importSuffix=.js,outputServices=grpc-js,env=node,esModuleInterop=true --proto_path=./src/grpc ./src/grpc/a2a.proto -I ./node_modules/google-proto-files && rm ./src/grpc/a2a.proto",
```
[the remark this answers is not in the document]
guglielmo-san: I will change it to use buf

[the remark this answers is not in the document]
guglielmo-san: added buf generation

src/grpc/utils/from_proto.ts
```diff
@@ -0,0 +1,185 @@
...
+  ): types.DeleteTaskPushNotificationConfigParams {
+    const name = request.name;
+    return {
+      id: extractTaskId(name),
+      pushNotificationConfigId: extractPushNotificationConfigId(name),
+    };
+  }
+
+  static message(message: Message): types.Message | undefined {
+    if (!message) {
```
ishymko: Why do we have undefined checks only in some methods? I believe it makes sense to be consistent: either push those checks to callers everywhere or do them here everywhere.

src/grpc/utils/from_proto.ts
```diff
@@ -0,0 +1,185 @@
...
+      return undefined;
+    }
+
+    return {
+      kind: 'message',
+      messageId: message.messageId,
+      parts: message.content.map((p) => FromProto.parts(p)),
+      contextId: message.contextId,
+      taskId: message.taskId,
+      role: message.role === Role.ROLE_AGENT ? 'agent' : 'user',
```
ishymko: nit: maybe a function with explicit switch and throw for unknown values?

src/grpc/utils/id_decoding.ts
```diff
@@ -0,0 +1,28 @@
+import { A2AError } from '../../server/error.js';
+
+const TASK_ID_REGEX = /tasks\/([^/]+)/;
+const CONFIG_ID_REGEX = /pushNotificationConfigs\/([^/]+)/;
```
ishymko: I believe it makes sense to match the entire line via `^` and `$` to avoid parsing strings like `tasks/a/b/c/d` into `a`.

src/grpc/utils/to_proto.ts
```diff
@@ -0,0 +1,463 @@
...
+      capabilities: ToProto.agentCapabilities(agentCard.capabilities),
+      securitySchemes: agentCard.securitySchemes
+        ? Object.fromEntries(
+            Object.entries(agentCard.securitySchemes).map(([key, value]) => [
+              key,
+              ToProto.securityScheme(value),
+            ])
+          )
+        : {},
+      security: agentCard.security ? agentCard.security.map((s) => ToProto.security(s)) : [],
```
ishymko: nit: here and below


```suggestion
      security: agentCard.security?.map((s) => ToProto.security(s)) ?? [],
```

src/samples/package.json
```diff
@@ -22,4 +22,4 @@
   "dependencies": {
     "uuid": "^11.1.0"
   }
-}
+}
```
ishymko: Revert?

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+import { A2AError } from '../error.js';
+
+/**
+ * Options for configuring the gRPC handler.
+ */
+export interface grpcHandlerOptions {
+  requestHandler: A2ARequestHandler;
+  userBuilder: (
+    call: grpc.ServerUnaryCall<unknown, unknown> | grpc.ServerWritableStream<unknown, unknown>
+  ) => Promise<User | undefined>;
```
ishymko: Can we have it as a type like we did for Express? 

Also why `Promise<User | undefined>`? I'd say that as long as you provide this callback it should always return a used (possibly unauthenticated one).

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+import { ServerCallContext } from '../context.js';
+import { Extensions } from '../../extensions.js';
+import { User, UnauthenticatedUser } from '../authentication/user.js';
+import { HTTP_EXTENSION_HEADER } from '../../constants.js';
+import { A2AError } from '../error.js';
+
+/**
+ * Options for configuring the gRPC handler.
+ */
+export interface grpcHandlerOptions {
```
ishymko: Should be `GrpcHandlerOptions`, types start with a capital letter and we have the same for `Json` and `Rest`.

src/server/transports/grpc/grpc_transport_handler.ts
```diff
@@ -0,0 +1,159 @@
...
+  TaskPushNotificationConfig,
+  TaskQueryParams,
+  TaskIdParams,
+  AgentCard,
+  DeleteTaskPushNotificationConfigParams,
+  ListTaskPushNotificationConfigParams,
+  GetTaskPushNotificationConfigParams,
+} from '../../../types.js';
+
+export class gRpcTransportHandler {
```
ishymko:
```suggestion
export class GrpcTransportHandler {
```

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Creates a gRPC transport handler.
+ * This handler implements the A2A gRPC service definition and acts as an
+ * adapter between the gRPC transport layer and the core A2A request handler.
+ *
+ * @param requestHandler - The core A2A request handler for business logic.
+ * @returns An object that implements the A2AServiceServer interface.
+ */
+export function grpcHandler(options: grpcHandlerOptions): A2AServiceServer {
```
ishymko: I'd define it as a class right away, the reason why it was defined like this for Express was to match Express API which operates on middlewares defined as functions. While here the usage is

```ts
const server = new Server();
server.addService(A2AServiceService, grpcHandlerInstance);
```

which operates on services defined as objects.

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+    details: a2aError.message,
+  };
+};
+
+const buildContext = async (
+  call: grpc.ServerUnaryCall<unknown, unknown> | grpc.ServerWritableStream<unknown, unknown>,
+  userBuilder: grpcHandlerOptions['userBuilder']
+): Promise<ServerCallContext> => {
+  const user = await userBuilder(call);
+  const extensionHeaders = call.metadata.get(HTTP_EXTENSION_HEADER.toLowerCase());
```
ishymko: nit: based on `.get` comment the key is normalzied: https://github.com/grpc/grpc-node/blob/164d14f28261ad1fbbc99d076fe850dad018b2bb/packages/grpc-js/src/metadata.ts#L44-L46

So `toLowerCase()` shouldn't be needed:
```suggestion
  const extensionHeaders = call.metadata.get(HTTP_EXTENSION_HEADER);
```

src/server/grpc/grpc_service.ts
```diff
@@ -0,0 +1,273 @@
...
+}
+
+/**
+ * Creates a gRPC transport handler.
+ * This handler implements the A2A gRPC service definition and acts as an
+ * adapter between the gRPC transport layer and the core A2A request handler.
+ *
+ * @param requestHandler - The core A2A request handler for business logic.
+ * @returns An object that implements the A2AServiceServer interface.
+ */
```
ishymko, line 52: Let's add a usage example with `@example` tag.

vitest.edge.config.ts
```diff
@@ -8,6 +8,9 @@ export default defineWorkersConfig(
       exclude: [
         // Express tests require Node.js-specific APIs (http, Express framework)
         'test/server/a2a_express_app.spec.ts',
+        // gRpc test require Node.js-specific gRPC module
+        'test/server/grpc_handler.spec.ts',
+        'test/utils/*.spec.ts',
```
ishymko: Utils folder name is too generic to exlude here. We can reflect source structure in tests: `test/server/grpc`, put all GRPC tests here and exclude it for Edge.

package.json
```diff
@@ -72,17 +80,26 @@
...
     "coverage": "vitest run --coverage",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",
+    "generate-grpc-types": "cd ./src/grpc && curl -o ./a2a.proto https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/grpc/a2a.proto && buf generate && rm ./a2a.proto",
     "test-build": "esbuild ./dist/client/index.js ./dist/server/index.js ./dist/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist"
   },
   "dependencies": {
     "uuid": "^11.1.0"
   },
   "peerDependencies": {
+    "@bufbuild/protobuf": "^2.10.2",
```
ishymko, line 105: Should it be a peer dependency (or maybe a dependency at all) with `stephenh-ts-proto` being used? I thought that this runtime support this is only required if `protoc-gen-es` is used.

src/types.ts
```diff
@@ -1764,7 +1764,7 @@ export interface Message1 {
  * The current status of the task, including its state and a descriptive message.
  */
 export interface TaskStatus {
-  message?: Message2;
+  message?: Message;
```
ishymko: Is this necessary?

src/server/grpc/grpc_handler.ts
guglielmo-san: Yes correct, i forgot to update this

src/server/grpc/grpc_handler.ts
guglielmo-san: I decided not to implement it as a class, as the `A2AServiceService` extends `UntypedServiceImplementation`. This requires the class to be indexed by string, and then to add `[name: string]: any;` in the class definition, which i did not like to add, this is why I decided to implement it as a function.

package.json
guglielmo-san: `@bufbuild/protobuf` is used to serialize and deserialize and is imported in `a2a.ts`. I added it to the document

src/server/grpc/grpc_handler.ts
guglielmo-san: I decided not to implement it as a class, as the `A2AServiceService` extends `UntypedServiceImplementation`. This requires the class to be indexed by string, and then to add `[name: string]: any;` in the class definition, which i did not like to add, this is why I decided to implement it as a function.

package.json
guglielmo-san: It is necessary for the Serialization and Deserialization in the `a2a.ts` file. I added it in to the document

src/types.ts
guglielmo-san: I temporsrily changed it as it was complaining in the proto to types conversion and forgot to review it again. I changed it back and it seems to be working now

src/grpc/utils/id_decoding.ts
guglielmo-san: in some cases we could have /** name=tasks/{id}/pushNotificationConfigs/{push_id} */
I updated the regex for the case of having only the task_id, and the case for both

src/grpc/utils/from_proto.ts
guglielmo-san: I have added only for the types that can be undefined from the proto generation

tck/agent/index.ts
```diff
@@ -16,6 +17,8 @@ import {
   restHandler,
   UserBuilder,
 } from '../../src/server/express/index.js';
+import { grpcHandler } from '../../src/server/grpc/grpc_handler.js';
+import { A2AServiceService } from '../../src/grpc/a2a_services.js';
```
ishymko: We should use `index.js` imports here to make we export all we need to use it.

I believe in this case `A2AServiceService` is going to be a problem.

src/server/grpc/index.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * gRPC integration for the A2A Server library.
+ * This module provides gRPC specific functionality.
+ */
+
```
ishymko, line 5: Should `UserBuilder` be here as well?

src/grpc/buf.gen.yaml
```diff
@@ -0,0 +1,10 @@
+# buf generate
+# Configuration for the buf generate command
+# Uses remote plugins, no separate install required.
+version: v2
+inputs:
+  - directory: .
+plugins:
+  - remote: buf.build/community/stephenh-ts-proto:v2.6.1
+    opt: oneof=unions-value,importSuffix=.js,outputServices=grpc-js,env=node,esModuleInterop=true,outputJsonMethods=false,outputPartialMethods=false
```
ishymko: According to [the docs](https://buf.build/docs/configuration/v2/buf-gen-yaml/#opt), it can be arranged like
```suggestion
    opt:
      - oneof=unions-value
      - importSuffix=.js
      - outputServices=grpc-js
      - env=node
      - esModuleInterop=true
      - outputJsonMethods=false
      - outputPartialMethods=false
```

src/grpc/pb/a2a_services.ts
```diff
@@ -0,0 +1,3626 @@
...
+ * - Tasks have a get interface and custom cancel and subscribe methods.
+ * - TaskPushNotificationConfig are a resource whose parent is a task.
+ *   They have get, list and create methods.
+ * - AgentCard is a static resource with only a get method.
+ * fields are not present as they don't comply with AIP rules, and the
+ * optional history_length on the get task method is not present as it also
+ * violates AIP-127 and AIP-131.
+ */
+export type A2AServiceService = typeof A2AServiceService;
+export const A2AServiceService = {
```
ishymko, line 3276: Can this name be customized to avoid `ServiceService`?

src/server/grpc/common.ts
```diff
@@ -0,0 +1,10 @@
+import * as grpc from '@grpc/grpc-js';
```
ishymko: Can we do `import type` only for what we use here? It seems like we don't import any code, and `import type` is not emitted in JS.

src/grpc/buf.yaml
```diff
@@ -0,0 +1,6 @@
+version: v2
+modules:
+  - path: .
```
ishymko: Maybe under `pb` so that it's `grpc/pb` similar to `types`?

package.json
```diff
@@ -72,17 +80,26 @@
     "lint:fix": "npx eslint . --fix",
     "coverage": "vitest run --coverage",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",
+    "generate-grpc-services": "cd ./src/grpc && curl -o ./a2a_services.proto https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/grpc/a2a.proto && buf generate && rm ./a2a_services.proto",
```
ishymko: Doesn't it require manual postprocessing now to remove duplicated types?

src/server/transports/grpc/grpc_transport_handler.ts
```diff
@@ -0,0 +1,159 @@
...
+  async getAgentCard(): Promise<AgentCard> {
+    return this.requestHandler.getAgentCard();
+  }
+
+  /**
+   * Gets the authenticated extended agent card.
+   */
+  async getAuthenticatedExtendedAgentCard(context: ServerCallContext): Promise<AgentCard> {
+    return this.requestHandler.getAuthenticatedExtendedAgentCard(context);
+  }
```
ishymko: Looks like `getAuthenticatedExtendedAgentCard` is unused and `getAgentCard` is called from `grpc_handler.ts`.

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+        FromProto.taskIdParams,
+        grpcTransportHandler.cancelTask.bind(grpcTransportHandler),
+        ToProto.task
+      );
+    },
+    async getAgentCard(
+      call: grpc.ServerUnaryCall<GetAgentCardRequest, AgentCard>,
+      callback: grpc.sendUnaryData<AgentCard>
+    ): Promise<void> {
+      return await wrapUnary(
```
ishymko: nit: `await` is redundant here and there are no `await`s in other methods. I also believe that `async` can be removed everywhere as you don't use `await`s, it just passes over promises.

src/server/grpc/grpc_service.ts
```diff
@@ -0,0 +1,273 @@
...
+          const response = ToProto.messageStreamResult(responsePart);
+          call.write(response);
+        }
+      } catch (error) {
+        call.emit('error', mapToError(error));
+      } finally {
+        call.end();
+      }
+    },
+
```
ishymko, line 138: nit: please add line breaks between other methods in this file. I was hoping to have a linter option for this, but wasn't able to find one :(

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,273 @@
...
+      return wrapUnary(
+        call,
+        callback,
+        FromProto.messageSendParams,
+        grpcTransportHandler.sendMessage.bind(grpcTransportHandler),
+        ToProto.messageSendResult
+      );
+    },
+
+    async sendStreamingMessage(
```
ishymko: The code looks very similar to `taskSubscription`, maybe it can be deduplicated?

src/types.ts
ishymko: Is it required?

vitest.edge.config.ts
ishymko, line 1: Please update non-edge vitest config to exclude generated code from coverage.

src/grpc/buf.gen.yaml
guglielmo-san: I will remove it anyway as it requires postprocessing

src/grpc/buf.yaml
guglielmo-san: I removed the buf generation as it requires post processing

package.json
guglielmo-san: Removed

src/server/transports/grpc/grpc_transport_handler.ts
guglielmo-san: We will use it once we update the 1.0.o, I don't think it is a big issue to leave it here unused for now

src/types.ts
guglielmo-san: it should have be reverted already

src/grpc/pb/a2a_services.ts
guglielmo-san: I exported it as A2AService in index.js

src/server/grpc/grpc_service.ts
ishymko, line 1: nit: rename the file to `grpc_service.ts` or just `service.ts` as we're already in the `grpc` folder?

src/server/grpc/grpc_handler.ts
```diff
@@ -0,0 +1,292 @@
...
+};
+
+const mapToError = (error: unknown): Partial<grpc.ServiceError> => {
+  const a2aError =
+    error instanceof A2AError
+      ? error
+      : A2AError.internalError(error instanceof Error ? error.message : 'Unknown Error');
+
+  const metadata = new grpc.Metadata();
+  metadata.set('a2a-error-code', a2aError.code.toString());
```
ishymko: This metadata key is not a part of the spec. The best we can do right now (until we have 1.0 where error mapping is properly specified for GRPC) is to set status codes the same way as other SDKs.

vitest.config.ts
```diff
@@ -9,7 +9,7 @@ export default defineConfig({
       provider: 'v8',
       reporter: ['text', 'json', 'html', 'json-summary'],
       include: ['src/**/*'],
-      exclude: ['src/samples/**/*', 'src/types/pb/**/*'],
+      exclude: ['src/samples/**/*', 'src/types/pb/**/*', 'src/grpc/**/*'],
```
ishymko: Maybe
```suggestion
      exclude: ['src/samples/**/*', 'src/types/pb/**/*', 'src/grpc/pb/**/*'],
```

guglielmo-san: Added

package.json
ishymko, line 1: #288 was merged while this was WIP, please sync with main add GRPC entry point similar to others.

PR #292, opened 2026-01-09
fix!: use ProtoJSON for REST transport

### LINKED ISSUE #291: HTTP+JSON payload shape mismatches proto‑JSON (ADK/Agent Engine 400s)
### What happened?

# Bug Report / Feature Request: HTTP+JSON Payload Shape

Reference PR in gemini-cli: https://github.com/google-gemini/gemini-cli/pull/16013

## Summary

HTTP+JSON requests from `@a2a-js/sdk@0.3.7` are rejected by ADK/Agent Engine
unless the client rewrites request/response payloads. The SDK emits a legacy
JSON shape that does not match the spec’s proto-JSON binding.

## Environment

* `@a2a-js/sdk@0.3.7`
* Node 18+
* ClientFactory + HTTP+JSON transport
* Server: ADK/Agent Engine (A2A REST adapter)

## Repro (API call that rejects)

The SDK sends a POST to `/v1/message:send` with this shape:

```json
{
  "message": {
    "kind": "message",
    "role": "user",
    "messageId": "msg-1",
    "parts": [{ "kind": "text", "text": "hello" }]
  },
  "configuration": { "blocking": true }
}
```

Response from ADK/Agent Engine: **HTTP 400** (request rejected).

## Spec vs SDK Mismatch (Summary)

Spec (proto-JSON): `ROLE_*` enums, `parts` oneof (no `kind`), `TASK_STATE_*`,
wrapper responses `{ task|message: ... }`.  
SDK 0.3.7: `role: "user"`, `parts` with `kind`, lower-case task states, expects
bare `Task`/`Message`.

## Impact

When talking to ADK/Agent Engine (which parses proto-JSON), requests are
rejected (invalid enum values / unknown fields), or responses cannot be
parsed because they are wrapped.

This forces client applications to implement a shim that:
* maps `user -> ROLE_USER`
* converts `parts` to proto-JSON oneof form (no `kind`)
* unwraps `{ task: ... }` / `{ message: ... }`
* maps `TASK_STATE_*` to lower-case states

## Request

Add a proto-JSON mode (or dedicated transport) for HTTP+JSON that:
* emits `ROLE_*` enums
* encodes `parts` oneof without `kind`
* unwraps `{ task|message: ... }`
* maps `TASK_STATE_*` for SDK consumers

### Relevant log output

```shell

```

### Code of Conduct

guglielmo-san:
# Description

This PR resolves an incompatibility in the payload shape between the REST Transport handlers and other A2A SDKs.

In this version, the transmitted payload now aligns with the field names and structure defined in the proto file. A new `a2a.ts` file has been introduced to represent the proto types, including their corresponding `fromJson` and `toJson` conversion functions.

**Technically this is a breaking change, however current implementation isn't compatible with other official SDKs, so fixing in-place.** Current situation with JSON format for HTTP+JSON is complicated as there is no explicit specification in 0.3, see [here](https://github.com/a2aproject/a2a-js/issues/291#issuecomment-3719125908) for more details.

Fixes #291  🦕

Release-As: 0.3.8

### DIFF
--- a/test/e2e.spec.ts
+++ b/test/e2e.spec.ts
@@ -121,18 +121,7 @@ describe('Client E2E tests', () => {
           const actual = await client.sendMessage({
             message: createTestMessage('1', 'test'),
           });
-          const sanitize = (obj: any): any => {
-            if (Array.isArray(obj)) return obj.map(sanitize);
-            if (obj && typeof obj === 'object') {
-              return Object.fromEntries(
-                Object.entries(obj)
-                  .filter(([, v]) => v !== undefined)
-                  .map(([k, v]) => [k, sanitize(v)])
-              );
-            }
-            return obj;
-          };
-          expect(sanitize(actual)).to.deep.equal(expected);
+          expect(removeUndefinedFields(actual)).to.deep.equal(expected);
         });
       });
 
@@ -174,24 +163,14 @@ describe('Client E2E tests', () => {
             actual.push(message);
           }
 
-          const sanitize = (obj: any): any => {
-            if (Array.isArray(obj)) return obj.map(sanitize);
-            if (obj && typeof obj === 'object') {
-              return Object.fromEntries(
-                Object.entries(obj)
-                  .filter(([, v]) => v !== undefined)
-                  .map(([k, v]) => [k, sanitize(v)])
-              );
-            }
-            return obj;
-          };
-          expect(sanitize(actual)).to.deep.equal(expected);
+          expect(removeUndefinedFields(actual)).to.deep.equal(expected);
         });
       });
     });
   });
 });
 
+const removeUndefinedFields = (obj: any) => JSON.parse(JSON.stringify(obj));
 function createTestMessage(id: string, text: string): Message {
   return {
     messageId: id,


src/client/transports/rest_transport.ts
```diff
@@ -46,35 +49,52 @@ export class RestTransport implements Transport {
   }
 
   async getExtendedAgentCard(options?: RequestOptions): Promise<AgentCard> {
-    return this._sendRequest<AgentCard>('GET', '/v1/card', undefined, options);
+    const response = await this._sendRequest<unknown>('GET', '/v1/card', undefined, options);
```
ishymko: Should we remove `TResponse` from `_sendRequest` as we don't use it anymore and always use untyped responses which are later converted by `fromJSON`?

Or it may be possible to pack `fromJSON`/`toJSON` into `_sendRequest` if we pass `MessageFns` for request and response and call those methods inside.

src/client/transports/rest_transport.ts
```diff
@@ -46,35 +49,52 @@ export class RestTransport implements Transport {
...
+    const response = await this._sendRequest<unknown>(
       'POST',
       `/v1/tasks/${encodeURIComponent(params.taskId)}/pushNotificationConfigs`,
-      {
-        pushNotificationConfig: params.pushNotificationConfig,
-      },
+      requestBody,
       options
     );
+
```
ishymko: ultra nit: seems like extra line breaks were added only in this method, remove?

src/server/express/rest_handler.ts
```diff
@@ -188,11 +193,18 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
     try {
       // Write first event
       if (!firstResult.done) {
-        res.write(formatSSEEvent(firstResult.value));
+        const proto = ToProto.messageStreamResult(
+          firstResult.value as Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent
```
ishymko: I think this cast and the same at 204 can be avoided if `stream` in `sendStreamResponse` is typed as 

```ts
AsyncGenerator<
  Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent,
  void,
  undefined
>
```

src/server/express/rest_handler.ts
```diff
@@ -332,9 +350,11 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
       const result = await restTransportHandler.getTask(
         req.params.taskId,
         context,
-        req.query.historyLength
+        req.query.historyLength ?? req.query.history_length
       );
-      sendResponse(res, HTTP_STATUS.OK, context, result);
+      //TODO: clarify for version 1.0.0 the shape of the historyLenght parameter, and if it should be added to the returned object
```
ishymko: nit: shouldn't this TODO be above near `req.query.historyLength ?? req.query.history_length`?

src/server/express/rest_handler.ts
```diff
@@ -269,7 +281,8 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
     asyncHandler(async (req, res) => {
       const context = await buildContext(req);
       const result = await restTransportHandler.getAuthenticatedExtendedAgentCard(context);
-      sendResponse(res, HTTP_STATUS.OK, context, result);
+      const protoResult = ToProto.agentCard(result);
+      sendResponse(res, HTTP_STATUS.OK, context, a2a.AgentCard.toJSON(protoResult));
```
ishymko: A thought about `sendResponse`: currently omitting `toJSON` isn't going to cause any compilation errors, maybe we can add a generic parameter and ask for both `MessageFns<T>` and `T` to invoke `toJSON` inside `sendResponse`?

package.json
```diff
@@ -72,7 +73,8 @@
     "lint:fix": "npx eslint . --fix",
     "coverage": "vitest run --coverage",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",
-    "test-build": "esbuild ./dist/client/index.js ./dist/server/index.js ./dist/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist"
+    "test-build": "esbuild ./dist/client/index.js ./dist/server/index.js ./dist/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist",
+    "generate-proto-types": "cd ./src/types && curl -o ./a2a.proto https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/grpc/a2a.proto && buf generate && rm ./a2a.proto"
```
ishymko: Should it be removed as currently patched version is used to work with both camelCase and snake_case JSON fields?

src/types/converters/from_proto.ts
ishymko, line 1: Maybe we can rearrange `types` folder a bit, so that generated code is all placed in one folder, as currently there is a mix of generated code and this `utils` folder at the same level.

Maybe something like `common/types`, `common/converters`?

src/client/transports/rest_transport.ts
guglielmo-san: Yes, good idea

package.json
guglielmo-san: true, forgot to remove it

src/types/converters/from_proto.ts
guglielmo-san: renamed as `types/pb` and `types/pb_converters`

src/client/transports/rest_transport.ts
```diff
@@ -187,7 +235,12 @@ export class RestTransport implements Transport {
     };
 
     if (body !== undefined && method !== 'GET') {
-      requestInit.body = JSON.stringify(body);
+      if (!requestType) {
+        throw new Error(
+          `Request body provided for ${method} ${path} but no toJson serializer provided.`
```
ishymko: nit: sometimes such errors are prefixed with "Bug: <error message>" to give a clear message that it's a library bug (as it's internal function so if we reach this line it means there is a bug in our code)

src/server/express/rest_handler.ts
```diff
@@ -131,18 +136,19 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
+    body?: T,
+    toJson?: a2a.MessageFns<T>['toJSON']
   ): void => {
     setExtensionsHeader(res, context);
     res.status(statusCode);
     if (statusCode === HTTP_STATUS.NO_CONTENT) {
       res.end();
     } else {
-      res.json(body);
+      res.json(toJson ? toJson(body!) : body);
```
ishymko: nit: should it throw like [here](https://github.com/a2aproject/a2a-js/pull/292/changes#r2685777588) to be consistent? It appears to be the same situation.

src/server/express/rest_handler.ts
```diff
@@ -131,18 +136,19 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
    * @param body - Response body (omitted for 204 responses)
    */
-  const sendResponse = (
+  const sendResponse = <T>(
     res: Response,
     statusCode: number,
     context: ServerCallContext,
-    body?: unknown
+    body?: T,
+    toJson?: a2a.MessageFns<T>['toJSON']
```
ishymko: Maybe let's do the same as in the client?

```suggestion
    responseType?: a2a.MessageFns<T>
```

src/types/converters/from_proto.ts
ishymko, line 1: `pb_converters` folder -> just `converters`?

vitest.config.ts
```diff
@@ -9,7 +9,7 @@ export default defineConfig({
       provider: 'v8',
       reporter: ['text', 'json', 'html', 'json-summary'],
       include: ['src/**/*'],
-      exclude: ['src/samples/**/*'],
+      exclude: ['src/samples/**/*', 'src/types/a2a.ts', 'src/types/google/**/*'],
```
ishymko: I believe it wasn't updated after restructuring files

```suggestion
      exclude: ['src/samples/**/*', 'src/types/pb/**/*'],
```

test/e2e.spec.ts
```diff
@@ -134,21 +134,38 @@ describe('Client E2E tests', () => {
             {
               id: taskId,
               contextId,
-              status: { state: 'submitted' },
+              status: {
+                state: 'submitted',
+                message: undefined,
+                timestamp: undefined,
```
ishymko: Shouldn't it be possible to remove `undefined` fields with the `sanitize` helper?

src/server/express/rest_handler.ts
```diff
@@ -131,18 +136,19 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
    * @param context - ServerCallContext for setting extension headers
    * @param body - Response body (omitted for 204 responses)
```
ishymko, line 137: nit: add new param to the JSDoc?

test/e2e.spec.ts
```diff
@@ -121,8 +121,18 @@ describe('Client E2E tests', () => {
...
+            if (Array.isArray(obj)) return obj.map(sanitize);
+            if (obj && typeof obj === 'object') {
+              return Object.fromEntries(
+                Object.entries(obj)
+                  .filter(([, v]) => v !== undefined)
+                  .map(([k, v]) => [k, sanitize(v)])
+              );
+            }
+            return obj;
+          };
```
ishymko: If `const sanitize = (obj: any) => JSON.parse(JSON.stringify(obj));` works with current state I'd keep it for simplicity. Also this function can be declared in this file to avoid duplicating in both tests. And the last nit is to rename it to make the intent explicit, i.e. `removeUndefinedFields`.

ishymko (approved):

PR #299, opened 2026-01-15
feat: implement gRPC client

### LINKED ISSUE #275: [Feat]: Add support for gRpc

guglielmo-san:
# Description

This PR extends the implementation of grpc to Client.
A new index.js is created to export the grpc specific functions, preventing the need for every client to install the `grpc/js` package

Release-As: 0.3.9

Fixes #275  🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-prettier": "^5.5.4",
     "express": "^5.1.0",
-    "gts": "^6.0.2",
     "json-schema-to-typescript": "^15.0.4",
     "prettier": "^3.6.2",
     "supertest": "^7.1.4",
@@ -95,7 +94,6 @@
     "vitest": "^3.2.4"
   },
   "scripts": {
-    "clean": "gts clean",
     "build": "tsup",
     "test": "vitest run",
     "test:edge": "vitest run --config vitest.edge.config.ts",
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "eslint-config-prettier": "^10.1.8",
         "eslint-plugin-prettier": "^5.5.4",
         "express": "^5.1.0",
-        "gts": "^6.0.2",
         "json-schema-to-typescript": "^15.0.4",
         "prettier": "^3.6.2",
         "supertest": "^7.1.4",
@@ -87,28 +86,6 @@
       "funding": {
         "url": "https://github.com/sponsors/philsturgeon"
       }
-    },
-    "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
-      "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@babel/helper-validator-identifier": "^7.27.1",
-        "js-tokens": "^4.0.0",
-        "picocolors": "^1.1.1"
-      },
-      "engines": {
-        "node": ">=6.9.0"
-      }
-    },
-    "node_modules/@babel/code-frame/node_modules/js-tokens": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz",
-      "integrity": "sha512-RdJUflcE3cUzKiMqQgsCu06FPu9UdIJO0beYbPhHN4k6apgJtifcoCtT9bcxOpYBtpD2kCM6Sbzg4CausW/PKQ==",
-      "dev": true,
-      "license": "MIT"
     },
     "node_modules/@babel/helper-string-parser": {
       "version": "7.27.1",
@@ -1155,22 +1132,6 @@
         "node": ">=18.18.0"
       }
     },
-    "node_modules/@humanwhocodes/config-array": {
-      "version": "0.13.0",
-      "resolved": "https://registry.npmjs.org/@humanwhocodes/config-array/-/config-array-0.13.0.tgz",
-      "integrity": "sha512-DZLEEqFWQFiyK6h5YIeynKx7JlvCYWL0cImfSRXZ9l4Sg2efkFGTuFf6vzXjK1cq6IYkU+Eg/JizXw+TD2vRNw==",
-      "deprecated": "Use @eslint/config-array instead",
-      "dev": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@humanwhocodes/object-schema": "^2.0.3",
-        "debug": "^4.3.1",
-        "minimatch": "^3.0.5"
-      },
-      "engines": {
-        "node": ">=10.10.0"
-      }
-    },
     "node_modules/@humanwhocodes/module-importer": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/@humanwhocodes/module-importer/-/module-importer-1.0.1.tgz",
@@ -1184,14 +1145,6 @@
         "type": "github",
         "url": "https://github.com/sponsors/nzakas"
       }
-    },
-    "node_modules/@humanwhocodes/object-schema": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/@humanwhocodes/object-schema/-/object-schema-2.0.3.tgz",
-      "integrity": "sha512-93zYdMES/c1D69yZiKDBj0V24vqNzB/koF26KPaagAfd3P/4gUlh3Dys5ogAK+Exi9QyzlD8x/08Zt7wIKcDcA==",
-      "deprecated": "Use @eslint/object-schema instead",
-      "dev": true,
-      "license": "BSD-3-Clause"
     },
     "node_modules/@humanwhocodes/retry": {
       "version": "0.4.3",
@@ -1768,44 +1721,6 @@
       },
       "funding": {
         "url": "https://paulmillr.com/funding/"
-      }
-    },
-    "node_modules/@nodelib/fs.scandir": {
-      "version": "2.1.5",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
-      "integrity": "sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@nodelib/fs.stat": "2.0.5",
-        "run-parallel": "^1.1.9"
-      },
-      "engines": {
-        "node": ">= 8"
-      }
-    },
-    "node_modules/@nodelib/fs.stat": {
-      "version": "2.0.5",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.5.tgz",
-      "integrity": "sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">= 8"
-      }
-    },
-    "node_modules/@nodelib/fs.walk": {
-      "version": "1.2.8",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.walk/-/fs.walk-1.2.8.tgz",
-      "integrity": "sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@nodelib/fs.scandir": "2.1.5",
-        "fastq": "^1.6.0"
-      },
-      "engines": {
-        "node": ">= 8"
       }
     },
     "node_modules/@paralleldrive/cuid2": {
@@ -2392,13 +2307,6 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/@types/minimist": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/@types/minimist/-/minimist-1.2.5.tgz",
-      "integrity": "sha512-hov8bUuiLiyFPGyFPE1lwWhmzYbirOXQNNo40+y3zow8aFVTeyn3VWL0VFFfdNddA8S4Vf0Tc062rzyNr7Paag==",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/@types/node": {
       "version": "22.19.3",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-22.19.3.tgz",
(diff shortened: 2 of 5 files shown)


test/e2e.spec.ts
```diff
@@ -47,19 +50,29 @@ const transportConfigs: TransportConfig[] = [
...
+    ClientFactoryOptions.createFrom(ClientFactoryOptions.default, {
+      transports: [new GrpcTransportFactory()],
+    })
+  );
 
   transportConfigs.forEach((transportConfig) => {
     describe(`[${transportConfig.name}]`, () => {
       let app: Express;
       let server: Server;
+      let grpc_server: grpc.Server;
```
ishymko: Snake case is not used usually in TS/JS for variables
```suggestion
      let grpcServer: grpc.Server;
```

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,335 @@
...
+            break;
+          case 'task':
+            yield FromProto.task(payload.value);
+            break;
+          case 'statusUpdate':
+            yield FromProto.taskStatusUpdateEvent(payload.value);
+            break;
+          case 'artifactUpdate':
+            yield FromProto.taskArtifactUpdateEvent(payload.value);
+            break;
```
ishymko: I believe there is a code for this in `FromProto.messageStreamResult` used in REST, can we use it here as well?

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,335 @@
...
+            'listTaskPushNotificationConfig',
+          ].includes(method)
+        ) {
+          return new PushNotificationNotSupportedError(error.details);
+        }
+        if (['getAgentCard', 'taskSubscription'].includes(method)) {
+          return new UnsupportedOperationError(error.details);
+        }
+        break;
+      //TODO: add case for grpc.status.INVALID_ARGUMENT and grpc.status.INTERNAL
```
ishymko: Can you elaborate on this TODO? Is it also blocked by 1.0 or by something else? Let's make it clear why we can't resolve this TODO right now.

src/types/converters/to_proto.ts
```diff
@@ -560,4 +561,8 @@ export class ToProto {
       name: generateTaskName(params.id),
     };
   }
+
+  static getAgentCardRequest(): GetAgentCardRequest {
+    return {};
+  }
```
ishymko, line 567: Are you sure this is needed? Maybe the caller can just do `() => {}`?

package.json
ishymko, line 34: Please update this place with the new entrypoint as well.

src/types/converters/from_proto.ts
```diff
@@ -107,8 +107,8 @@ export class FromProto {
       kind: 'message',
       messageId: message.messageId,
       parts: message.content.map((p) => FromProto.part(p)),
-      contextId: message.contextId,
-      taskId: message.taskId,
+      contextId: message.contextId || undefined,
+      taskId: message.taskId || undefined,
```
ishymko, line 111: Is it the only place we need it? What about arrays an strings in other types?

src/types/converters/to_proto.ts
guglielmo-san: Yes it is not needed, I just thought it looks nicer

src/types/converters/from_proto.ts
guglielmo-san: I added here now just to make the test pass, I will add it in all the other places when the a2a_types is also updated

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,322 @@
...
+    parser: (req: TParams) => TReq
+  ): AsyncGenerator<A2AStreamEventData, void, undefined> {
+    const streamResponse = call(
+      parser(params),
+      this._buildMetadata(options),
+      this.grpcCallOptions ?? {}
+    );
+    try {
+      for await (const response of streamResponse) {
+        console.log('response', response);
```
ishymko: Remove?

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,322 @@
...
+type GrpcStreamCall<TReq, TRes> = (
+  request: TReq,
+  metadata?: grpc.Metadata,
+  options?: Partial<grpc.CallOptions>
+) => grpc.ClientReadableStream<TRes>;
+
+export interface GrpcTransportOptions {
+  endpoint: string;
+  grpcChannelCredentials?: grpc.ChannelCredentials;
+  grpcCallOptions?: Partial<grpc.CallOptions>;
```
ishymko, line 45: I believe now `abortSignal` we have in `RequestOptions` (set when `Client` is called) isn't going to be used in gRPC transport.

Is it possible to use it?

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,322 @@
...
+  callback: (error: grpc.ServiceError | null, response: TRes) => void
+) => grpc.ClientUnaryCall;
+
+type GrpcStreamCall<TReq, TRes> = (
+  request: TReq,
+  metadata?: grpc.Metadata,
+  options?: Partial<grpc.CallOptions>
+) => grpc.ClientReadableStream<TRes>;
+
+export interface GrpcTransportOptions {
```
ishymko, line 42: Should we allow providing your own client `A2AServiceClient` similar like we do it for `fetch` function for HTTP-based transports?

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,322 @@
...
+    params: DeleteTaskPushNotificationConfigParams,
+    options?: RequestOptions
+  ): Promise<void> {
+    await this._sendGrpcRequest(
+      'deleteTaskPushNotificationConfig',
+      params,
+      options,
+      this.grpcClient.deleteTaskPushNotificationConfig.bind(this.grpcClient),
+      ToProto.deleteTaskPushNotificationConfigParams,
+      () => {}
```
ishymko, line 155: Should it be named same way as it is done for `ToProto.getAgentCardRequest` in this PR? Or both places reverted to `() => {}`.

test/e2e.spec.ts
```diff
@@ -47,19 +50,29 @@ const transportConfigs: TransportConfig[] = [
     preferredTransport: 'HTTP+JSON',
     serverPath: '/a2a/rest',
   },
+  {
+    name: 'GRPC',
+    preferredTransport: 'GRPC',
+    serverPath: '/a2a/grpc',
```
ishymko: I believe `serverPath` doesn't make sense for gRPC, in the code it's handled in a special way and is replaced with `localhost:${port}` after server is started.

src/client/transports/grpc/grpc_transport.ts
guglielmo-san: Uhm right, it is possible to use it with a custom implementation bridging between abort signal and grpc

src/client/transports/grpc/grpc_transport.ts
guglielmo-san: I had it in the first implementation, but considering that a client can be as complex as with the URL and the connections credentials, I opted to remove it, as those two params can be sent as options. Having client, url and credentials as options would redundant

src/client/transports/grpc/grpc_transport.ts
guglielmo-san: In this case there is no custom empty proto object that is returned

ishymko: Maybe `ToProto.empty` in both places?

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -0,0 +1,349 @@
...
+          resolve(converter(response));
+        }
+      );
+
+      if (options?.signal) {
+        if (options.signal.aborted) {
+          clientCall.cancel();
+        } else {
+          onAbort = () => clientCall.cancel();
+          options.signal.addEventListener('abort', onAbort);
```
ishymko, line 227: Maybe we can do 
```suggestion
          options.signal.addEventListener('abort', onAbort, { once: true });
```

like in [this](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#implementing_an_abortable_api) example and get rid of `removeEventListener`?

The behavior described in [`once` documentation](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) looks like what we need.

guglielmo-san: uhm but if the abort is not called the listener is not removed anyway when the function completes.
if the client reuses the same AbortFunction this might cause memory leaks

ishymko: Indeed!

PR #303, opened 2026-01-16
fix: use proto default values in generated a2a_types.ts

guglielmo-san:
# Description

This PR standardizes the handling of optional fields in the generated a2a_types and converters. It aligns the codebase with Protobuf standards by replacing undefined checks with empty string checks for missing string values.

**Key Changes**
- Generated Types (a2a_types.ts):
  - Removed the optionality from generated string fields.
  - Updated fromJSON and toJSON methods to validate against empty strings ("") instead of undefined for missing values.

- Converters:
  - ToProto: Updated to convert undefined domain fields into empty strings, ensuring compliance with the new non-optional proto types.
  - FromProto: Updated to convert empty strings from the proto layer back into undefined, preserving the existing undefined semantics for the internal domain model.

**Motivation**
These changes ensure consistency with the Protobuf specification, where missing strings must default to the empty string rather than undefined.

### DIFF
--- a/src/types/converters/from_proto.ts
+++ b/src/types/converters/from_proto.ts
@@ -488,13 +488,8 @@ export class FromProto {
     event: StreamResponse
   ): types.Message | types.Task | types.TaskStatusUpdateEvent | types.TaskArtifactUpdateEvent {
     switch (event.payload?.$case) {
-      case 'msg': {
-        const message = FromProto.message(event.payload.value);
-        if (!message) {
-          throw A2AError.internalError('Invalid message in StreamResponse');
-        }
-        return message;
-      }
+      case 'msg':
+        return FromProto.message(event.payload.value);
       case 'task':
         return FromProto.task(event.payload.value);
       case 'statusUpdate':
--- a/src/types/converters/to_proto.ts
+++ b/src/types/converters/to_proto.ts
@@ -36,6 +36,7 @@ import {
   CancelTaskRequest,
   TaskSubscriptionRequest,
   CreateTaskPushNotificationConfigRequest,
+  GetAgentCardRequest,
 } from '../pb/a2a_types.js';
 import { generatePushNotificationConfigName, generateTaskName } from './id_decoding.js';
 
@@ -559,5 +560,9 @@ export class ToProto {
     return {
       name: generateTaskName(params.id),
     };
+  }
+
+  static getAgentCardRequest(): GetAgentCardRequest {
+    return {};
   }
 }
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
       ],
       "client": [
         "./dist/client/index.d.ts"
+      ],
+      "client/grpc": [
+        "./dist/client/transports/grpc/index.d.ts"
       ]
     }
   },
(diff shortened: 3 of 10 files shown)


src/types/converters/from_proto.ts
```diff
@@ -105,10 +105,10 @@ export class FromProto {
 
     return {
       kind: 'message',
-      messageId: message.messageId,
+      messageId: message.messageId || undefined,
```
ishymko: Should we do `|| undefined` for fields which do not accept `undefined? Now it doesn't matter as we don't have `strict` enabled, but I believe it's better to avoid it so that we don't migrate this code later. 

Let's do `|| undefined` only when target field allows undefined it its type.

src/types/pb/google/protobuf/descriptor.ts
ishymko, line 1: Can we avoid putting `google/` files under `src/types`? From what I see `a2a_types.ts` only imports `timestamp.ts` and `google/api` isn't going to be used for types without services.

src/types/buf.gen.yaml
ishymko, line 1: I don't see our new ProtoJSON version here, I believe running `buf generate` against this config is going to change the content of `a2a_types.ts`. If new ts-proto version is not yet in buf plugins repo, let's remove this for now to avoid confusion.

src/types/pb/google/protobuf/descriptor.ts
guglielmo-san: Oh right, I run again the generator to check it works with buf and i forgot to remove the new files created
Although we cannot use the `google/` types generated in the grpc folder, as they do not contain the `ToJson/FromJson` methods

src/types/converters/from_proto.ts
guglielmo-san: true, makes sense

src/types/buf.gen.yaml
guglielmo-san: It is actually using the v2.11.0, which includes the protoJSON format. I did not specifically add the field 'protoJsonFormat' as it default to true

src/types/pb/google/protobuf/descriptor.ts
guglielmo-san: So i removed them all except timestamp.ts

src/types/buf.gen.yaml
ishymko: My bad, forgot that it's the default!

src/types/README.md
```diff
@@ -0,0 +1,29 @@
...
+```bash
+npx buf generate
+```
+
+This will generate the TypeScript files in the `./pb` directory (as configured in `buf.gen.yaml`).
+
+## Post-Processing
+
+After generating the types, it may be convenient to remove the files generated in the `google` directory, as
+only the `timestamp.ts` is used in the SDK types without services.
```
ishymko: Is it possible to use [`inputs` buf config](https://buf.build/docs/configuration/v2/buf-gen-yaml/#inputs) for this?

src/types/converters/to_proto.ts
ishymko, line 1: (likely) nit: Should we use `?? ''` here? `|| undefined` was used in`from_proto.ts` to handle empty strings, while here we only fallback from `undefined`.

src/types/README.md
guglielmo-san: I tried it already in the past but it does not work as expected
I found a better way now: the parameter `useDate=string`. In this way it converts the google.protobuf.timestamp to a string, without the need for the import, and is compliant with the spec aswell https://a2a-protocol.org/latest/specification/#561-timestamps

ishymko: SG for now, however we may want to change it if we want to use proto generated types everywhere, otherwise it has to be parsed each time to do anything. However as now we have `string` in `types.ts` either way it makes sense to skip `google.protobuf.timestamp`.

PR #311, opened 2026-01-22
fix: do not use ReadableStream async iterator in SSE stream parsing

### LINKED ISSUE #308: [Bug]: value of response.body.pipeThrough is undefined (SSE)
### What happened?

On iOS, I'm seeing the logged error.

Looks like the `response.body.pipeThrough` from:
https://github.com/a2aproject/a2a-js/blob/3d9a052fe8cd917fcff1798b03a4865effabf0be/src/sse_utils.ts#L111
is not available on Safari.

### Relevant log output

```shell
n of e.body.pipeThrough is not a function.
In 'n of e.body.pipeThrough(new TextDecoderStream)', 'n of e.body.pipeThrough' is undefined
```

### Code of Conduct

ilbertt:
# Description

The async iterator in `ReadableStream` is not available in browsers like Safari: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility). This PR changes it to a lower-level API which has better support across different environments.

Fixes #308.

### DIFF
--- a/test/sse_utils.spec.ts
+++ b/test/sse_utils.spec.ts
@@ -93,10 +93,13 @@ describe('SSE Utils', () => {
     });
   });
 
-  describe('parseSseStream', () => {
+  describe.each([
+    ['with native async iterator', createMockResponse],
+    ['without native async iterator', createMockResponseWithoutAsyncIterator],
+  ])('parseSseStream (%s)', (_, createResponse) => {
     it('should parse a single data event', async () => {
       const sseData = 'data: {"kind":"message"}\n\n';
-      const response = createMockResponse(sseData);
+      const response = createResponse(sseData);
 
       const events: SseEvent[] = [];
       for await (const event of parseSseStream(response)) {
@@ -110,7 +113,7 @@ describe('SSE Utils', () => {
 
     it('should parse an error event', async () => {
       const sseData = 'event: error\ndata: {"code":-32001}\n\n';
-      const response = createMockResponse(sseData);
+      const response = createResponse(sseData);
 
       const events: SseEvent[] = [];
       for await (const event of parseSseStream(response)) {
@@ -124,7 +127,7 @@ describe('SSE Utils', () => {
 
     it('should parse multiple events', async () => {
       const sseData = 'data: {"id":1}\n\ndata: {"id":2}\n\n';
-      const response = createMockResponse(sseData);
+      const response = createResponse(sseData);
 
       const events: SseEvent[] = [];
       for await (const event of parseSseStream(response)) {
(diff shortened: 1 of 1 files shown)


src/sse_utils.ts
```diff
@@ -137,3 +137,43 @@ export async function* parseSseStream(
     yield { type: eventType, data: eventData };
   }
 }
+
+function createAsyncIterator(
+  stream: ReadableStream<string>
+): () => AsyncGenerator<Awaited<string>, void, unknown> {
```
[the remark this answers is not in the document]
ilbertt: Fixed!

src/sse_utils.ts
```diff
@@ -137,3 +137,43 @@ export async function* parseSseStream(
...
+ * @returns The stream with the async iterator
+ *
+ * @example
+ * ```ts
+ * for await (const value of pipeThrough(response.body)) {
+ *   console.log(value);
+ * }
+ * ```
+ */
+function pipeThrough(body: ReadableStream<Uint8Array<ArrayBuffer>>): ReadableStream<string> {
```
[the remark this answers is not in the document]
ilbertt: @gemini-code-assist It actually does since the latest TS versions

src/sse_utils.ts
```diff
@@ -137,3 +137,43 @@ export async function* parseSseStream(
...
+        if (done) {
+          break;
+        }
+        yield value;
+      }
+    } finally {
+      reader.releaseLock();
+    }
+  };
+}
```
ishymko: What do you think about using this API always?

Something like

```ts
async function* readFrom(stream: ReadableStream<string>): AsyncGenerator<string, void, void> {
  const reader = stream.getReader();
  try {
    while (true) {
      const { done, value } = await reader.read();
      if (done) {
        break;
      }
      yield value;
    }
  } finally {
    reader.releaseLock();
  }
}
```

and then calling it in without patching logic

```ts
for await (const value of readFrom(response.body.pipeThrough(new TextDecoderStream())))
```

As we should either way write this code as a polyfill. Native implementation can be a bit faster but I don't think it's going be anything noticeable in the IO-bound context.

ishymko (commented):
Thank you for the prompt fix!

BTW were you able to test it on a real iOS? We don't have this kind of integration tests running on a real iOS so if you were able to confirm it it'd be great.

ilbertt:
> BTW were you able to test it on a real iOS? We don't have this kind of integration tests running on a real iOS so if you were able to confirm it it'd be great.

I tested the workaround from [my comment](https://github.com/a2aproject/a2a-js/issues/308#issuecomment-3783927607), and it was working. If you can release a beta/patch version, I'm happy to pull it in and test it

src/sse_utils.ts
ilbertt: Could make sense, although it also makes sense to show that we need a workaround for some environments in my opinion.

I agree the approach you proposed is cleaner though.

ishymko: > also makes sense to show that we need a workaround for some environments in my opinion

Yeah, fair point. So how about always using the older API but with a comment linking to MDN?

This will also be enforced via a test you added. 

Patching approach is more advanced for sure and it makes use of built-in implementation when available, but in this case I'd prefer simplicity.

test/sse_utils.spec.ts
```diff
@@ -174,4 +202,73 @@ describe('SSE Utils', () => {
...
+        if (parsedEvents.length === 1) {
+          break;
+        }
+      }
+
+      expect(parsedEvents).toHaveLength(1);
+      expect(JSON.parse(parsedEvents[0].data)).toEqual({ kind: 'first' });
+    });
+
+    it('should release reader lock when breaking early without async iterator', async () => {
```
ishymko: This test still passes with `reader.releaseLock();` commented out.

I tried to do `getReader()` and it also works outside of the loop even without `reader.releaseLock();`. It only fails inside the loop saying that the stream is locked (which makes sense).

Maybe GC does something for us in this case when reader goes out of scope, so not sure if it's possible to test properly.

test/sse_utils.spec.ts
```diff
@@ -30,6 +30,34 @@ function createMockResponse(sseData: string, chunkSize: number = 2): Response {
...
+  const stream = new ReadableStream({
+    pull(controller) {
+      if (chunkIndex < chunks.length) {
+        controller.enqueue(chunks[chunkIndex]);
+        chunkIndex++;
+      } else {
+        controller.close();
+      }
+    },
+  });
```
ishymko: Looks redundant to me as there is only one chunk always.

Can we simplify here or simply extract stream creation part from `createMockResponse` into a separate function and call from both places?

test/sse_utils.spec.ts
```diff
@@ -174,4 +202,73 @@ describe('SSE Utils', () => {
       expect(JSON.parse(parsedEvents[1].data)).toEqual(errorEvent);
     });
   });
+
+  describe('Async iterator fallback', () => {
```
ishymko: Should be possible to utilize existing `parseSseStream` tests by parametrizing via `describe.each` or similar.

As what we want to achieve is that everything works in both environments.

src/sse_utils.ts
ilbertt: Done!

test/sse_utils.spec.ts
ilbertt: Refactored the test!

test/sse_utils.spec.ts
ilbertt: Done!

test/sse_utils.spec.ts
```diff
@@ -175,3 +209,58 @@ describe('SSE Utils', () => {
     });
   });
 });
+
+/**
+ * Tests that parseSseStream works correctly in both environments:
+ * - With native ReadableStream async iterator support
+ * - Without native async iterator (simulating older browsers/runtimes)
+ */
```
ishymko: Can we just modify existing tests under `describe('parseSseStream'` with the `.each` and pass those factory functions as you did here?

ilbertt: Ah, sorry. I didn't see those tests. I've pushed the change and removed my tests as they were pretty much duplicates

PR #318, opened 2026-02-09
fix: throw on JSON-RPC response id mismatch

### LINKED ISSUE #176: [Bug]: Throw in case of JSON-RPC request id mismatch
Extracted from a code review comment. Currently mismatching JSON-RPC id produces an error log but execution continues. Wrong JSON-RPC id means that protocol invariant is broken, so "keep going" rather does not apply here.

nit: Go SDK doesn't perform any validation, but I'd probably be throwing in case of a mismatch

_Originally posted by @yarolegovich in https://github.com/a2aproject/a2a-js/pull/169#discussion_r2517645896_

corvid-agent:
## Summary

Replaces `console.error`/`console.warn` calls with thrown errors when a JSON-RPC response id does not match the request id. Per the JSON-RPC 2.0 specification, this is a protocol invariant violation — continuing execution with a mismatched response can lead to silent data corruption.

### Changes

- **`_sendRpcRequest`**: `console.error` → `throw new Error` on id mismatch
- **`_processSseEventData`**: `console.warn` → `throw new Error` on id mismatch
- **`_processSseEventData` catch block**: Added the new error message to the pass-through condition so it is not swallowed and re-wrapped as a parse error

### Context

From the code review on PR #169: *"Go SDK doesn't perform any validation, but I'd probably be throwing in case of a mismatch"* — @yarolegovich

Fixes #176

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -233,7 +233,7 @@ export class JsonRpcTransport implements Transport {
     const rpcResponse: JSONRPCResponse = await httpResponse.json();
     if (rpcResponse.id !== requestId) {
       throw new Error(
-        `CRITICAL: RPC response ID mismatch for method ${method}. Expected ${requestId}, got ${rpcResponse.id}.`
+        `JSON-RPC response ID mismatch for method ${method}. Expected ${requestId}, got ${rpcResponse.id}.`
       );
     }
 
@@ -329,7 +329,7 @@ export class JsonRpcTransport implements Transport {
 
     if (a2aStreamResponse.id !== originalRequestId) {
       throw new Error(
-        `SSE Event's JSON-RPC response ID mismatch. Client request ID: ${originalRequestId}, event response ID: ${a2aStreamResponse.id}.`
+        `JSON-RPC response ID mismatch in SSE event. Expected ${originalRequestId}, got ${a2aStreamResponse.id}.`
       );
     }
 


src/client/transports/json_rpc_transport.ts
```diff
@@ -343,7 +343,8 @@ export class JsonRpcTransport implements Transport {
       if (
         e instanceof Error &&
         (e.message.startsWith('SSE event contained an error') ||
-          e.message.startsWith("SSE event JSON-RPC response is missing 'result' field"))
+          e.message.startsWith("SSE event JSON-RPC response is missing 'result' field") ||
+          e.message.startsWith("SSE Event's JSON-RPC response ID mismatch"))
```
ishymko: It was there before, but as long as it has to be updated either way, let's refactor this part to avoid catching errors thrown in the same code block and match them against a message. Similar was done in #167 in other places,  `JSON.parse` should be in the `try...catch`.

corvid-agent: Done — isolated `JSON.parse` in its own try/catch and moved the validation logic outside so there is no more string matching on re-thrown errors. Follows the same pattern as #167.

src/client/transports/json_rpc_transport.ts
```diff
@@ -232,7 +232,7 @@ export class JsonRpcTransport implements Transport {
 
     const rpcResponse: JSONRPCResponse = await httpResponse.json();
     if (rpcResponse.id !== requestId) {
-      console.error(
+      throw new Error(
         `CRITICAL: RPC response ID mismatch for method ${method}. Expected ${requestId}, got ${rpcResponse.id}.`
```
ishymko: nit: let's unify error messages here and in the SSE part, also `CRITICAL` can be removed now, as it's not logged and is thrown either way.

ishymko (approved):

src/client/transports/json_rpc_transport.ts
corvid-agent: Done — unified both messages to follow the same `JSON-RPC response ID mismatch [context]. Expected [expected], got [actual].` pattern and removed the `CRITICAL` prefix.

ishymko (approved):
🚀

PR #319, opened 2026-02-09
feat: expose HTTP status code and headers on SDK error classes

### LINKED ISSUE #317: [Feat]: Passthrough http status code in error

When handling errors from `@a2a-js/sdk`, the underlying HTTP status codes are abstracted away, making it impossible to implement proper error handling logic based on the HTTP response. This limits the ability to differentiate between various failure scenarios that require different handling strategies.

For example, when a server returns HTTP 402 (Payment Required), 401 (Unauthorized), or 503 (Service Unavailable), the application needs to handle each differently, but currently all are wrapped in SDK errors without exposing the status code. This forces developers to treat all HTTP errors identically, losing important context about the failure reason.

For instance:
- Payment-required services (402) should trigger payment flows, not be treated as errors
- Authentication failures (401) should prompt re-authentication
- Rate limiting (429) should implement backoff strategies
- Server errors (5xx) vs client errors (4xx) require different retry logic

Add `statusCode` and optionally `headers` properties to **all SDK error classes** that can originate from HTTP responses:

```typescript
// Add to base error class or each specific error class
class TaskNotFoundError extends Error {
  statusCode?: number;
  headers?: Record<string, string>;
  cause?: Error;
}

class InvalidAgentResponseError extends Error {
  statusCode?: number;
  headers?: Record<string, string>;
  cause?: Error;
}

class UnsupportedOperationError extends Error {
  statusCode?: number;
  headers?: Record<string, string>;
  cause?: Error;
}

// ... etc for all HTTP-related SDK errors
```

This would enable proper HTTP-aware error handling across all error types:

```typescript
try {
  await client.sendMessage(message);
} catch (err) {
  // Check if it's any SDK error with a status code
  if (err instanceof Error && 'statusCode' in err) {
    const statusCode = (err as any).statusCode;
    
    switch (statusCode) {
      case 402:
        // Payment required - handle payment flow
        break;
      case 401:
        // Unauthorized - trigger authentication
        break;
      case 404:
        // TaskNotFoundError with 404 status
        break;
      case 429:
        // Rate limited - implement backoff
        break;
      case 503:
        // Service unavailable - retry later
        break;
      default:
        if (statusCode >= 500) {
          // Server error - treat as temporary failure
        }
    }
  }
}
```

**Why all error classes need this:**

Different SDK error types can originate from different HTTP status codes:
- `TaskNotFoundError` → 404 Not Found
- `UnsupportedOperationError` → 405 Method Not Allowed  
- `InvalidAgentResponseError` → 400 Bad Request, 402 Payment Required, etc.
- `AuthenticatedExtendedCardNotConfiguredError` → 401 Unauthorized, 403 Forbidden

Applications need to handle these based on both the error type (what went wrong) and the status code (how the server categorized it).

The only alternative right now is that server have to parse the error string from the status code.

### Additional context

_No response_

### Code of Conduct

corvid-agent:
## Summary
- Adds optional `statusCode` and `headers` properties to all A2A SDK error classes (`TaskNotFoundError`, `TaskNotCancelableError`, `PushNotificationNotSupportedError`, `UnsupportedOperationError`, `ContentTypeNotSupportedError`, `InvalidAgentResponseError`, `AuthenticatedExtendedCardNotConfiguredError`) so that consumers can implement HTTP-aware error handling.
- Updates the REST transport (`RestTransport`) to extract HTTP status code and response headers from the `Response` object and pass them through when constructing error instances via `mapToError`.
- Exports the new `A2AErrorHttpDetails` interface from the client entry point for use by consumers.
- Adds comprehensive tests covering all error types, verifying that `statusCode` and `headers` are correctly populated on errors thrown by the REST transport.

Closes #317

## Test plan
- [x] All 414 existing tests pass (no regressions)
- [x] New tests verify `statusCode` is set on `TaskNotFoundError` (404), `TaskNotCancelableError` (409), `PushNotificationNotSupportedError` (400), `UnsupportedOperationError` (405), `ContentTypeNotSupportedError` (415), `InvalidAgentResponseError` (502), and `AuthenticatedExtendedCardNotConfiguredError` (401)
- [x] New test verifies that response headers (e.g. `Retry-After`, `X-RateLimit-Remaining`) are available on errors for rate-limiting / backoff strategies
- [x] Backward compatibility verified: constructor changes are additive (second parameter is optional), so existing code including JSON-RPC error subclasses continues to work without modification
- [x] TypeScript lint passes with no new errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -45,3 +45,4 @@ export {
   TaskNotFoundError,
   UnsupportedOperationError,
 } from '../errors.js';
+export type { A2AErrorHttpDetails } from '../errors.js';
--- a/src/client/transports/rest_transport.ts
+++ b/src/client/transports/rest_transport.ts
@@ -1,6 +1,7 @@
 import { TransportProtocolName } from '../../core.js';
 import {
   A2A_ERROR_CODE,
+  type A2AErrorHttpDetails,
   AuthenticatedExtendedCardNotConfiguredError,
   ContentTypeNotSupportedError,
   InvalidAgentResponseError,
@@ -260,6 +261,7 @@ export class RestTransport implements Transport {
   private async _handleErrorResponse(response: Response, path: string): Promise<never> {
     let errorBodyText = '(empty or non-JSON response)';
     let errorBody: RestErrorResponse | undefined;
+    const httpDetails = RestTransport.extractHttpDetails(response);
 
     try {
       errorBodyText = await response.text();
@@ -274,7 +276,7 @@ export class RestTransport implements Transport {
     }
 
     if (errorBody && typeof errorBody.code === 'number') {
-      throw RestTransport.mapToError(errorBody);
+      throw RestTransport.mapToError(errorBody, httpDetails);
     }
 
     throw new Error(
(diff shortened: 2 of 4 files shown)


JakubWorek:
Closing as it was done in https://github.com/a2aproject/a2a-js/pull/587

Issue #321, opened 2026-02-11
[Feat]: Support 1.0 spec


ishymko:
https://a2a-protocol.org/v1.0.0/specification/

The development is happening in [`epic/1.0_breaking_changes`](https://github.com/a2aproject/a2a-js/tree/epic/1.0_breaking_changes).

FilipMasar:
Hi, when is this expected to be done? Asking so we can make a decision whether to start with v0.3 or wait for v1 support

ishymko:
Hi @FilipMasar, there is a high chance of having some pre-release version within a few weeks which is going to have 1.0 spec support. There is still a decision to be made on 0.3 spec compatibility, but this one will definitely take longer.

salqadri:
What's the ETA on this? It's been several months now.

ishymko:
## Update

Hi all, 1.0 alpha with 1.0 spec support is already available: https://github.com/a2aproject/a2a-js/releases/tag/v1.0.0-alpha.0, can be installed via

```
npm install @a2a-js/sdk@next
```

More details can be found [here](https://github.com/a2aproject/a2a-js/blob/v1.0.0-alpha.0/docs/migration-guide.md).

ishymko:
## Update

Hi all, **1.0 beta** with **both** 1.0 and 0.3 support is already available: https://github.com/a2aproject/a2a-js/releases/tag/v1.0.0-beta.0, can be installed via

```
npm install @a2a-js/sdk@next
```

See linked GitHub release description for mode details.

JakubWorek:
## Update

Hi all, **1.0.0** with **both** 1.0 and 0.3 support is already available: https://github.com/a2aproject/a2a-js/releases/tag/v1.0.0, can be installed via

```
npm install @a2a-js/sdk
```

See linked GitHub release description for mode details.

PR #337, opened 2026-02-20
feat(api)!: Update codebase to use only Proto based types

bartek-gralewicz:
# Description

Updated the codebase to only use Proto based types and remove the old JSON types.ts completely.

This change unifies the usage of types with other SDKs and is part of the 1.0.0 scope. Equivalent PR on the Python SDK: https://github.com/a2aproject/a2a-python/pull/572.

# Important

This is a breaking change and therefore it is going to be merged into an epic `v1.0` branch instead of `main`. Later 1.0 release will introduce this in the `main` branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6574,9 +6574,9 @@
       }
     },
     "node_modules/qs": {
-      "version": "6.14.1",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.1.tgz",
-      "integrity": "sha512-4EK3+xJl8Ts67nLYNwqw/dsFVnCf+qR7RgXSK9jEEm9unao3njwMDdmsdvoKBKHzxd7tCYz5e5M+SnMjdtXGQQ==",
+      "version": "6.15.0",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.15.0.tgz",
+      "integrity": "sha512-mAZTtNCeetKMH+pSjrb76NAM8V9a05I9aBZOHztWy/UqcJdQYNsf59vrRKWnojAT9Y+GbIvoTBC++CPHqpDBhQ==",
       "dev": true,
       "license": "BSD-3-Clause",
       "dependencies": {


src/server/request_handler/default_request_handler.ts
```diff
@@ -116,9 +125,13 @@ export class DefaultRequestHandler implements A2ARequestHandler {
     // Ensure taskId is present
     const taskId = incomingMessage.taskId || uuidv4();
 
-    if (incomingMessage.referenceTaskIds && incomingMessage.referenceTaskIds.length > 0) {
+    if (
+      (incomingMessage as Message & { referenceTaskIds?: string[] }).referenceTaskIds &&
+      (incomingMessage as Message & { referenceTaskIds?: string[] }).referenceTaskIds!.length > 0
+    ) {
```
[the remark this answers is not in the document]
bartek-gralewicz: Breaking changes are expected. I added a note to the PR description to make it clear. Resolving this comment and rerunning the review.

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -73,7 +93,12 @@ export class GrpcTransport implements Transport {
     params: MessageSendParams,
     options?: RequestOptions
   ): Promise<SendMessageResult> {
-    const rpcResponse = await this._sendGrpcRequest(
+    const rpcResponse = await this._sendGrpcRequest<
+      SendMessageRequest,
+      SendMessageResponse,
+      MessageSendParams,
+      SendMessageResult
```
ishymko: Here and below: is it necessary? I believe that for most `_sendGrpcRequest` calls type inference should be able to figure out types correctly and I don't see changes to `_sendGrpcRequest` here.

src/client/transports/transport.ts
```diff
@@ -23,19 +23,19 @@ export interface Transport {
...
     params: GetTaskPushNotificationConfigParams,
     options?: RequestOptions
-  ): Promise<TaskPushNotificationConfig>;
+  ): Promise<JsonRpcTaskPushNotificationConfig>;
 
   listTaskPushNotificationConfig(
     params: ListTaskPushNotificationConfigParams,
     options?: RequestOptions
-  ): Promise<TaskPushNotificationConfig[]>;
+  ): Promise<JsonRpcTaskPushNotificationConfig[]>;
```
ishymko: This interface should be transport agnostic without using `JsonRpcXxx` types, just proto-based types. If some envelope data structure is needed (as it's [needed for JSON-RPC](https://a2a-protocol.org/latest/specification/#93-base-request-structure)) it should reside in `json_rpc_transport.ts` and applied before sending over the wire.

src/json_rpc_types.ts
```diff
@@ -0,0 +1,388 @@
...
+  | GetTaskRequest
+  | CancelTaskRequest
+  | SetTaskPushNotificationConfigRequest
+  | GetTaskPushNotificationConfigRequest
+  | TaskResubscriptionRequest
+  | ListTaskPushNotificationConfigRequest
+  | DeleteTaskPushNotificationConfigRequest
+  | GetAuthenticatedExtendedCardRequest;
+
+export interface SendMessageRequest {
```
ishymko: Do we actually need to define each type individually here? The structure is [the same](https://a2a-protocol.org/latest/specification/#93-base-request-structure) and is already defined in the `json_rpc_transport.ts` with

```ts
interface JSONRPCRequest {
  jsonrpc: '2.0';
  method: string;
  params: unknown;
  id: string | number | null;
}

interface JSONRPCSuccessResponse<T> {
  jsonrpc: '2.0';
  result: T;
  id: string | number | null;
}
```

It should be possible to assume that `params` contains the actual proto payload serialized via ProtoJSON.

src/client/client.ts
ishymko, line 1: Good news: this legacy client can be dropped right away (we already marked it with `@deprecated`), feel free to handle it in a separate PR but either way changes here don't matter.

We have a list of breaking changes for 1.0 release in #179 - they are somewhat interleaved with spec support and this change but thing were not properly arranged in terms of what should follow what - apologize for not flagging it earlier.

bartek-gralewicz: For now I'll keep updating the client for the simplicity of running linters on the entire project without issues. I can remove this functionality in a separate PR once this activity is concluded.

src/samples/agents/movie-agent/index.ts
```diff
@@ -303,9 +315,14 @@ const movieAgentCard: AgentCard = {
       ],
       inputModes: ['text'], // Explicitly defining for skill
       outputModes: ['text', 'task-status'], // Explicitly defining for skill
+      security: [],
     },
   ],
   supportsAuthenticatedExtendedCard: false,
+  preferredTransport: 'jsonrpc',
```
jezekra1, line 322: The `preferredTransport` was removed in v1, should updated AgentCard format be a part of this PR? :)

src/samples/agents/movie-agent/index.ts
```diff
@@ -78,22 +79,23 @@ class MovieAgentExecutor implements AgentExecutor {
...
-        state: 'working',
-        message: {
-          kind: 'message',
-          role: 'agent',
+        state: TaskState.TASK_STATE_WORKING,
+        update: {
+          role: Role.ROLE_AGENT,
           messageId: uuidv4(),
-          parts: [{ kind: 'text', text: 'Processing your question, hang tight!' }],
+          content: [{ part: { $case: 'text', value: 'Processing your question, hang tight!' } }],
```
jezekra1, line 89: There is no `content` in a2a spec v1

```
message Message {
  // The unique identifier (e.g. UUID) of the message. This is created by the message creator.
  string message_id = 1 [(google.api.field_behavior) = REQUIRED];
  // Optional. The context id of the message. If set, the message will be associated with the given context.
  string context_id = 2;
  // Optional. The task id of the message. If set, the message will be associated with the given task.
  string task_id = 3;
  // Identifies the sender of the message.
  Role role = 4 [(google.api.field_behavior) = REQUIRED];
  // Parts is the container of the message content.
  repeated Part parts = 5 [(google.api.field_behavior) = REQUIRED];
  // Optional. Any metadata to provide along with the message.
  google.protobuf.Struct metadata = 6;
  // The URIs of extensions that are present or contributed to this Message.
  repeated string extensions = 7;
  // A list of task IDs that this message references for additional context.
  repeated string reference_task_ids = 8;
}
```
https://a2a-protocol.org/latest/definitions/

what version of the protocol are you using? Is this just refactor to protobuf types with the current version and switch v1 types will be coming later? Maybe I am misunderstanding this PR, I thought this was the v1 upgrade based on the equivalent Python PR :)

bartek-gralewicz: This is not a full v1.0 migration, only a protobuf types refactor like you wrote. This PR is also not designed to be merged into `main` but to a dedicated epic branch for v1.0 since some changes here are breaking changes anyway.

src/samples/agents/movie-agent/index.ts
bartek-gralewicz: Additional changes that are converting this SDK to v1.0 state will come in separate PRs. This PR is focused on switching to protobuf types.

src/json_rpc_types.ts
bartek-gralewicz: I applied some changes that reduced the code of these interface definitions.

src/client/transports/grpc/grpc_transport.ts
bartek-gralewicz: Removed redundant type specifications as suggested.

src/client/transports/transport.ts
bartek-gralewicz: Applied necessary changes. The interface is now transport agnostic.

src/client/client.ts
```diff
@@ -153,10 +155,15 @@ export class A2AClient {
...
-    return await this.invokeJsonRpc<MessageSendParams, SendMessageResponse>(
-      (t, p, id) => t.sendMessage(p, A2AClient.emptyOptions, id),
-      params
-    );
+    const resultFn: JsonRpcCaller<MessageSendParams, SendMessageResponse> = async (t, p, id) => {
+      const result = await t.sendMessage(p, A2AClient.emptyOptions, id);
+      if ('messageId' in result) {
+        return { payload: { $case: 'msg', value: result } };
+      }
+      return { payload: { $case: 'task', value: result } };
```
ishymko: Maybe this conversion logic can be moved to `ToProto`?

src/client/client.ts
```diff
@@ -197,10 +204,27 @@ export class A2AClient {
         'Agent does not support push notifications (AgentCard.capabilities.pushNotifications is not true).'
       );
     }
+    const taskIdMatch = params.name.match(/^tasks\/([^/]+)/);
+    if (!taskIdMatch) {
+      throw new Error(`Invalid task name format: ${params.name}`);
+    }
+    const taskId = taskIdMatch[1];
```
ishymko: Is it the same as in `src/client/transports/rest_transport.ts` where `import { extractTaskId } from '../../types/converters/id_decoding.js';` was used?

src/client/transports/transport.ts
ishymko, line 53: I believe `XxxParams` here are still JSON-RPC based types, while it should be proto-based.

I'd recommend to remove reexports of JSON-RPC types in `index.ts` to catch such things easier, also I don't think we need to export them in general, they should remain transport specific implementation detail while public interface should fully switch to proto types.

src/client/multitransport-client.ts
ishymko, line 1: Same as [here](https://github.com/a2aproject/a2a-js/pull/337/changes#r2904107492), I believe public API wasn't updated to proto-based types here.

src/server/transports/rest/rest_transport_handler.ts
ishymko, line 1: I believe this file also imports JSON-RPC types.

src/json_rpc_types.ts
```diff
@@ -0,0 +1,281 @@
...
+  | TaskResubscriptionRequest
+  | ListTaskPushNotificationConfigRequest
+  | DeleteTaskPushNotificationConfigRequest
+  | GetAuthenticatedExtendedCardRequest;
+
+interface BaseRequest {
+  id: string | number;
+  jsonrpc: '2.0';
+  method: string;
+  params?: { [k: string]: unknown } | JsonRpcTaskPushNotificationConfig;
```
ishymko: Why there is a special case for `JsonRpcTaskPushNotificationConfig`?

src/client/client.ts
bartek-gralewicz: It could but after the most recent changes, this logic is no longer a part of `client.ts`.

src/client/client.ts
bartek-gralewicz: It is almost the same (slightly different regex) but after the most recent changes, this is no longer needed in `client.ts`.

src/client/transports/transport.ts
bartek-gralewicz: You are right, sorry for leaving these in. I removed exports from `index.ts` and updated imports and types where JSON-RPC was still present.

src/json_rpc_types.ts
bartek-gralewicz: Because some interfaces (like `SetTaskPushNotificationConfigRequest`) require this type for `params`.

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -70,126 +70,126 @@ export class GrpcTransport implements Transport {
...
+    params: SendMessageRequest,
     options?: RequestOptions
   ): Promise<SendMessageResult> {
     const rpcResponse = await this._sendGrpcRequest(
       'sendMessage',
       params,
       options,
       this.grpcClient.sendMessage.bind(this.grpcClient),
-      ToProto.messageSendParams,
+      (req: SendMessageRequest) => req,
```
ishymko: Here and below: since it's echoing input should be possible to remove invocation inside `_sendXxx`.

src/client/transports/json_rpc_transport.ts
```diff
@@ -59,115 +74,155 @@ export class JsonRpcTransport implements Transport {
...
-    params: MessageSendParams,
+    params: SendMessageRequest,
     options?: RequestOptions,
     idOverride?: number
   ): Promise<SendMessageResult> {
+    const rpcParams: MessageSendParams = {
+      message: params.request!,
+      configuration: params.configuration,
+      metadata: params.metadata,
+    };
```
ishymko: Given that in 1.0 ProtoJSON format is used in JSON-RPC as well, it should be possible to follow the same pattern [REST client does](https://github.com/a2aproject/a2a-js/blob/d45c212ea6008a4b29ce36e8bec7e6561fce738b/src/client/transports/rest_transport.ts#L222-L244): accept `MessageFns<T>` and use `toJSON`/`fromJSON` for `params` field of the JSON RPC request.

src/client/transports/rest_transport.ts
```diff
@@ -49,150 +55,141 @@ export class RestTransport implements Transport {
...
     yield* this._sendStreamingRequest('/v1/message:stream', requestBody, options);
   }
 
   async setTaskPushNotificationConfig(
-    params: TaskPushNotificationConfig,
+    params: CreateTaskPushNotificationConfigRequest,
     options?: RequestOptions
   ): Promise<TaskPushNotificationConfig> {
-    const requestBody = ToProto.taskPushNotificationConfig(params);
+    const requestBody = params.config as TaskPushNotificationConfig;
```
ishymko: Seems like it's only due to `| undefined` so better to do

```suggestion
    const requestBody = params.config!;
```

src/client/client.ts
```diff
@@ -1,29 +1,24 @@
...
+import {
+  CancelTaskRequest,
+  CreateTaskPushNotificationConfigRequest,
+  DeleteTaskPushNotificationConfigRequest,
+  GetTaskPushNotificationConfigRequest,
+  GetTaskRequest,
+  ListTaskPushNotificationConfigRequest,
+  SendMessageRequest,
+  TaskSubscriptionRequest,
+} from '../types/pb/a2a_types.js';
```
ishymko: `index.ts` re-exports `a2a_types` and there are already imports from this file.

src/server/events/execution_event_queue.ts
```diff
@@ -40,8 +40,8 @@ export class ExecutionEventQueue {
         const event = this.eventQueue.shift()!;
         yield event;
         if (
-          event.kind === 'message' ||
-          (event.kind === 'status-update' && (event as TaskStatusUpdateEvent).final)
+          'messageId' in event ||
+          ('status' in event && 'taskId' in event && (event as TaskStatusUpdateEvent).final)
         ) {
```
ishymko, line 45: Let's create an issue under 1.0 epic (#321) to refactor this so that we can avoid such runtime checks, probably a wrapper event object with explicit event kind like before (a wrapper around proto-based objects which don't have kind anymore like previously in JSON Schema).

src/server/express/rest_handler.ts
```diff
@@ -19,10 +19,18 @@ import { HTTP_EXTENSION_HEADER } from '../../constants.js';
...
+  MessageFns,
+  SendMessageResponse,
+  StreamResponse,
+  Task,
+  TaskPushNotificationConfig,
+} from '../../types/pb/a2a_types.js';
 import { ToProto } from '../../types/converters/to_proto.js';
-import { Message, Task, TaskArtifactUpdateEvent, TaskStatusUpdateEvent } from '../../types.js';
+import { Message, TaskArtifactUpdateEvent, TaskStatusUpdateEvent } from '../../index.js';
+import { MessageSendParams } from '../../json_rpc_types.js';
```
ishymko: REST handler shouldn't import JSON RPC types.

src/server/request_handler/a2a_request_handler.ts
```diff
@@ -1,17 +1,19 @@
...
+import {
+  MessageSendParams,
   TaskQueryParams,
   TaskIdParams,
-  TaskPushNotificationConfig,
   GetTaskPushNotificationConfigParams,
   ListTaskPushNotificationConfigParams,
   DeleteTaskPushNotificationConfigParams,
-} from '../../types.js';
+} from '../../json_rpc_types.js';
```
ishymko: Transport agnostic A2A Handler shouldn't import JSON RPC types.

src/json_rpc_types.ts
ishymko, line 1: Should be possible to remove JSON-RPC types altogether (except for errors) and use generated pb functions `toJSON`/`fromJSON` for `params` content of  JSON-RPC payload.

bartek-gralewicz: I removed almost all JSON-RPC types and replaced them with `toJSON`/`fromJSON` methods. Remaining types are for errors and responses.

Response types will eventually need to be removed as well but to reduce the amount of changes done at once, I left these types for now.

src/server/request_handler/a2a_request_handler.ts
bartek-gralewicz: Corrected.

src/server/express/rest_handler.ts
bartek-gralewicz: Corrected.

src/client/client.ts
bartek-gralewicz: These imports are now taken directly from `../index.js`.

There is one JSON-RPC type imported here:
```
import { JsonRpcTransport } from './transports/json_rpc_transport.js';
```

I assume that this is ok since `A2AClient` is built on top of JSON-RPC transport. Please correct me if I am wrong here.

ishymko: Ah, right, it's this legacy client I mentioned [here](https://github.com/a2aproject/a2a-js/pull/337#discussion_r2854036016) which only supports `JsonRpcTransport`. We can remove this file separately.

src/server/events/execution_event_queue.ts
bartek-gralewicz: Added a new task under 1.0 epic: https://github.com/a2aproject/a2a-js/issues/351

src/client/transports/grpc/grpc_transport.ts
bartek-gralewicz: Removed in `_sendGrpcStreamingRequest` but left in `_sendGrpcRequest` as not all conversions are `A => A`.

Examples:
- `FromProto.listTaskPushNotificationConfig`
- `FromProto.sendMessageResult`

PR #346, opened 2026-03-10
fix: Fixing missing push configuration after normalization for rest transport

### LINKED ISSUE #336: [Bug]: Push Configuration Config gets removed from config in rest handler
### What happened?

We tried to use the rest handler including push notification config. Sadly there is a normalize function, removing the push configuration from the parameters before it get's evaluated and used within the rest handler.

Parameter definition: https://github.com/a2aproject/a2a-js/blob/main/src/server/transports/rest/rest_types.ts#L87
Normalize method: https://github.com/a2aproject/a2a-js/blob/main/src/server/transports/rest/rest_transport_handler.ts#L404
Access on the property within rest handler: https://github.com/a2aproject/a2a-js/blob/main/src/server/request_handler/default_request_handler.ts#L229

This definitely looks like a bug, that it get's removed.

### Relevant log output

```shell
no log output
```

### Code of Conduct

bartek-gralewicz:
# Description

This PR solves a bug that was mentioned here: #336.

The issue occurred due to missing mapping of `pushNotificationConfig` in the mapping method. Since mapping is no longer needed, all logic related to field mapping was removed in this PR.

Changes:

- Extension to an existing unit test to check for `pushNotificationConfig` presence. This test failed as expected before removing the normalization, hence, catching the bug for the future.
- Remove rest types and related unions.
- Removed snake_case cases from unit testes.
- Removed normalization functions but remained field checks to align with currently existing unit tests.

Fixes #336

### DIFF
--- a/test/client/card-resolver.spec.ts
+++ b/test/client/card-resolver.spec.ts
@@ -139,7 +139,7 @@ describe('DefaultAgentCardResolver', () => {
   });
 
   const expectedAgentCard: AgentCard = {
-    protocolVersion: '1.0.0',
+    protocolVersion: '0.3.0',
     name: 'Unified Agent',
     description: '',
     documentationUrl: undefined,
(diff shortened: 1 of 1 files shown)


src/server/transports/rest/rest_transport_handler.ts
```diff
@@ -152,23 +139,26 @@ export class RestTransportHandler {
...
     context: ServerCallContext
   ): Promise<Message | Task> {
-    const normalized = this.normalizeMessageParams(params);
-    return this.requestHandler.sendMessage(normalized, context);
+    if (!params.message) {
+      throw A2AError.invalidParams('message is required');
+    }
+    if (!params.message.messageId) {
+      throw A2AError.invalidParams('message.messageId is required');
+    }
```
ishymko, line 149: nit: maybe a function `validateMessage` shared with `sendMessageStream`?

bartek-gralewicz: Extracted `params` validation to new `validateMessageSendParams` function and used in both places.

PR #372, opened 2026-03-23
chore: Merge changes from the main branch to the v1.0 epic

bartek-gralewicz:
# Description

Merged recent changes done on the `main` branch to the `epic/1.0...` branch to keep these align for future development.


test/server/express/rest_handler.spec.ts
bartek-gralewicz, line 1: @ishymko, should I remove all "snake_case" tests from here?

src/client/card-resolver.ts
```diff
@@ -1,5 +1,6 @@
 import { AGENT_CARD_PATH } from '../constants.js';
 import { AgentCard } from '../index.js';
+import { AgentCard as PBAgentCard } from '../types/pb/a2a_types.js';
```
ishymko: I believe in this branch those AgentCards are the same? I believe this `normalizeAgentCard` is not required here.

test/server/express/rest_handler.spec.ts
ishymko: We can keep it as per ProtoJSON parsers should support both: https://protobuf.dev/programming-guides/json/.

It means that we're testing the parser here but given that the library we used didn't have this feature originally and we had to contribute ourselves it makes sense to keep it :D

ishymko (approved):

Issue #322, opened 2026-02-11
[Task]: 1.0: Use types generated from protos instead of types.ts


ishymko:
[0.3](https://github.com/a2aproject/A2A/tree/v0.3.0) spec defined [a2a.json](https://github.com/a2aproject/A2A/blob/v0.3.0/specification/json/a2a.json) JSON Schema which was used to generate [types.ts](https://github.com/a2aproject/a2a-js/blob/main/src/types.ts) via [this script](https://github.com/a2aproject/a2a-js/blob/5ea1c94acaf2cae5aa1bc10ea5816f5400a62423/package.json#L104).

[1.0 spec states](https://a2a-protocol.org/latest/specification/#14-normative-content) that
> SDK language bindings, schemas, and any other derived forms MUST be regenerated from the proto (directly or via code generation) rather than edited manually.

#292 introduced types generated from proto for REST transport with the idea that they could be used as a source of truth in the future in 1.0.

Python SDK already did it in https://github.com/a2aproject/a2a-python/pull/572.

It should use the latest 1.0 proto, so resolving this effectively means supporting 1.0 data model, see the above Python SDK PR and [the changelog](https://a2a-protocol.org/latest/whats-new-v1/) (note that at the moment of writing this it has inaccuracies, see pending fix https://github.com/a2aproject/A2A/pull/1460).

PR #375, opened 2026-03-25
feat!: Update codebase to use A2A 1.0.0 data model

bartek-gralewicz:
# Description

Update the codebase to use proto types generated for 1.0.0 spec version.

# List of changes

- Removed old type files: `a2a_services.ts` and `a2a_types.ts`.
- Update the codebase to use new types:
  -  `Message.content` -> `Message.parts`,
  - `TaskStatus.update` -> `TaskStatus.message`
  - `SendMessageConfiguration.blocking` -> `SendMessageConfiguration.returnImmediately` (reversed behavior),
  -  The AgentCard Structure updates:  `supportedInterfaces` instead of `url`, `preferredTransport`, and `additionalInterfaces`, `securityRequirements` instead of `security`.
- `setTaskPushNotificationConfig` -> `createTaskPushNotificationConfig` (+ arguments update).
- `listTasks` method is not implemented yet. If used, it will throw `NotImplemented` error. There is a `TODO` next to it to mark the place in code.

Fixes (partially) #179  #322 🦕


tck/agent/index.ts
```diff
@@ -164,22 +177,40 @@ class SUTAgentExecutor implements AgentExecutor {
 const SUTAgentCard: AgentCard = {
   name: 'SUT Agent',
   description: 'A sample agent to be used as SUT against tck tests.',
-  // Main URL points to JSON-RPC endpoint (preferred transport)
-  url: 'http://localhost:41241/a2a/jsonrpc',
+  supportedInterfaces: [
+    {
+      url: 'http://localhost:41241/a2a/jsonrpc',
+      protocolBinding: 'JSONRPC',
+      tenant: '',
```
ishymko, line 184: I think this is an existing issue but it's not very convenient that one has to provide and empty string for `tenant` (and other fields for scalar proto types).

The problem I believe is that a2a.proto uses `google.api.field_behavior` for optionality but [generator we use](https://github.com/stephenh/ts-proto) doesn't support it and emits all properties as non-optional which makes compiler demanding them.

Consider this non-blocking (maybe let's create an issue) as I don't have an answer how to approach it :) We may try to patch the generator (did it in the past once, but this change is more opinionated) or introduce some factory functions on our own.

src/server/transports/jsonrpc/jsonrpc_transport_handler.ts
ishymko: I think it requires update, `TaskPushNotificationConfig` is flat now: https://github.com/a2aproject/A2A/blob/c1169f45728ae5bf98d8ab0ab06e187f94002f15/specification/a2a.proto#L464-L478.

src/server/grpc/grpc_service.ts
```diff
@@ -242,6 +239,12 @@ export function grpcService(options: GrpcServiceOptions): A2AServiceServer {
         (res) => res
       );
     },
+    listTasks(
+      _call: grpc.ServerUnaryCall<ListTasksRequest, ListTasksResponse>,
+      _callback: grpc.sendUnaryData<ListTasksResponse>
+    ): Promise<void> {
+      throw new UnsupportedOperationError('Method listTasks not implemented yet.');
```
ishymko: Should it use `callback` to pass not implemented status?

src/server/events/execution_event_queue.ts
```diff
@@ -1,6 +1,13 @@
-import { TaskStatusUpdateEvent } from '../../index.js';
+import { TaskStatusUpdateEvent, TaskState } from '../../index.js';
 import { ExecutionEventBus, AgentExecutionEvent } from './execution_event_bus.js';
 
+const terminalStates: TaskState[] = [
+  TaskState.TASK_STATE_COMPLETED,
+  TaskState.TASK_STATE_FAILED,
+  TaskState.TASK_STATE_CANCELED,
+  TaskState.TASK_STATE_REJECTED,
+];
```
ishymko: There is one in `src/server/request_handler/default_request_handler.ts` already, maybe can be extracted into a shared location?

ishymko (approved):

src/server/transports/jsonrpc/jsonrpc_transport_handler.ts
bartek-gralewicz: Great catch! Thanks!

I believe it is now correct. Could you please verify?

ishymko: Yep, makes sense, the payload in `params` should be ProtoJSON formatted so can go straight to `fromJSON`.

PR #381, opened 2026-03-31
feat!: Remove obsolete JSON-RPC types

### LINKED ISSUE #179: [Cleanup]: 1.0 breaking changes

bartek-gralewicz:
# Description

Remove obsolete JSON-RPC types.

# List of changes

- Remove temporary JSON-RPC types.
- Use Proto types payload directly.

Fixes #179 🦕


ishymko (approved):

src/server/transports/jsonrpc/jsonrpc_transport_handler.ts
```diff
@@ -34,6 +33,18 @@ export type A2ARequest = {
...
+  code: number;
+  data?: { [k: string]: unknown };
+  message: string;
+}
+
+export interface JSONRPCErrorResponse {
+  error: JSONRPCError;
+  id: string | number | null;
+  jsonrpc: '2.0';
+}
```
ishymko: Defined for client as well I think, maybe can be extracted into common?

src/index.ts
```diff
@@ -8,7 +8,6 @@
 import { Message, Task, TaskStatusUpdateEvent, TaskArtifactUpdateEvent } from './types/pb/a2a.js';
 
 export * from './types/pb/a2a.js';
-export type { A2AResponse } from './a2a_response.js';
```
ishymko, line 11: Should probably cause `feat!` as we remove a public export.

ishymko (approved):

Issue #328, opened 2026-02-11
[Task]: 1.0: Update gRPC transport


ishymko:
Make sure that generated services correspond to the latest version. Should require minimal changes after #322.

Issue #327, opened 2026-02-11
[Task]: 1.0: Update REST transport


ishymko:
See https://github.com/a2aproject/A2A/pull/1389: compound id's like `tasks/{id}` are no longer used in 1.0.

Issue #326, opened 2026-02-11
[Task]: 1.0: Update JSON-RPC transport


ishymko:
JSON-RPC method names got changed in 1.0: [before](https://a2a-protocol.org/v0.3.0/specification/#356-method-mapping-reference-table), [after](https://a2a-protocol.org/latest/specification/#53-method-mapping-reference).

PR #401, opened 2026-04-07
feat: Align transport method names with 1.0 spec

bartek-gralewicz:
# Description

Align transport method names with 1.0 spec.

Aligning done based on the spec documentation: https://a2a-protocol.org/latest/specification/#53-method-mapping-reference

Fixes:
- #326
- #327
- #328

### DIFF
--- a/src/client/transports/rest_transport.ts
+++ b/src/client/transports/rest_transport.ts
@@ -63,7 +63,7 @@ export class RestTransport implements Transport {
   async getExtendedAgentCard(options?: RequestOptions): Promise<AgentCard> {
     const response = await this._sendRequest<undefined, AgentCard>(
       'GET',
-      '/v1/card',
+      '/v1/extendedAgentCard',
       undefined,
       options,
       undefined,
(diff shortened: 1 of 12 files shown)


src/client/transports/rest_transport.ts
```diff
@@ -63,7 +63,7 @@ export class RestTransport implements Transport {
   async getExtendedAgentCard(options?: RequestOptions): Promise<AgentCard> {
     const response = await this._sendRequest<undefined, AgentCard>(
       'GET',
-      '/v1/card',
+      '/v1/extendedAgentCard',
```
ishymko: Here and below: `v1` was removed from the path for REST in 1.0 spec (yeah... v1 remove in v1): https://a2a-protocol.org/latest/specification/#53-method-mapping-reference

bartek-gralewicz: Removed `/v1` prefix.

ishymko (approved):

PR #403, opened 2026-04-08
feat: Enable ITK tests

bartek-gralewicz:
# Description

Enable ITK tests.

Currently, the testing is done between the current a2a-js state and Go SDK 1.0 for the following configurations:
- Streaming
  - JSON-RPC / gRPC
  - HTTP_JSON
- Non-Streaming
  - JSON-RPC / gRPC
  - HTTP_JSON

# Notes

At the moment, there is an extensive middle layer in `itk_agent.ts`. This middle layer is planned to be removed once ITK tests are less spec dependent and a2a-js is closer to 1.0 release state.

Fixes (partially) #321 🦕

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -76,11 +76,32 @@ export class JsonRpcTransport implements Transport {
       SendMessageRequest
     );
 
-    if (!rpcResponse.result?.payload?.value) {
-      throw new Error('Invalid response structure from agent.');
+    const result = rpcResponse.result;
+    if (!result) {
+      throw new Error('Invalid response structure from agent: missing result.');
     }
 
-    return rpcResponse.result.payload.value;
+    let value: unknown;
+    const resultRecord = result as unknown as Record<string, unknown>;
+    if (result.payload?.value) {
+      value = result.payload.value;
+    } else if (resultRecord.task) {
+      value = resultRecord.task;
+    } else if (resultRecord.message) {
+      value = resultRecord.message;
+    } else if (resultRecord.id && resultRecord.status) {
+      // Looks like a Task directly
+      value = result;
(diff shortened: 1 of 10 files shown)


src/client/transports/json_rpc_transport.ts
```diff
@@ -78,11 +78,21 @@ export class JsonRpcTransport implements Transport {
...
     );
 
-    if (!rpcResponse.result?.payload?.value) {
-      throw new Error('Invalid response structure from agent.');
+    const result = rpcResponse.result as {
+      payload?: { value: unknown };
+      task?: unknown;
+      message?: unknown;
+    };
+    if (result?.payload?.value) {
```
guglielmo-san: can we use the ts-proto util fromJSON?
same should be done on the StreamResponse I think

bartek-gralewicz: Used `fromJSON` and `toJSON` where needed in client/server side of these methods. Thanks for spotting that!

src/server/transports/jsonrpc/jsonrpc_transport_handler.ts
```diff
@@ -118,22 +120,25 @@ export class JsonRpcTransportHandler {
...
-                payload = { $case: 'statusUpdate', value: event as TaskStatusUpdateEvent };
+                result = {
+                  statusUpdate: TaskStatusUpdateEvent.toJSON(event),
+                };
               } else if ('artifact' in event) {
-                payload = { $case: 'artifactUpdate', value: event as TaskArtifactUpdateEvent };
+                result = {
+                  artifactUpdate: TaskArtifactUpdateEvent.toJSON(event),
+                };
               }
```
guglielmo-san, line 137: let's add a TODO here to improve this conversion once agentEventStream will be AsyncGenerator of StreamResponse

guglielmo-san (approved):

PR #404, opened 2026-04-09
feat!: Remove specific transport exports

bartek-gralewicz:
# Description

Remove specific transport exports from the client's index file. We should only export generic transport types and transport specific factories.

Fixes (partially) #179 🦕

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -11,16 +11,9 @@ export {
 export { Client, type ClientConfig, type RequestOptions } from './multitransport-client.js';
 export type { Transport, TransportFactory } from './transports/transport.js';
 export { ClientFactory, ClientFactoryOptions } from './factory.js';
-export {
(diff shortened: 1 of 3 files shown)


ishymko (approved):

ishymko (commented):
Should probably also update gRPC [here](https://github.com/a2aproject/a2a-js/blob/epic/1.0_breaking_changes/src/client/transports/grpc/index.ts)?

(We have a separate entry point for it as it won't work in browser.)

bartek-gralewicz:
> Should probably also update gRPC [here](https://github.com/a2aproject/a2a-js/blob/epic/1.0_breaking_changes/src/client/transports/grpc/index.ts)?
> 
> (We have a separate entry point for it as it won't work in browser.)

Definitely! I missed this one, now it is updated as well.

PR #405, opened 2026-04-09
feat!: Make ServerCallContext parameter mandatory across all places

### LINKED ISSUE #179: [Cleanup]: 1.0 breaking changes

bartek-gralewicz:
# Description

Made `ServerCallContext` a mandatory parameter across all extension points.

Notes:

`InMemoryTaskStore` does not use the `ServerCallContext` but still accepts it as an argument to align with `TaskStore` interface which it implements.

Fixes #179  🦕

### DIFF
--- a/src/server/agent_execution/request_context.ts
+++ b/src/server/agent_execution/request_context.ts
@@ -13,15 +13,15 @@ export class RequestContext {
     userMessage: Message,
     taskId: string,
     contextId: string,
-    task: Task | undefined,
-    referenceTasks: Task[] | undefined,
-    context: ServerCallContext
+    context: ServerCallContext,
+    task?: Task,
+    referenceTasks?: Task[]
   ) {
     this.userMessage = userMessage;
     this.taskId = taskId;
     this.contextId = contextId;
+    this.context = context;
     this.task = task;
     this.referenceTasks = referenceTasks;
-    this.context = context;
   }
 }
--- a/src/server/agent_execution/request_context.ts
+++ b/src/server/agent_execution/request_context.ts
@@ -7,15 +7,15 @@ export class RequestContext {
   public readonly contextId: string;
   public readonly task?: Task;
   public readonly referenceTasks?: Task[];
-  public readonly context?: ServerCallContext;
+  public readonly context: ServerCallContext;
 
   constructor(
     userMessage: Message,
     taskId: string,
     contextId: string,
-    task?: Task,
-    referenceTasks?: Task[],
-    context?: ServerCallContext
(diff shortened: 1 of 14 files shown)


src/server/agent_execution/request_context.ts
```diff
@@ -7,15 +7,15 @@ export class RequestContext {
...
   constructor(
     userMessage: Message,
     taskId: string,
     contextId: string,
-    task?: Task,
-    referenceTasks?: Task[],
-    context?: ServerCallContext
+    task: Task | undefined,
+    referenceTasks: Task[] | undefined,
+    context: ServerCallContext
```
ishymko: Is it common to invoke it with `task` and/or `referenceTasks` being `undefined` in our code? In this case we may change params order to avoid code like `. . . undefined, undefined, context)`

bartek-gralewicz: I found only a single constructor usage in our codebase: [here](https://github.com/a2aproject/a2a-js/blob/7920e1447f68f70543e5f0f618c529ebee60b31e/src/server/request_handler/default_request_handler.ts#L166). All inputs are provided there but in case there will be some updates in the future, I changed the order in the constructor so that `context` is provided before optional arguments.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -32,7 +33,7 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
   }
 
   async send(task: Task): Promise<void> {
-    const pushConfigs = await this.pushNotificationStore.load(task.id);
+    const pushConfigs = await this.pushNotificationStore.load(task.id, new ServerCallContext());
```
ishymko: Hmm, we should rather propagate it from the above here as well.

bartek-gralewicz: Updated `PushNotificationSender` to include `context` and used context data from `default_request_handler` when PushNotificationSender is called.

guglielmo-san (approved):

PR #407, opened 2026-04-14
fix: allow DataPart to accept JSON values

### LINKED ISSUE #338: [Bug]: `DataPart` should accept any type as data, not just objects with string keys.
### What happened?

According to the [A2A spec's proto file](https://github.com/a2aproject/A2A/blob/629190aea9fe5838bb712008e6d24156680af197/specification/a2a.proto#L228), the `DataPart` accepts any object, not just maps, whereas the [JS implementation](https://github.com/a2aproject/a2a-js/blob/d4c261a4aad1b041965d2efe199e45942819d03a/src/types.ts#L679) specifically only allows objects with string keys.

Perhaps the JS implementation is still in the process of adopting the updated proto?

Bortlesboat:
## Summary
- widen DataPart.data to accept any JSON-compatible value instead of only string-keyed objects
- add converter regression tests for array and scalar data parts on the gRPC path
- add a focused TypeScript typecheck fixture that fails if arrays or scalars stop being accepted again

## Verification
- npx tsc --project test/typecheck/tsconfig.json
- npm test -- test/server/grpc/to_proto.spec.ts test/server/grpc/from_proto.spec.ts
- npx eslint test/server/grpc/to_proto.spec.ts test/server/grpc/from_proto.spec.ts test/typecheck/data_part_json_value.ts
- npx prettier --check test/server/grpc/to_proto.spec.ts test/server/grpc/from_proto.spec.ts test/typecheck/data_part_json_value.ts test/typecheck/tsconfig.json

Note: I did not run 
px prettier --check src/types.ts because the generated file is not currently in repo-wide Prettier baseline and the current formatter wants to restyle the whole file.

Closes #338

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,6 +49,7 @@ export type A2ARequest =
  * via the `definition` "Part".
  */
 export type Part = TextPart | FilePart | DataPart;
+export type JsonValue = string | number | boolean | null | JsonValue[] | { [k: string]: JsonValue };
 /**
  * Defines a security scheme that can be used to secure an agent's endpoints.
  * This is a discriminated union type based on the OpenAPI 3.0 Security Scheme Object.
@@ -675,9 +676,7 @@ export interface DataPart {
   /**
    * The structured data content.
    */
-  data: {
-    [k: string]: unknown;
-  };
+  data: JsonValue;
   /**
    * The type of this part, used as a discriminator. Always 'data'.
    */
--- a/test/server/grpc/from_proto.spec.ts
+++ b/test/server/grpc/from_proto.spec.ts
@@ -221,6 +221,21 @@ describe('FromProto', () => {
       expect(result).toEqual({ kind: 'data', data });
     });
 
+    it('should convert data parts with array and scalar values', () => {
+      const arrayData = ['foo', { nested: true }, 3];
+      const scalarData = 7;
+
+      const arrayPart = {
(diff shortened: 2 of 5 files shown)


Bortlesboat:
I traced the failing `tsc` error locally and the type mismatch is only the visible symptom.

Right now this branch widens `src/types.ts`, but the transport contract underneath is still object-only in two places:

- `src/types/pb/a2a_types.ts` still models `DataPart.data` as `{ [key: string]: any }`
- `src/grpc/pb/a2a_services.ts` still serializes `DataPart` through `Struct.wrap` / `Struct.unwrap`

So a cast-only fix in `to_proto.ts` would make CI greener without making scalar / array `DataPart` values actually work end-to-end over gRPC.

I?m keeping this as a draft for now rather than papering over that mismatch. The real next step is either:

1. update the source schema + gRPC protobuf contract together so `DataPart` can carry full JSON values at runtime, or
2. narrow this branch back to the object-only shape the current protocol actually supports.

Bortlesboat:
Closing — maintainer resolved #338 via the v1.0 proto upgrade (`google.protobuf.Value` instead of `Struct`) rather than a type-only widening. Widening `DataPart.data` to `JsonValue` here breaks the current v0.3 gRPC path that calls `Struct.wrap` in `src/grpc/pb/a2a_services.ts`, which only accepts objects. Better to wait for the 1.0 codegen bump.

PR #410, opened 2026-04-15
feat: Remove proto conversion methods for listTaskPushNotificationConfig type

bartek-gralewicz:
# Description

Remove `ToProto`/`FromProto` methods for `listTaskPushNotificationConfig` type.

### DIFF
--- a/src/client/transports/grpc/grpc_transport.ts
+++ b/src/client/transports/grpc/grpc_transport.ts
@@ -7,7 +7,7 @@ import {
   ListTaskPushNotificationConfigsRequest,
   SubscribeToTaskRequest,
 } from '../../../grpc/pb/a2a.js';
-import { Task, AgentCard } from '../../../types/pb/a2a.js';
+import { Task, AgentCard, ListTaskPushNotificationConfigsResponse } from '../../../types/pb/a2a.js';
 import {
   CancelTaskRequest,
   DeleteTaskPushNotificationConfigRequest,
@@ -153,7 +153,7 @@ export class GrpcTransport implements Transport {
       params,
       options,
       this.grpcClient.listTaskPushNotificationConfigs.bind(this.grpcClient),
-      FromProto.listTaskPushNotificationConfig
+      (res) => (res as ListTaskPushNotificationConfigsResponse).configs
     );
     return rpcResponse;
   }
--- a/src/client/transports/grpc/grpc_transport.ts
+++ b/src/client/transports/grpc/grpc_transport.ts
@@ -147,13 +147,17 @@ export class GrpcTransport implements Transport {
   async listTaskPushNotificationConfig(
     params: ListTaskPushNotificationConfigsRequest,
     options?: RequestOptions
-  ): Promise<TaskPushNotificationConfig[]> {
-    const rpcResponse = await this._sendGrpcRequest(
+  ): Promise<ListTaskPushNotificationConfigsResponse> {
+    const rpcResponse = await this._sendGrpcRequest<
+      ListTaskPushNotificationConfigsRequest,
+      ListTaskPushNotificationConfigsResponse,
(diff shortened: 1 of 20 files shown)


src/client/transports/grpc/grpc_transport.ts
```diff
@@ -153,7 +153,7 @@ export class GrpcTransport implements Transport {
       params,
       options,
       this.grpcClient.listTaskPushNotificationConfigs.bind(this.grpcClient),
-      FromProto.listTaskPushNotificationConfig
+      (res) => (res as ListTaskPushNotificationConfigsResponse).configs
```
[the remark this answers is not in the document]
bartek-gralewicz: That is not true. Without explicitly specifying the type, the helper treats `res` as `unknown` so `.config` cannot be extracted.

src/server/grpc/grpc_service.ts
```diff
@@ -171,7 +171,12 @@ export function grpcService(options: GrpcServiceOptions): A2AServiceServer {
         callback,
         (req) => req,
         requestHandler.listTaskPushNotificationConfigs.bind(requestHandler),
-        ToProto.listTaskPushNotificationConfig
+        (res) => {
+          return {
+            configs: res,
+            nextPageToken: '',
+          };
+        }
```
guglielmo-san: Can't we change the default_request_handler to return a ListTaskPushNotificationConfigsResponse ?

src/client/transports/grpc/grpc_transport.ts
guglielmo-san: Is there a reason to return only the configs and not the whole ListTaskPushNotificationConfigResponse? Even if we currently do not support pagination I think it makes more sense to return the right object

src/server/grpc/grpc_service.ts
bartek-gralewicz: Updated to return `ListTaskPushNotificationConfigsResponse`.

src/client/transports/grpc/grpc_transport.ts
bartek-gralewicz: Initially I wanted to remove `ToProto`/`FromProto` conversions and leave the logic unchanged but it is a good idea to tackle this already here and update the return type. The changes are applied as suggested.

src/server/grpc/grpc_service.ts
guglielmo-san, line 75: I think that we can remove/simplify the wrapUnary in a seperate PR, as of now the parser and converter are almost not used

src/client/transports/grpc/grpc_transport.ts
guglielmo-san, line 228: This converter can be removed now, as well the duplicate TRes and TRespose
Let's do that in a separate PR

src/client/transports/json_rpc_transport.ts
```diff
@@ -122,13 +122,12 @@ export class JsonRpcTransport implements Transport {
...
     options?: RequestOptions
-  ): Promise<TaskPushNotificationConfig[]> {
+  ): Promise<ListTaskPushNotificationConfigsResponse> {
     const rpcResponse = await this._sendRpcRequest<
       ListTaskPushNotificationConfigsRequest,
       ListTaskPushNotificationConfigsResponse
     >('ListTaskPushNotificationConfigs', params, options, ListTaskPushNotificationConfigsRequest);
-    const configs = rpcResponse.result.configs || [];
-    return configs.map((c: unknown) => TaskPushNotificationConfig.fromJSON(c));
+    return rpcResponse.result;
```
guglielmo-san:
```suggestion
    return ListTaskPushNotificationConfigsResponse.fromJSON(rpcResponse.result);

```

guglielmo-san (approved):

PR #413, opened 2026-04-16
feat: Use StreamResponse as push notification payload

### LINKED ISSUE #193: [Task]: 1.0: Use StreamResponse as push notifications payload

The current implementation of the [Push Notification Sender](https://github.com/a2aproject/a2a-js/blob/main/src/server/push_notification/push_notification_sender.ts) and the [Default Push Notification Sender](https://github.com/a2aproject/a2a-js/blob/main/src/server/push_notification/default_push_notification_sender.ts) only supports the task payload type.
According to the [A2A protocol specification](https://a2a-protocol.org/latest/specification/#433-push-notification-payload), multiple additional payload types such as artifactUpdate, message, and statusUpdate are allowed. These payloads are not supported in the SDK, which leads to incomplete protocol compliance and prevents integrators from using the full set of push notification features defined by the A2A standard.

1.0 specifies push notifications payload as `StreamResponse`: [4.3.3. Push Notification Payload](https://a2a-protocol.org/latest/specification/#433-push-notification-payload):

> The payload uses the same [StreamResponse](https://a2a-protocol.org/latest/specification/#323-stream-response) format as streaming operations, allowing push notifications to deliver the same event types as real-time streams.

### Code of Conduct

bartek-gralewicz:
# Description

- Update the push notification methods to use `StreamResponse` instead of relying only on `Task` object type.
- Replaced all `A2AStreamEventData` occurrences with `StreamResponse`.

By removing usage for `A2AStreamEventData` union type, dirty conversions were possible to be removed for JSON-RPC transport.

Fixes #193  🦕

### DIFF
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -12,6 +12,7 @@ import {
   GetTaskPushNotificationConfigRequest,
   GetTaskRequest,
   ListTaskPushNotificationConfigsRequest,
+  ListTaskPushNotificationConfigsResponse,
   SendMessageConfiguration,
   SendMessageRequest,
   SubscribeToTaskRequest,
@@ -213,7 +214,7 @@ export class Client {
   listTaskPushNotificationConfig(
     params: ListTaskPushNotificationConfigsRequest,
     options?: RequestOptions
-  ): Promise<TaskPushNotificationConfig[]> {
+  ): Promise<ListTaskPushNotificationConfigsResponse> {
     if (!this.agentCard.capabilities?.pushNotifications) {
       throw new PushNotificationNotSupportedError();
     }
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -140,10 +140,10 @@ export class Client {
       const result = await this.transport.sendMessage(beforeArgs.input.value, beforeArgs.options);
 
       let streamValue: StreamResponse;
-      if ('artifacts' in result) {
-        streamValue = { payload: { $case: 'task', value: result } };
-      } else {
+      if ('messageId' in result) {
         streamValue = { payload: { $case: 'message', value: result } };
+      } else {
+        streamValue = { payload: { $case: 'task', value: result } };
       }
 
       const afterArgs: AfterArgs<'sendMessageStream'> = {
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -15,6 +15,7 @@ import {
   ListTaskPushNotificationConfigsResponse,
   SendMessageConfiguration,
   SendMessageRequest,
+  StreamResponse,
   SubscribeToTaskRequest,
   ListTasksRequest,
   ListTasksResponse,
@@ -143,8 +144,16 @@ export class Client {
 
     if (!this.agentCard.capabilities?.streaming) {
       const result = await this.transport.sendMessage(beforeArgs.input.value, beforeArgs.options);
+
+      let streamValue: StreamResponse;
+      if ('artifacts' in result) {
+        streamValue = { payload: { $case: 'task', value: result } };
+      } else {
+        streamValue = { payload: { $case: 'message', value: result } };
+      }
+
       const afterArgs: AfterArgs<'sendMessageStream'> = {
-        result: { method, value: result },
+        result: { method, value: streamValue },
         agentCard: this.agentCard,
         options: beforeArgs.options,
       };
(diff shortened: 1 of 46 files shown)


src/client/multitransport-client.ts
```diff
@@ -143,8 +138,16 @@ export class Client {
 
     if (!this.agentCard.capabilities?.streaming) {
       const result = await this.transport.sendMessage(beforeArgs.input.value, beforeArgs.options);
+
+      let streamValue: StreamResponse;
+      if ('artifacts' in result) {
```
ishymko: Is it a safe way to differentiate? I believe a `Task` may not have any artifacts, maybe we can check for `messageId` instead?

src/server/grpc/grpc_service.ts
```diff
@@ -131,7 +131,7 @@ export function grpcService(options: GrpcServiceOptions): A2AServiceServer {
         call,
         (req) => req,
         requestHandler.sendMessageStream.bind(requestHandler),
-        ToProto.messageStreamResult
+        (res) => res
```
ishymko, line 134: Maybe at this point we can drop it everywhere? And/or `(req) => req`?

src/server/request_handler/default_request_handler.ts
```diff
@@ -412,8 +410,10 @@ export class DefaultRequestHandler implements A2ARequestHandler {
     try {
       for await (const event of eventQueue.events()) {
         await resultManager.processEvent(event); // Update store in background
-        await this._sendPushNotificationIfNeeded(event, context);
-        yield event; // Stream the event to the client
+        const streamResponse = await this._sendPushNotificationIfNeeded(event, context);
+        if (streamResponse) {
+          yield streamResponse; // Stream the event to the client
```
ishymko: I feel like we should not rely on push notification sending to convert between inner model and `StreamResponse`. Structurally push notification sending is a secondary functionality when comparing to response streaming.

Maybe we should use `_mapEventToStreamResponse` here and pass the result into `_sendPushNotificationIfNeeded`?

src/types/converters/to_proto.ts
```diff
@@ -10,34 +10,37 @@ import {
 
 export class ToProto {
   static messageStreamResult(
-    event: Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent
+    event: Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent | StreamResponse
```
ishymko, line 13: Should we start getting rid of it as now everything is proto?

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -32,42 +32,67 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
 
+  private _getTaskId(streamResponse: StreamResponse): string {
+    const payload = streamResponse.payload;
+    if (!payload) {
+      throw new Error('StreamResponse payload is undefined');
+    }
+    switch (payload.$case) {
+      case 'task':
+        return payload.value.id;
+      case 'message':
```
ishymko: I don't think we need `message` here at all, push notifications are going to be sent for running tasks and `Message` can't be there: see _Behavior_ section [here](https://a2a-protocol.org/latest/specification/#312-send-streaming-message).

src/server/grpc/grpc_service.ts
bartek-gralewicz: I already added simplification changes to methods here and in `grpc_transport` under `bgralewicz/implement_event_wrapper` branch. It will be the next open PR after this one.

src/types/converters/to_proto.ts
bartek-gralewicz: These changes are a part of `bgralewicz/implement_event_wrapper` branch that is planned to be open after this PR. Once the event wrapper is present, many more occurrences can be removed/simplified.

src/client/multitransport-client.ts
bartek-gralewicz: `artifacts` is a required member of `Task` object (it can be an empty list though).

I replaced the check for `messageId` as suggested.

src/server/request_handler/default_request_handler.ts
bartek-gralewicz: Used `_mapEventToStreamResponse` outside of `_sendPushNotificationIfNeeded` and passing `streamResponse` as a new argument.

src/server/push_notification/default_push_notification_sender.ts
bartek-gralewicz: Due to new `_exhaustive: never` check in this switch, I need to keep `case 'message'` present. I extracted it to a separate case where the error is thrown: `'Push notification should not be sent for message payload.'`. Relevant unit test was also updated for such case.

ishymko (approved):

PR #419, opened 2026-04-21
feat: Support multi-tenancy

### LINKED ISSUE #325: [Task]: 1.0: Support multi-tenancy
https://github.com/a2aproject/A2A/pull/1195 updates core proto with extra HTTP bindings prefixed with `/{tenant}`. It should be supported.

It's not fully clear how to represent it for JSON-RPC, also HTTP+JSON/REST protocol binding in the spec doesn't mention it, see https://github.com/a2aproject/A2A/issues/1273 and https://a2a-protocol.org/latest/specification/#53-method-mapping-reference.

bartek-gralewicz:
# Description

Added support for multi-tenancy.

 ## Multi-Tenancy

 This handler supports multi-tenant deployments through the `tenant` field present
 on all request objects (per A2A spec Sections 3.1.x and 4.4.6). The tenant value
 flows through the system as follows:

 1. **Transport layer** extracts tenant from the protocol-specific source:
    - REST: URL path prefix (`/:tenant/...`)
    - JSON-RPC: `params.tenant` in the request body
    - gRPC: `tenant` field in the request message

 2. **`ServerCallContext.tenant`** carries the tenant to all downstream components,
    including `TaskStore`, `PushNotificationStore`, and `AgentExecutor`.

 3. **`InMemoryTaskStore`** and **`InMemoryPushNotificationStore`** use `context.tenant`
    to scope data with composite keys (`{tenant}:{id}`), providing tenant isolation.
 
Similar PR done in Python SDK: https://github.com/a2aproject/a2a-python/pull/758

Fixes #325 🦕

### DIFF
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -77,13 +77,19 @@ export class Client {
   /**
    * If the current agent card supports the extended feature, it will try to fetch the extended agent card from the server,
    * Otherwise it will return the current agent card value.
+   *
+   * The tenant for the request is inferred from the agent card's interface declaration.
    */
   async getAgentCard(options?: RequestOptions): Promise<AgentCard> {
     if (this.agentCard.capabilities?.extendedAgentCard) {
+      const tenant =
+        this.agentCard.supportedInterfaces?.find(
+          (iface) => iface.protocolBinding === this.transport.protocolName
+        )?.tenant ?? '';
       this.agentCard = await this.executeWithInterceptors(
         { method: 'getAgentCard' },
         options,
-        (_, options) => this.transport.getExtendedAgentCard(options)
+        (_, options) => this.transport.getExtendedAgentCard({ tenant }, options)
       );
     }
     return this.agentCard;
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -78,18 +78,16 @@ export class Client {
    * If the current agent card supports the extended feature, it will try to fetch the extended agent card from the server,
    * Otherwise it will return the current agent card value.
    *
-   * The tenant for the request is inferred from the agent card's interface declaration.
+   * When a default tenant is configured (via `TenantTransportDecorator`, wired
+   * automatically by `ClientFactory` from `AgentInterface.tenant`), the tenant
+   * is applied to the request transparently.
    */
   async getAgentCard(options?: RequestOptions): Promise<AgentCard> {
     if (this.agentCard.capabilities?.extendedAgentCard) {
-      const tenant =
-        this.agentCard.supportedInterfaces?.find(
-          (iface) => iface.protocolBinding === this.transport.protocolName
-        )?.tenant ?? '';
       this.agentCard = await this.executeWithInterceptors(
         { method: 'getAgentCard' },
         options,
-        (_, options) => this.transport.getExtendedAgentCard({ tenant }, options)
+        (_, options) => this.transport.getExtendedAgentCard({ tenant: '' }, options)
       );
     }
     return this.agentCard;
--- a/src/server/express/rest_handler.ts
+++ b/src/server/express/rest_handler.ts
@@ -264,7 +264,60 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
   // ============================================================================
 
   /**
+   * Middleware that resolves tenant from the URL path parameter and normalizes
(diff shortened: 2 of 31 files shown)


src/server/grpc/grpc_service.ts
```diff
@@ -241,15 +243,17 @@ const mapToError = (error: unknown): Partial<grpc.ServiceError> => {
...
 ): Promise<ServerCallContext> => {
   const user = await userBuilder(call);
   const extensionHeaders = call.metadata.get(HTTP_EXTENSION_HEADER);
   const extensionString = extensionHeaders.map((v) => v.toString()).join(',');
-
-  return new ServerCallContext(Extensions.parseServiceParameter(extensionString), user);
+  const tenant = (call.request as Record<string, unknown>)?.tenant as string | undefined;
+  return tenant
+    ? new ServerCallContext(Extensions.parseServiceParameter(extensionString), user, tenant)
+    : new ServerCallContext(Extensions.parseServiceParameter(extensionString), user);
```
ishymko: nit: I think should be possible to just do

```suggestion
  return new ServerCallContext(Extensions.parseServiceParameter(extensionString), user, tenant);
```

as `tenant` is optional

src/server/push_notification/push_notification_store.ts
ishymko, line 1: Here and for task store: let's avoid introducing extra constraints on the task id names to properly parse composite string keys and have nested maps instead.

src/client/multitransport-client.ts
ishymko, line 1: Let's also update factory in the same way as [in Python](https://github.com/a2aproject/a2a-python/blob/8a0f38df17fcce71ba1a66e63bbcb6121bf44378/src/a2a/client/client_factory.py#L354-L357) to support ["default tenant" from the `AgentInterface`](https://a2a-protocol.org/latest/specification/#446-agentinterface) so that it's added to all requests automatically when specified.

src/server/express/rest_handler.ts
```diff
@@ -380,14 +412,24 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
-      sendResponse<ListTasksResponse>(res, HTTP_STATUS.OK, context, result, ListTasksResponse);
-    })
-  );
+  registerRoute('get', '/tasks', async (req, res) => {
+    const context = await buildContext(req);
+    const query = { ...req.query };
+    // For listTasks, tenant comes from the URL path (not from query params).
+    // The path tenant is already in the context; pass it along in the query object
+    // so RestTransportHandler can include it in the ListTasksRequest.
+    if (req.params.tenant) {
```
ishymko: I think the same is true for GetTask but we don't have this special handling there. Maybe `applyPathTenant` can be updated (and possible renamed :D) to support query params as well?

src/server/express/rest_handler.ts
```diff
@@ -260,6 +263,40 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
+  /**
+   * Helper to register routes with and without optional tenant prefix.
+   */
+  const registerRoute = (
+    method: 'get' | 'post' | 'delete' | 'put',
+    path: string,
+    handler: AsyncRouteHandler
+  ) => {
+    router[method](path, asyncHandler(handler));
+    router[method](`/:tenant${path}`, asyncHandler(handler));
```
ishymko: Related to https://github.com/a2aproject/a2a-js/pull/419/changes#r3129246003: maybe we can do it in a way that `asyncHandler` for tenant prefix route has an extra middleware/wrapper which deals with a tenant before passing it down? A modified version of `applyPathTenant` which is able to deal with both path, body and query depending on the HTTP verb. This way there will be no need to do `applyPathTenant` explicitly in each method.

src/server/push_notification/push_notification_store.ts
bartek-gralewicz: Used nested map instead of relying on a naming schema

src/client/multitransport-client.ts
bartek-gralewicz: Added dedicated transport decorated (similar way as in Python SDK) so now default tenants are automatically set.

src/server/express/rest_handler.ts
bartek-gralewicz: Added the logic of `tenantMiddleware` wrapper layer. No need of applying tenants individually anymore.

ishymko (approved):

PR #422, opened 2026-04-23
feat: A2A Version Header

bartek-gralewicz:
# Description

Added an A2A Version Header. The followings are now true:
- Clients send A2A-version header with each request,
- Server validates version and return `VersionNotSupportedError` if unsupported,
- Empty/absent versions are interpreted as 0.3.

### DIFF
--- a/itk/itk_agent.ts
+++ b/itk/itk_agent.ts
@@ -526,8 +526,26 @@ async function main() {
         protocolVersion: '1.0',
       },
       {
+        url: `http://127.0.0.1:${httpPort}/jsonrpc`,
+        protocolBinding: 'JSONRPC',
+        tenant: '',
+        protocolVersion: '0.3',
+      },
+      {
         url: `127.0.0.1:${grpcPort}`,
         protocolBinding: 'GRPC',
+        tenant: '',
+        protocolVersion: '1.0',
+      },
+      {
+        url: `127.0.0.1:${grpcPort}`,
+        protocolBinding: 'GRPC',
+        tenant: '',
+        protocolVersion: '0.3',
+      },
+      {
+        url: `http://127.0.0.1:${httpPort}/rest`,
+        protocolBinding: 'HTTP+JSON',
         tenant: '',
         protocolVersion: '1.0',
       },
@@ -535,7 +553,7 @@ async function main() {
         url: `http://127.0.0.1:${httpPort}/rest`,
         protocolBinding: 'HTTP+JSON',
         tenant: '',
-        protocolVersion: '1.0',
+        protocolVersion: '0.3',
       },
     ],
 
--- a/src/client/multitransport-client.ts
+++ b/src/client/multitransport-client.ts
@@ -1,3 +1,4 @@
+import { A2A_VERSION_HEADER, A2A_PROTOCOL_VERSION } from '../constants.js';
 import { PushNotificationNotSupportedError } from '../errors.js';
 import { TaskPushNotificationConfig, Task, AgentCard, SendMessageResult } from '../index.js';
 import {
@@ -68,11 +69,33 @@ export interface RequestOptions {
 }
 
 export class Client {
+  /**
+   * The A2A protocol version sent with every request via the A2A-Version header.
+   * Derived from the agent card's matching interface protocolVersion,
+   * falling back to {@link A2A_PROTOCOL_VERSION} if no match is found.
+   * Clients MUST send this header per §3.6.1.
+   */
+  public readonly protocolVersion: string;
+
   constructor(
     public readonly transport: Transport,
     private agentCard: AgentCard,
     public readonly config?: ClientConfig
-  ) {}
+  ) {
+    this.protocolVersion = this.resolveProtocolVersion();
+  }
+
+  /**
+   * Resolves the protocol version from the agent card's supported interfaces
+   * matching the current transport's protocol name. Falls back to the SDK's
+   * built-in protocol version if no match is found.
+   */
+  private resolveProtocolVersion(): string {
+    const agentInterface = this.agentCard.supportedInterfaces?.find(
+      (i) => i.protocolBinding.toUpperCase() === this.transport.protocolName.toUpperCase()
+    );
+    return agentInterface?.protocolVersion || A2A_PROTOCOL_VERSION;
+  }
 
   /**
    * If the current agent card supports the extended feature, it will try to fetch the extended agent card from the server,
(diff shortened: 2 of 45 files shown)


src/server/context.ts
```diff
@@ -1,14 +1,46 @@
...
+  constructor(
+    extensionsOrOptions?: Extensions | ServerCallContextOptions,
+    user?: User,
+    requestedVersion?: string
+  ) {
+    if (
+      extensionsOrOptions &&
+      !Array.isArray(extensionsOrOptions) &&
+      typeof extensionsOrOptions === 'object'
+    ) {
```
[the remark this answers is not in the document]
bartek-gralewicz: `Extension` is not a class. It cannot be used for `instanceof` check.

src/client/multitransport-client.ts
```diff
@@ -68,12 +69,34 @@ export interface RequestOptions {
...
+   * Resolves the protocol version from the agent card's supported interfaces
+   * matching the current transport's protocol name. Falls back to the SDK's
+   * built-in protocol version if no match is found.
+   */
+  private resolveProtocolVersion(): string {
+    const agentInterface = this.agentCard.supportedInterfaces?.find(
+      (i) => i.protocolBinding.toUpperCase() === this.transport.protocolName.toUpperCase()
+    );
+    return agentInterface?.protocolVersion || A2A_PROTOCOL_VERSION;
+  }
```
ishymko: I am not sure we need this, it can get problematic as we can have multiple interfaces for the same protocol name and different version.

What we maybe can do is to expose version in the transport interface (as we already did with `protocolName`)? Transport is a public property of `Client` so one can examine it if required.

Python currently has all the dispatching logic [in the factory](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/client_factory.py) and transport abstraction works quite well with different protocol versions.

src/client/multitransport-client.ts
```diff
@@ -353,6 +376,26 @@ export class Client {
...
+      return options!;
+    }
+    return {
+      ...options,
+      serviceParameters: {
+        [A2A_VERSION_HEADER]: this.protocolVersion,
+        ...existing,
+      },
+    };
+  }
```
ishymko, line 378: We have some machinery for service parameters [here](https://github.com/a2aproject/a2a-js/blob/main/src/client/service-parameters.ts).

I think version header can be defined as a `ServiceParametersUpdate` and used natively with `ServiceParameters.createFrom`.

```ts
export function withA2AVersion(version: string): ServiceParametersUpdate {
  return (parameters: ServiceParameters) => {
    parameters[A2A_VERSION_HEADER] = version;
  };
}

ServiceParameters.createFrom(options?.serviceParameters, withA2AVersion(this.protocolVersion))
```

src/server/version.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param agentCard - The agent card to extract versions from.
+ * @param protocolBinding - The protocol binding to filter versions by.
+ * @returns A Set of supported version strings (Major.Minor format).
+ */
+export function getSupportedVersions(
+  agentCard: AgentCard,
+  protocolBinding?: TransportProtocolName
+): Set<string> {
+  const versions = new Set<string>();
+  versions.add(A2A_DEFAULT_VERSION);
```
ishymko: Doing this means that any agent card supports 0.3. I believe it's not the case.

src/constants.ts
```diff
@@ -12,6 +12,26 @@ export const AGENT_CARD_PATH = '.well-known/agent-card.json';
...
+ * The A2A protocol version implemented by this SDK (Major.Minor).
+ * Patch version numbers SHOULD NOT be used per §3.6.
+ */
+export const A2A_PROTOCOL_VERSION = '1.0';
+
+/**
+ * The default A2A version assumed when the A2A-Version header is empty or absent.
+ * Per §3.6.2: "Agents MUST interpret empty value as 0.3 version."
+ */
+export const A2A_DEFAULT_VERSION = '0.3';
```
ishymko: I'd have this constant as a local one when parsing headers, with https://github.com/a2aproject/a2a-js/pull/422/changes#r3131393526 I believe it should have only one usage.

Naming it default feels like a mixed signal to me.

src/server/context.ts
```diff
@@ -1,16 +1,56 @@
...
+    requestedVersion?: string,
+    tenant?: string
+  );
+  constructor(options: ServerCallContextOptions);
+  constructor(
+    extensionsOrOptions?: Extensions | ServerCallContextOptions,
+    user?: User,
+    requestedVersion?: string,
+    tenant?: string
+  ) {
```
ishymko: Let's just use and support options only, for 1.0 we can do a breaking change and supporting both requires things which are not very nice.

src/client/multitransport-client.ts
bartek-gralewicz: `protocolVersion` exposed via transport interface.

src/client/multitransport-client.ts
bartek-gralewicz: Applied changes as suggested.

Note: The transport's `protocolVersion` is used even if the user provides the `A2A-Version`.

src/server/version.ts
bartek-gralewicz: Removed the defaulting mechanism. Now, all supported protocol versions must be provided explicitly.

src/constants.ts
bartek-gralewicz: Changed to local and renamed to `ABSENT_HEADER_VERSION`.

src/server/context.ts
bartek-gralewicz: Used `ServerCallContextOptions` as the only source of values.

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -69,6 +72,10 @@ export class GrpcTransport implements Transport {
     return PROTOCOL_NAME;
   }
 
+  get protocolVersion(): string {
+    return this._protocolVersion;
+  }
```
ishymko, line 75: I think it should rather return a static value for this transport (1.0, and we'll use 0.3 for compat transports). While we can have it as a constructor parameter and a field, I think that in practice it's better to have different versions implemented independently in different classes. Same for other transports.

bartek-gralewicz: Good point! Applied.

For compat layer, I'll most likely write new transport classes that will import the old v0.3 version.

itk/itk_agent.ts
ishymko, line 1: Hmm, why do we need it? I think in fact it's not the case that this ITK agent supports 0.3 protocol version?

ishymko (approved):

PR #425, opened 2026-04-24
fix(security): bump uuid to v14

TheIsrael1:
## Summary
This PR updates `@a2a-js/sdk`'s direct `uuid` dependency from `^11.1.0` to `^14.0.0` and adds the minimal Node 18 crypto shims needed for this repo's CI environments.

## Problem
The package currently depends on `uuid:^11.1.0`, which is flagged by Snyk advisory `SNYK-JS-UUID-16133035` for versions earlier than `14.0.0`:
https://security.snyk.io/vuln/SNYK-JS-UUID-16133035

For downstream consumers, `@a2a-js/sdk` can therefore show up as a meaningful source of `uuid` exposure in dependency scans.

## Why this PR takes the version-bump route
I tested the published `uuid@14` package directly against this repo's current setup and the package runtime/build path works.

The issue I hit was specific to how this repo executes source in CI on Node 18:
- the Vitest job did not provide `globalThis.crypto` consistently
- the TCK job runs `tck/agent/index.ts` via `tsx`, and that source-run path also hit `crypto is not defined`

This PR keeps the actual fix scoped to the dependency upgrade, and only adds small environment shims where CI needs Web Crypto on Node 18.

## Scope
This PR changes:
- `package.json`
- `package-lock.json`
- `vitest.config.ts`
- `test/setup.ts`
- `tck/agent/index.ts`

No protocol behavior or public API is changed.

## Validation
Ran locally after upgrading to `uuid@14`:
- `npm test`
- `npm run build`
- `npm run test-build`
- `npx -y node@18 node_modules/tsx/dist/cli.mjs tck/agent/index.ts` plus a JSON-RPC `message/send` request to confirm the Node 18 TCK path no longer fails with `crypto is not defined`

All passed.

## Why not rewrite UUID generation instead?
I also explored replacing `uuid` usage with `crypto.randomUUID()`, but that is a meaningfully larger change and not necessary here.

Given that `uuid@14` works, the smaller upstream fix is:
1. bump the dependency to the non-vulnerable major
2. provide Web Crypto in the Node 18 CI environments that execute source directly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,6 @@
       "name": "@a2a-js/sdk",
       "version": "0.3.13",
       "license": "Apache-2.0",
-      "dependencies": {
-        "uuid": "^11.1.0"
-      },
       "devDependencies": {
         "@betterer/cli": "^6.0.0-alpha.1",
         "@betterer/typescript": "^6.0.0-alpha.1",
@@ -7647,19 +7644,6 @@
       "license": "BSD-2-Clause",
       "dependencies": {
         "punycode": "^2.1.0"
-      }
-    },
-    "node_modules/uuid": {
-      "version": "11.1.0",
-      "resolved": "https://registry.npmjs.org/uuid/-/uuid-11.1.0.tgz",
-      "integrity": "sha512-0/A9rDy9P7cJ+8w1c9WD9V//9Wj15Ce2MPz8Ri6032usz+NfePxx5AcN3bN+r6ZL6jEo066/yNYB3tn4pQEx+A==",
-      "funding": [
-        "https://github.com/sponsors/broofa",
-        "https://github.com/sponsors/ctavan"
-      ],
-      "license": "MIT",
-      "bin": {
-        "uuid": "dist/esm/bin/uuid"
       }
     },
     "node_modules/vary": {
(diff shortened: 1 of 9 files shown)


ishymko:
Hi @TheIsrael1, thanks for the PR!

Taking one step back: reading the https://security.snyk.io/vuln/SNYK-JS-UUID-16133035 report: 

> allows silent partial writes into caller-provided buffers

The code which uses uuid does not use caller-provided buffers and is not affected by this vulnerability. In addition, it only uses v4 which is not affected by the issue: https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq.

Dropping node 18 support with bump to uuid@14 is a breaking change though, shims are fixing this for CI, but in real deployment it's not going to work.

That being said, we can still bump version in 1.0 dev branch (`epic/1.0_breaking_changes`) as we dropped node 18 support there already.

But given that in fact this code is not affected by this vulnerability, I think fixing it in a breaking way for 0.3.x build is not reasonable.

TheIsrael1:
Hi @ishymko thanks for reviewing this! I had my coding agent do most of the research and make the PR! good catch on your end 🙏

Yes the 1.0 branch looks like a good place to bump it since it drops support for node 18. Thanks.

PR #427, opened 2026-04-27
feat: Enriched Error Model

bartek-gralewicz:
# Description

Update error handling to the new enriched model. These changes are compliant with https://a2a-protocol.org/latest/specification/#332-error-handling and https://a2a-protocol.org/latest/specification/#116-error-handling.

# Bug fix

Error code for `TaskNotCancelableError` was previously set to 409 when it should be 400. This PR fixes this issue.

# Notes

There are a few new proto-generated files which are auto-generated based on updated `bug.gen.yaml` and were not modified manually. Each one starts with a note `// Code generated by protoc-gen-ts_proto. DO NOT EDIT.`.


src/client/transports/rest_transport.ts
```diff
@@ -402,29 +435,23 @@ export class RestTransport implements Transport {
...
+    const details = error.error?.details;
+    if (Array.isArray(details)) {
+      const errorInfo = details.find((d) => d['@type'] === ERROR_INFO_TYPE);
+      if (errorInfo && typeof errorInfo['reason'] === 'string') {
+        const ErrorClass = A2A_REASON_TO_ERROR_CLASS[errorInfo['reason'] as string];
+        if (ErrorClass) return new ErrorClass(message);
       }
     }
 
+    // Strategy 2: Legacy flat format — match by error class name.
```
ishymko: Let's remove all the legacy handling as it's the new standard now, we can do it for the legacy 0.3 compat transport. But I'd prefer to get rid of it here, either way it wasn't specified properly in 0.3 and earlier SDK designs sometimes did not expose domain error information properly at all.

test/client/transports/grpc_transport.spec.ts
ishymko, line 1: Can we update https://github.com/a2aproject/a2a-js/blob/epic/1.0_breaking_changes/test/e2e.spec.ts to make sure that client is able to parse errors returned by the server?

I'd also think about updating ITK to cover this in cross-SDK testing...

src/client/transports/rest_transport.ts
```diff
@@ -46,11 +49,19 @@ export interface RestTransportOptions {
...
+  message?: string;
+  details?: Array<Record<string, unknown>>;
+}
+
 interface RestErrorResponse {
   name?: string;
   message?: string;
   code?: number;
   data?: Record<string, unknown>;
+  error?: RestRpcStatus;
```
ishymko: Let's update it to match https://a2a-protocol.org/latest/specification/#116-error-handling.

With [this](https://github.com/a2aproject/a2a-js/pull/427/changes#r3154018043) it should only be `error` at the top level?

src/client/transports/rest_transport.ts
bartek-gralewicz: Legacy types removed.

test/client/transports/grpc_transport.spec.ts
bartek-gralewicz: Updated the e2e test and ITK suite.

Note regarding ITK:
ITK is controlled via Python orchestrator that is on another repository. At the moment, without external modifications, I could not add error checking using standard ITK. I created another test and added it to the ITK workflow. It spawns an agent and checks for multiple error cases.

Once ITK includes such functionality, this test can be transferred there.

src/client/transports/rest_transport.ts
bartek-gralewicz: You are right! Updated the interface to match the spec.

src/client/transports/rest_transport.ts
bartek-gralewicz: Removed `httpStatus` fallback as well as it is no longer needed.

test/client/transports/grpc_transport.spec.ts
ishymko: I think it mostly duplicates lower level tests, let's rely on ITK for this if/when such cases are added there. The main point of ITK tests is that it tests cross-SDK compatibility, while we can assert against static error codes and messages in normal unit tests.

ishymko (approved):

PR #429, opened 2026-04-27
feat: Content type update to application/a2a+json

bartek-gralewicz:
# Description

Updated the content type from `application/json` to `application/a2a+json`.

# Important Note:

The spec mentions that the REST SHOULD use `application/a2a+json` (not MUST). Due to this, the server side needs to accept both versions.

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -17,6 +17,7 @@ import {
   TaskPushNotificationConfig,
   SendMessageResult,
   A2A_PROTOCOL_VERSION,
+  A2A_CONTENT_TYPE,
 } from '../../index.js';
 import { RequestOptions } from '../multitransport-client.js';
 import { parseSseStream } from '../../sse_utils.js';
@@ -233,7 +234,7 @@ export class JsonRpcTransport implements Transport {
       id: requestId,
     };
 
-    const httpResponse = await this._fetchRpc(rpcRequest, 'application/json', options);
+    const httpResponse = await this._fetchRpc(rpcRequest, A2A_CONTENT_TYPE, options);
 
     if (!httpResponse.ok) {
       let errorBodyText = '(empty or non-JSON response)';
@@ -273,14 +274,14 @@ export class JsonRpcTransport implements Transport {
 
   private async _fetchRpc(
     rpcRequest: JSONRPCRequest,
-    acceptHeader: string = 'application/json',
+    acceptHeader: string = A2A_CONTENT_TYPE,
     options?: RequestOptions
   ): Promise<Response> {
     const requestInit: RequestInit = {
       method: 'POST',
       headers: {
         ...options?.serviceParameters,
-        'Content-Type': 'application/json',
+        'Content-Type': A2A_CONTENT_TYPE,
         Accept: acceptHeader,
       },
       body: JSON.stringify(rpcRequest),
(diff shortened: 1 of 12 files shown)


src/constants.ts
```diff
@@ -26,6 +26,13 @@ export const A2A_VERSION_HEADER = 'A2A-Version';
  */
 export const A2A_PROTOCOL_VERSION = '1.0';
 
+/**
+ * The A2A JSON content type per §11.1.
+ * REST responses SHOULD use this content type.
+ * Push notification payloads MUST use this content type (§14.1).
+ */
+export const A2A_CONTENT_TYPE = 'application/a2a+json';
```
bartek-gralewicz, line 40: Open question:

I could also export a legacy version here so that there are no hardcoded content types in the code anywhere but we would need to export this on the similar level of visibility as the new one.


```
export const A2A_LEGACY_CONTENT_TYPE = 'application/json';
```

src/client/transports/rest_transport.ts
ishymko, line 1: Worth mentioning that this "SHOULD" clause was added after 1.0 spec release (https://github.com/a2aproject/A2A/pull/1753/), it does not exist [here](https://a2a-protocol.org/v1.0.0/specification/).

I believe it's still fine to use it as it's mentioned in examples even at release commit.

src/server/express/json_rpc_handler.ts
ishymko, line 1: Custom content type only applies to [HTTP+JSON/REST protocol binding](https://a2a-protocol.org/latest/specification/#11-httpjsonrest-protocol-binding), JSON-RPC [still mentions `application/json`](https://a2a-protocol.org/latest/specification/#9-json-rpc-protocol-binding) which makes sense as the payload is different.

src/client/transports/json_rpc_transport.ts
ishymko, line 1: Same as https://github.com/a2aproject/a2a-js/pull/429/changes#r3147901816.

src/server/express/rest_handler.ts
ishymko, line 1: Maybe there is a way to add a middleware once which will set it for each response?

(not that important as we don't have that many "exit" points to update, but just curious)

bartek-gralewicz: Added a middleware logic and removed individual setups.

src/constants.ts
ishymko: `application/json` is not legacy as it's still used in JSON-RPC. We can export `JSON_CONTENT_TYPE` though if there is no such constant in express or standard library.

src/server/express/rest_handler.ts
```diff
@@ -101,7 +101,14 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
 
-  router.use(express.json(), restErrorHandler);
+  router.use(
+    (_req: Request, res: Response, next: NextFunction) => {
+      res.setHeader('Content-Type', A2A_CONTENT_TYPE);
+      next();
+    },
+    express.json({ type: ['application/json', A2A_CONTENT_TYPE] }),
+    restErrorHandler
+  );
```
ishymko, line 116: Do we have any tests we can update to verity this?

ishymko (approved):

src/constants.ts
bartek-gralewicz: Added `JSON_CONTENT_TYPE` constant.

src/server/express/rest_handler.ts
bartek-gralewicz: Added new tests.

PR #430, opened 2026-04-27
feat: Add support for custom authentication scheme and credentials in auth headers

bartek-gralewicz:
# Description

According to the protocol spec, the agent MUST include auth credentials per the push notification's config. This PR adds support for this.

Note: The old `token` behavior is still supported and the old tests pass without any need for updating.

### DIFF
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -105,11 +105,11 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
    */
   private _buildAuthHeaders(pushConfig: TaskPushNotificationConfig): Record<string, string> {
     const headers: Record<string, string> = { 'Content-Type': 'application/json' };
+    const auth = pushConfig.authentication;
 
-    if (pushConfig.authentication?.scheme && pushConfig.authentication?.credentials) {
-      headers['Authorization'] =
-        `${pushConfig.authentication.scheme} ${pushConfig.authentication.credentials}`;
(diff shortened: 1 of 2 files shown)


src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -90,6 +92,30 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
+   * Per §4.3.3, the agent MUST include auth credentials per the push
+   * notification config's `authentication` field when sending notifications.
+   *
+   * Priority:
+   * 1. `pushConfig.authentication` (AuthenticationInfo with scheme + credentials)
+   *    → sets `Authorization: <scheme> <credentials>` per RFC 9110 §11.4
+   * 2. `pushConfig.token` (legacy) → sets the custom token header (deprecated)
+   */
+  private _buildAuthHeaders(pushConfig: TaskPushNotificationConfig): Record<string, string> {
+    const headers: Record<string, string> = { 'Content-Type': 'application/json' };
```
ishymko, line 107: I believe with #429 it should be

```suggestion
    const headers: Record<string, string> = { 'Content-Type': 'application/a2a+json' };
```

ishymko (approved):

src/server/push_notification/default_push_notification_sender.ts
bartek-gralewicz: Yes, I wanted to separate these changes as I was unsure what will be the merging order. It will use the global constant here.

PR #432, opened 2026-04-28
fix: Fix historyLength filtering semantics

bartek-gralewicz:
# Description

Fixed a bug in `historyLength` filtering semantics. The behavior for undefined `historyLength` and equal to 0 were reversed.

## Before
- `historyLength` undefined => **no history**
- `historyLength = 0` => **all history**
- `historyLength = N` => at most N most recent messages returned

## After ([spec](https://a2a-protocol.org/latest/specification/#324-history-length-semantics) aligned)
- `historyLength` undefined => **all history**
- `historyLength = 0` => **no history**
- `historyLength = N` => at most N most recent messages returned


src/server/request_handler/default_request_handler.ts
ishymko, line 499: We have `historyLength` parameter in [`listTasks`](https://a2a-protocol.org/latest/specification/#314-list-tasks) and it's a field of [`SendMessageConfiguration`](https://a2a-protocol.org/latest/specification/#322-sendmessageconfiguration) used in streaming and non-streaming send message, do we have the same handling already applied there? I think we may benefit from a helper shared in both places which references the spec.

bartek-gralewicz: `listTasks` used history filtering in `list` method that is in `store.ts`. Regarding sending messages, there was no logic for that so good catch! I added a helper method and all mentioned methods use it. There are also new tests added for all such methods that focus on properly applied history semantics.

New observation from newly added tests is a bug in [`ResultManager`](https://github.com/a2aproject/a2a-js/blob/955b52b77686485c6e175b6c99fbc1a42b7aa64a/src/server/result_manager.ts#L38). It overwrote the task with a shallow copy what removed the history data. The fix for such is also applied within this PR.

src/server/request_handler/default_request_handler.ts
```diff
@@ -385,12 +387,20 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         );
       }
 
+      if ('id' in finalResult) {
```
ishymko: Maybe we can somehow utilize `AgentExecutionEvent` in such cases? Maybe by updating the type to allow narrowing it down via generics in case we only allow certain event types?

Another option is [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) we already have in some places (i.e. [here](https://github.com/a2aproject/a2a-js/blob/955b52b77686485c6e175b6c99fbc1a42b7aa64a/src/server/utils.ts#L41-L44)). For the core data model we may put in some shared entry point available for both client and server.

src/server/result_manager.ts
```diff
@@ -35,19 +35,36 @@ export class ResultManager {
...
+        const seenIds = new Set(existingHistory.map((m) => m.messageId));
+        const mergedHistory = [
+          ...existingHistory,
+          ...eventHistory.filter((m) => !seenIds.has(m.messageId)),
+        ];
+
+        this.currentTask = {
+          ...taskEvent,
+          history: mergedHistory,
+        };
```
ishymko: I think this should not be required for 1.0 constraints of even sequence listed [here](https://a2a-protocol.org/latest/specification/#312-send-streaming-message). `Task` event should only be seen once for one message processing.

Let's enforce these constraints separately and move on with the current shallow copy for now.

bartek-gralewicz: Restored the `ResultManager` to its original state and fixed the test to be aligned with the flow described in spec.

src/server/request_handler/default_request_handler.ts
bartek-gralewicz: Restored the `ResultManager` logic to the original state and updated the test to properly mimic the flow from the spec.

src/server/request_handler/default_request_handler.ts
bartek-gralewicz: Using the `AgentExecutionEvent` would be difficult here without exposing a richer types as we work on `Task | Message` types here, not event-level types. I applied the type guard option.

ishymko (approved):

PR #437, opened 2026-04-29
feat: Enforce events ordering

bartek-gralewicz:
# Description

This PR introduces a stream ordering enforcement as stated in [protocol spec 3.1.2](https://a2a-protocol.org/latest/specification/#311-send-message) ("Behavior" section).

The stream should distinguish 2 cases:
- The `Message` is the first streamed object.
  - In such case, the stream should not accept any more objects and close right after the initial message. The `UnsupportedOperationError` is thrown if such case occurs.
- The `Task` is the first streamed object.
  - After the initial task, only [TaskStatusUpdateEvent](https://a2a-protocol.org/latest/specification/#421-taskstatusupdateevent) or [TaskArtifactUpdateEvent](https://a2a-protocol.org/latest/specification/#422-taskartifactupdateevent) can occur. In other cases, a `UnsupportedOperationError` is thrown.


src/server/request_handler/default_request_handler.ts
```diff
@@ -464,12 +464,28 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+      eventBus.publish(
+        AgentEvent.task({
+          id: errorTaskStatus.taskId,
+          contextId: errorTaskStatus.contextId,
+          status: errorTaskStatus.status,
+          artifacts: [],
+          history: requestContext.task?.history ? [...requestContext.task.history] : [],
+          metadata: {},
+        })
+      );
```
ishymko: I think we may encounter this condition already after some events published by the agent executor, so it may not be correct to have another task event here.

The way how it's done in Python is "mark task as failed if it was created": 
https://github.com/a2aproject/a2a-python/blob/c0c6c08bcf84a4ef3642dfc0087028812c4ccd51/src/a2a/server/agent_execution/active_task.py#L728-L745

I think it's a reasonable behavior, we should not create a task just to mark it as failed.

src/server/request_handler/default_request_handler.ts
```diff
@@ -836,6 +852,36 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
     }
   }
+
+  /**
+   * Advances the stream pattern state based on the incoming event per §3.1.2.
+   *
+   * Determines whether the event is valid for the current pattern and returns
+   * the (possibly transitioned) pattern. Throws error for invalid transitions.
+   */
+  private _advanceStreamPattern(
```
ishymko, line 870: Nice way to present this logic via "FSM"! :)

src/server/request_handler/default_request_handler.ts
bartek-gralewicz: The task is already marked as failed [here](https://github.com/a2aproject/a2a-js/blob/709d674fde927b81d79599faabbca3c26b180d7d/src/server/request_handler/default_request_handler.ts#L441). The change to add the task before the update was motivated by https://github.com/a2aproject/a2a-js/pull/437#discussion_r3166428765.

If the task is already present, then there is no need for this artificial failed task indeed. I can remove it.

bartek-gralewicz: Changed the behavior to only publish the `TaskStatusUpdateEvent` if the task exists and is in the `TaskStore`. Otherwise the `eventBus` finishes and the method returns.

src/server/request_handler/default_request_handler.ts
```diff
@@ -442,14 +442,21 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
       );
-      // Publish a synthetic error event if needed
+      // Only publish a synthetic error event if the task already exists in the store.
+      // If the agent failed before creating a task, signal the event bus to finish
+      // so the stream consumer doesn't hang indefinitely.
+      const task = await this.taskStore.load(requestContext.taskId, context);
+      if (!task) {
+        eventBus.finished();
+        return;
+      }
```
ishymko, line 457: Can we use `resultManager.getCurrentTask`? I see we use this pattern already in `_handleProcessingError`.

bartek-gralewicz: Yes, that is a better approach. I updated the code to use `resultManager.getCurrentTask()`. Since this is a synchronous call, the `async` in `catch` is no longer required.

ishymko (approved):

PR #444, opened 2026-05-04
ci: Update samples to 1.0

bartek-gralewicz:
# Description

Update sample agent's protocol versions to 1.0.

# Fix

Add `bootstrap-sha` to the release-please config to apply release only on top of currently newest release (0.3.13).

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -2,6 +2,8 @@
   "release-type": "node",
   "prerelease": true,
   "prerelease-type": "alpha",
+  "versioning": "prerelease",
+  "bootstrap-sha": "82d36c7",
   "packages": {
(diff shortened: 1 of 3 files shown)


release-please-config.json
```diff
@@ -2,6 +2,8 @@
   "release-type": "node",
   "prerelease": true,
   "prerelease-type": "alpha",
+  "versioning": "prerelease",
+  "bootstrap-sha": "82d36c7",
```
ishymko: I'd use `last-release-sha` instead, not fully sure what's the difference but logically makes more sense as we had releases in the past already.

ishymko (approved):

PR #446, opened 2026-05-04
ci: Add --tag next for prereleases

bartek-gralewicz:
# Description

Added handling on prereleases. These should include `--tag next` during the `npm publish` call.

The `if` check is based on the release event type as documented here: https://docs.github.com/en/webhooks/webhook-events-and-payloads#release

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Publish to NPM
         run: |
           if [ "${{ github.event.release.prereleased }}" = "true" ]; then
-            npm publish --provenance --access public --tag beta --dry-run
+            npm publish --provenance --access public --tag next --dry-run
           else
             npm publish --provenance --access public --dry-run
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -28,6 +28,13 @@ jobs:
         with:
           node-version: 18
           registry-url: 'https://wombat-dressing-room.appspot.com/'
-      - run: npm publish --provenance --access public
+
+      - name: Publish to NPM
+        run: |
+          if [ "${{ github.event.release.prereleased }}" = "true" ]; then
```
ishymko: I believe it's actually `prerelease` (no `d`), it's an [object in the event's payload](https://docs.github.com/en/webhooks/webhook-events-and-payloads#release) not ["Activity type" `prereleased`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release)

Image

bartek-gralewicz: True, my mistake. Corrected

ishymko (approved):

PR #471, opened 2026-05-13
chore: Restore 0.3 spec types

### LINKED ISSUE #453: chore: Restore 0.3 spec types
During 1.0 alpha creation, old data model (including `types.ts`) were removed and replaced with a new data model. For enabling backward compatibility, the 0.3 compliant type definitions need to be restored.

To not pollute the codebase with the compat layer, all compat-related utilities, including old data model should be isolated in a separate folder (e.g. `src/compat`).

Relevant PR on the Python SDK: https://github.com/a2aproject/a2a-python/pull/754.

JakubWorek:
# Description

Restored v0.3 spec types:
- legacy JSON Schema
- legacy proto types
- legacy `rest_types.ts` file

Fixes #453


src/compat/v0_3/types/pb/a2a.ts
bartek-gralewicz, line 1: There are small differences between this version of this file and the one that was removed (and is still on `main` branch). Formatting is not important and can be updated but some definitions are changed. Like `interface FilePart` has new `name` member it didn't have before. `interface Part` received `metadata` etc.

It would be important to know why these updates occurred and if they could do any harm to the compatibility implementation.

JakubWorek: I've updated the url in scripts to point directly to 0.3.0 tag specification and regenerated the types.

src/compat/v0_3/types/rest_types.ts
```diff
@@ -0,0 +1,120 @@
...
+ * REST PushNotificationConfig (same as internal, no snake_case fields).
+ */
+export interface RestPushNotificationConfig {
+  id: string;
+  url: string;
+  authentication?: {
+    schemes: string[];
+    credentials?: string;
+  };
+}
```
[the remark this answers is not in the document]
JakubWorek: [fix: make id optional and add token field to RestPushNotificatio…](https://github.com/a2aproject/a2a-js/pull/471/commits/4a920f0c99e97d5de8e893750a59bd847591ffa5)

bartek-gralewicz (approved):

PR #478, opened 2026-05-15
feat: defined compat-specific consts and mappings

### LINKED ISSUE #472: feat(compat): Define compat-specific constants and mappings
Create a `constants.ts` file to keep compat-related constants value like:
- `A2A_LEGACY_PROTOCOL_VERSION = '0.3'`
- `LEGACY_HTTP_EXTENSION_HEADER = 'X-A2A-Extensions'`
- `LEGACY_JSON_CONTENT_TYPE = 'application/json'` (REST in v0.3 used plain JSON)

Define method name mappings like `message/send` → `SendMessage` / `SendMessage` → `message/send` (bidirectional).

JakubWorek:
# Description

Defined compat layer specific constants (eg `LEGACY_HTTP_EXTENSION_HEADER`).
Created mappings between v0.3 and v1.0 method names and helper methods for translating them.

Closes #472  🦕

### DIFF
--- a/src/compat/v0_3/constants.ts
+++ b/src/compat/v0_3/constants.ts
@@ -54,9 +54,23 @@ export const LEGACY_JSONRPC_TO_V1: Readonly<Record<string, string>> = Object.fre
 });
 
 /**
+ * v1.0 PascalCase method names that exist in v1.0 but have NO equivalent in
+ * v0.3 (neither JSON-RPC nor gRPC). Translating one of these names into the
+ * v0.3 coordinate system yields a "method not implemented in v0.3" error
+ * (JSON-RPC code -32004 / `A2AError.unsupportedOperation`) instead of the
+ * generic invalid-request error.
+ *
+ * Per §3.5.6 of the v0.3 spec, `tasks/list` was gRPC/REST-only in v0.3 (no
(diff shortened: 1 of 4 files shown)


src/compat/v0_3/constants.ts
```diff
@@ -0,0 +1,133 @@
...
+      grpc,
+      LEGACY_JSONRPC_TO_V1[jsonRpc],
+    ])
+  )
+);
+
+/**
+ * v1.0 PascalCase method name -> v0.3 gRPC PascalCase method name.
+ *
+ * `ListTasks` is omitted: v0.3 has no equivalent gRPC method.
```
bartek-gralewicz: Would be nice to have some sort of explicit check for examples where `ListTasks` is trying to be converted to `v0.3` (like a method-not-implemented-error).

This can be done in a translation layer if that is easier - up to you

JakubWorek: created lookup method `lookupFromV1` that throws `unsupportedOperation` error when method is not implemented in v0.3

bartek-gralewicz (approved):
At some point, we'll probably need some simple REST mapping as well (dropped `/v1` prefix in the endpoint). For now, this looks like a good base.

PR #481, opened 2026-05-18
chore: add translation units

### LINKED ISSUE #455: chore: Exchange helper for AgentCard
Add an exchange logic allowing for passing received 0.3 AgentCard to the 1.0 spec SDK.

### LINKED ISSUE #473: chore(compat): Add translation units
# Description

Add utilities that will perform translation from/to legacy types so that it is possible to easily switch data models. 

The translation should include objects like:
- `Part` - that has 2 representation (new/legacy): `kind:'text'|'file'|'data'` ↔ `content.$case:'text'|'url'|'raw'|'data'`,
- Enums for `TaskState`, `Role`,
- `Message`, `Task`, event kinds,
- Agent card
- Push notifications
- Errors

# Potential approach

Storing translation logic in separate utility files under `src/compat/v0_3/translate/`. Having dedicated utility files like `.../translate/enums.ts` etc.

JakubWorek:
# Description

This PR introduces translatios for v0_3 compatibility layer

The `translate/` subdirectory contains bidirectional payload translators between the modern v1.0 protobuf types (in `src/types/pb/a2a.ts`) and the legacy v0.3 JSON types (in `src/compat/v0_3/types/types.ts`).

The naming convention is direction-anchored:

- `toCore<Entity>` converts a v0.3 value into the equivalent v1.0 proto value.
- `toCompat<Entity>` converts a v1.0 proto value into the equivalent v0.3 JSON value.

Translators are split per entity group (`parts.ts`, `messages.ts`, `tasks.ts`, `push_notifications.ts`, `security.ts`, `agent_card.ts`, `requests.ts`, `enums.ts`, `versions.ts`) for clarity and tree-shaking. The full surface is re-exported from `translate/index.ts`.

Notable policy decisions based on [A2A Python SDK compat layer](https://github.com/a2aproject/a2a-python/tree/main/src/a2a/compat/v0_3):

- `PushNotificationAuthenticationInfo.schemes` is truncated to a single scheme going v0.3 → v1.0; only the first entry is kept.
- The v1.0 `OAuthFlows.deviceCode` flow is silently dropped going v1.0 → v0.3 (v0.3 has no equivalent).
- `TaskStatusUpdateEvent.final` is computed from the status state going v1.0 → v0.3 (`true` for `completed`, `canceled`, `failed`, `rejected`).
- `SendMessageConfiguration.returnImmediately` ↔ `MessageSendConfiguration.blocking` with inverted polarity.
- `toCompatAgentCard` filters `supportedInterfaces` to those whose `protocolVersion` is empty or in `[0.3, 1.0)` and throws `VersionNotSupportedError` if none qualify.

Closes #473 🦕

### DIFF
--- a/src/compat/v0_3/translate/enums.ts
+++ b/src/compat/v0_3/translate/enums.ts
@@ -9,24 +9,23 @@ import type { TaskStatus as LegacyTaskStatus } from '../types/types.js';
  * The v0.3 JSON `TaskStatus.state` string-literal union, kept around as a
  * narrowed alias so call sites don't have to spell it out repeatedly.
  */
-export type LegacyTaskState = LegacyTaskStatus['state'];
+type LegacyTaskState = LegacyTaskStatus['state'];
 
 /** v0.3 JSON `TaskState` literal → v1.0 proto `TaskState` enum. */
-export const TASK_STATE_COMPAT_TO_CORE: Readonly<Record<LegacyTaskState, V1TaskState>> =
-  Object.freeze({
-    unknown: V1TaskState.TASK_STATE_UNSPECIFIED,
-    submitted: V1TaskState.TASK_STATE_SUBMITTED,
-    working: V1TaskState.TASK_STATE_WORKING,
-    completed: V1TaskState.TASK_STATE_COMPLETED,
-    failed: V1TaskState.TASK_STATE_FAILED,
-    canceled: V1TaskState.TASK_STATE_CANCELED,
-    'input-required': V1TaskState.TASK_STATE_INPUT_REQUIRED,
-    rejected: V1TaskState.TASK_STATE_REJECTED,
-    'auth-required': V1TaskState.TASK_STATE_AUTH_REQUIRED,
-  });
+const TASK_STATE_COMPAT_TO_CORE: Readonly<Record<LegacyTaskState, V1TaskState>> = Object.freeze({
+  unknown: V1TaskState.TASK_STATE_UNSPECIFIED,
+  submitted: V1TaskState.TASK_STATE_SUBMITTED,
+  working: V1TaskState.TASK_STATE_WORKING,
+  completed: V1TaskState.TASK_STATE_COMPLETED,
+  failed: V1TaskState.TASK_STATE_FAILED,
+  canceled: V1TaskState.TASK_STATE_CANCELED,
+  'input-required': V1TaskState.TASK_STATE_INPUT_REQUIRED,
+  rejected: V1TaskState.TASK_STATE_REJECTED,
+  'auth-required': V1TaskState.TASK_STATE_AUTH_REQUIRED,
+});
 
 /** v1.0 proto `TaskState` enum → v0.3 JSON `TaskState` literal. */
-export const TASK_STATE_CORE_TO_COMPAT: ReadonlyMap<V1TaskState, LegacyTaskState> = new Map(
+const TASK_STATE_CORE_TO_COMPAT: ReadonlyMap<V1TaskState, LegacyTaskState> = new Map(
   (Object.entries(TASK_STATE_COMPAT_TO_CORE) as [LegacyTaskState, V1TaskState][]).map(
     ([literal, enumValue]) => [enumValue, literal]
   )
--- a/src/compat/v0_3/translate/enums.ts
+++ b/src/compat/v0_3/translate/enums.ts
@@ -1,0 +1,71 @@
+/**
+ * Shared enum translators for v1.0 ↔ v0.3 conversions.
+ */
+
+import { Role as V1Role, TaskState as V1TaskState } from '../../../types/pb/a2a.js';
+import type { TaskStatus as LegacyTaskStatus } from '../types/types.js';
+
+/**
+ * The v0.3 JSON `TaskStatus.state` string-literal union, kept around as a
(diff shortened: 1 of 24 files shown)


src/compat/v0_3/translate/enums.ts
bartek-gralewicz, line 1: I believe that we will not need conversion tables outside this file (only conversion methods). Therefore, no need to export these maps.

src/compat/v0_3/translate/index.ts
bartek-gralewicz, line 1: Based on [this](https://github.com/a2aproject/a2a-js/pull/481/changes#r3266689423) comment, we can either explicitly export relevant methods or use `export * from ...` like below assuming we only export relevant logic from these files.

src/compat/v0_3/translate/messages.ts
bartek-gralewicz, line 1: The name is misleading considering the contents. Consider extracting artifacts to a separate translation file.

src/compat/v0_3/translate/parts.ts
```diff
@@ -0,0 +1,178 @@
...
+      if (metadata !== undefined) result.metadata = metadata;
+      return result;
+    }
+
+    // Non-object values are wrapped so the v0.3 `data: { [k]: unknown }`
+    // type accepts them; mark this in metadata so `toCorePart` can unwrap.
+    const wrappedMetadata: { [k: string]: unknown } = {
+      ...(metadata ?? {}),
+      [DATA_PART_COMPAT_KEY]: true,
+    };
```
bartek-gralewicz: I am hesitant with this approach. If we were to communicate with a v0.3 server, we would send such `wrappedMetadata` to it. It works well only if we do back-and-forth conversions locally without emitting any `Part`s.

Since this wrapping only adds a boolean which we strip in a reverse conversion, what do we gain by this key?

src/compat/v0_3/translate/push_notifications.ts
```diff
@@ -0,0 +1,142 @@
...
+ * **Lossy when `schemes.length > 1`** — only the first scheme is kept.
+ * Callers that need to preserve all declared schemes must do so out-of-band.
+ */
+export function toCoreAuthenticationInfo(
+  compat: legacy.PushNotificationAuthenticationInfo
+): V1AuthenticationInfo {
+  return {
+    scheme: compat.schemes && compat.schemes.length > 0 ? compat.schemes[0]! : '',
+    credentials: compat.credentials ?? '',
+  };
```
bartek-gralewicz, line 48: I would propose that we at least log a warning about a lossy conversion here.

src/compat/v0_3/translate/requests.ts
```diff
@@ -0,0 +1,474 @@
...
+/** v1.0 proto `SendMessageConfiguration` → v0.3 `MessageSendConfiguration`. */
+export function toCompatSendMessageConfiguration(
+  core: V1SendMessageConfiguration
+): legacy.MessageSendConfiguration {
+  const result: legacy.MessageSendConfiguration = {
+    blocking: !core.returnImmediately,
+  };
+  if (core.acceptedOutputModes.length > 0) {
+    result.acceptedOutputModes = [...core.acceptedOutputModes];
+  }
```
bartek-gralewicz: If the number of modes is 0, this field will be omitted, while the reverse method (`toCoreSendMessageConfiguration`) accepts `[]`.

An explicit empty array does not make a round trip through these conversion. Either preserve, or add a comment that such case is dropped.

src/compat/v0_3/translate/requests.ts
bartek-gralewicz, line 1: Multiple occurrences of `as unknown as ...` are present in this file. This is a risk as the compiler will not verify if the types actually match properly.

src/compat/v0_3/translate/requests.ts
```diff
@@ -0,0 +1,474 @@
...
+/* --------------------------------- StreamResponse --------------------------------- */
+
+/**
+ * v0.3 `SendStreamingMessageSuccessResponse` → v1.0 proto
+ * `StreamResponse`.
+ *
+ * The four cases mirror the v1.0 oneof:
+ * `Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent`.
+ */
+export function toCoreStreamResponse(
```
bartek-gralewicz, line 215: In 0.3, the type for response was `SendStreamingMessageSuccessResponse` (actual success response) but in 1.0 it is a union of response and error. 

I would suggest adding the following lines at the top of this method:
```
if('error' in compat)
{
    throw...
}
// The rest of the method as it is now.
```

src/compat/v0_3/translate/parts.ts
JakubWorek: Yeah, I noticed it in Python SDK - this was about types - primitives/arrays/null aren't representable in v0.3's { [k]: unknown } schema. We will drop this solution and throw A2A error when parts are incompatible.

src/compat/v0_3/translate/requests.ts
JakubWorek: v0.3 accepted this as an optional field. I verified that v1.0 treats [] and absent the same. I've changed it so that we always emit acceptedOutputModes. Added explanation as comment in code and roundtrip test for that.

bartek-gralewicz (approved):

PR #495, opened 2026-05-22
feat: implement legacy v0.3 REST support with compatible transport handler and middleware routing

### LINKED ISSUE #482: feat(compat): Implement REST v0.3 transport handler
New `src/compat/v0_3/server/rest/rest_transport_handler.ts` and `src/compat/v0_3/server/express/rest_handler.ts`.

This should use translation mechanisms implemented here: #473.

JakubWorek:
# Description

This pull request introduces support for legacy v0.3 REST API endpoints alongside the existing v1.0 endpoints in the server, ensuring backward compatibility and proper routing for both protocol versions. The main changes involve updating the server's express handler to mount the legacy router, exposing relevant legacy transport handlers, and adding tests to verify correct dispatching, error handling, and header management for both API versions.

**Legacy REST API support and routing:**

* The `restHandler` in `src/server/express/rest_handler.ts` now mounts the `legacyRestRouter` at the top level, ensuring all `/v1/...` paths are handled by the legacy router when appropriate. The v0.3 and v1.0 route sets are disjoint, so middleware and handlers do not interfere. 
* The legacy REST transport handler, error mapper, and related types are now exported from `src/compat/v0_3/server/index.ts` for use elsewhere in the codebase.

**Testing and validation:**

* The test suite in `test/server/express/rest_handler.spec.ts` adds coverage for legacy v0.3 REST dispatch, including:
  - Correct routing of requests to legacy or v1.0 handlers based on path and protocol version.
  - Defaulting to v0.3 when version headers are absent on legacy paths.
  - Proper error mapping and response body shapes for legacy requests.
  - Ensuring legacy endpoints do not register unsupported routes (e.g., `/v1/tasks`).
  - Validation of content-type headers and extension header propagation for both legacy and v1.0 responses.

Closes #482  🦕

### DIFF
--- a/src/compat/v0_3/server/express/rest_handler.ts
+++ b/src/compat/v0_3/server/express/rest_handler.ts
@@ -22,7 +22,7 @@ import express, {
   type Response,
 } from 'express';
 
-import { A2A_VERSION_HEADER } from '../../../../constants.js';
+import { A2A_VERSION_HEADER, HTTP_EXTENSION_HEADER } from '../../../../constants.js';
 import { Extensions } from '../../../../extensions.js';
 import { ServerCallContext } from '../../../../server/context.js';
 import { UserBuilder } from '../../../../server/express/common.js';
@@ -34,6 +34,7 @@ import {
   LEGACY_HTTP_EXTENSION_HEADER,
   LEGACY_JSON_CONTENT_TYPE,
 } from '../../constants.js';
+import { isLegacyVersion } from '../../translate/versions.js';
 import type * as legacy from '../../types/types.js';
 import { A2AError as LegacyA2AError } from '../error.js';
 import {
@@ -81,18 +82,26 @@ type AsyncRouteHandler = (req: Request, res: Response) => Promise<void>;
 /**
  * Creates an Express router exposing the v0.3 HTTP+JSON/REST endpoints.
  *
- * The routes are mount-relative (no `/v1` prefix on the route literals
- * themselves). The caller is expected to mount the router under `/v1`,
- * which is the path prefix used by the v0.3 reference implementation.
- * The core `restHandler` does this automatically; standalone callers
- * must do it explicitly.
+ * The router uses **header-based dispatch**, not path-based: a
+ * middleware at the top of the chain parses `A2A-Version` (defaulting
+ * to `'0.3'` per §3.6.2 when absent) and short-circuits any request
+ * whose version is not in the legacy range `[0.3, 1.0)` by calling
+ * `next('router')`, falling through to the parent's v1.0 middleware.
+ *
+ * The routes are registered at the canonical v0.3 reference URLs
+ * (`/v1/card`, `/v1/message:send`, `/v1/tasks/:taskId`, …). The router
+ * is mounted path-less by the core `restHandler` so the v1.0 spec's
+ * tenant routes (`/:tenant/...`) remain free to use `v1` (or any other
+ * label) as a tenant identifier.
  *
  * The router:
  *   - Parses `application/json` bodies via {@link LEGACY_JSON_CONTENT_TYPE}.
- *   - Reads protocol extensions from the {@link LEGACY_HTTP_EXTENSION_HEADER}
- *     (`X-A2A-Extensions`) header (v0.3 used the `X-` prefix; v1.0 dropped it).
- *   - Defaults a missing `A2A-Version` header to {@link A2A_LEGACY_PROTOCOL_VERSION}
- *     (`'0.3'`).
+ *   - Reads protocol extensions from `X-A2A-Extensions` (the v0.3
+ *     header) OR `A2A-Extensions` (the v1.0 header) for tolerance with
+ *     v1.0-shaped clients hitting the legacy endpoints; if both are
+ *     present the legacy header wins. Responses use the v0.3 spelling.
+ *   - Defaults a missing `A2A-Version` header to
+ *     {@link A2A_LEGACY_PROTOCOL_VERSION} (`'0.3'`).
  *   - Sets the response `Content-Type` to {@link LEGACY_JSON_CONTENT_TYPE}
  *     (`application/json`, not `application/a2a+json`).
  *   - Returns errors in the bare v0.3 `{ code, message, data? }` shape.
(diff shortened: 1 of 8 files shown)


src/compat/v0_3/server/express/rest_handler.ts
```diff
@@ -0,0 +1,432 @@
...
+ * which is the path prefix used by the v0.3 reference implementation.
+ * The core `restHandler` does this automatically; standalone callers
+ * must do it explicitly.
+ *
+ * The router:
+ *   - Parses `application/json` bodies via {@link LEGACY_JSON_CONTENT_TYPE}.
+ *   - Reads protocol extensions from the {@link LEGACY_HTTP_EXTENSION_HEADER}
+ *     (`X-A2A-Extensions`) header (v0.3 used the `X-` prefix; v1.0 dropped it).
+ *   - Defaults a missing `A2A-Version` header to {@link A2A_LEGACY_PROTOCOL_VERSION}
+ *     (`'0.3'`).
```
bartek-gralewicz: What if we convert a request from a v1.0 to v0.3? The extension could be `A2A-Extensions` (without `X-`). Maybe accepting both in the compat layer would be a safer approach?

src/compat/v0_3/server/express/rest_handler.ts
```diff
@@ -0,0 +1,432 @@
+/**
+ * v0.3 HTTP+JSON (REST) Express handler.
+ *
+ * Mounts the v0.3 REST endpoints (under the `/v1/...` path prefix used
+ * by the v0.3 reference implementation) onto an Express router and
+ * delegates to {@link LegacyRestTransportHandler} for the actual
+ * v0.3 ↔ v1.0 translation work.
```
bartek-gralewicz, line 7: This approach is not compatible with the tenants that are a part of `v1.0` spec. Now, setting a tenant `v1` is the same as calling a request in `v0.3` version. Can we resolve over the version instead? 

We could mount the legacy router without a path prefix: `router.use(legacyRestRouter(options))` and have itd `buildContext` call `next()` whenever the resolved version is 1.0.

src/compat/v0_3/server/transports/rest/rest_transport_handler.ts
```diff
@@ -0,0 +1,399 @@
...
+   *   agent does not advertise streaming.
+   */
+  async resubscribe(
+    taskId: string,
+    context: ServerCallContext
+  ): Promise<
+    AsyncGenerator<legacy.SendStreamingMessageSuccessResponse['result'], void, undefined>
+  > {
+    await this.requireCapability('streaming');
+    const stream = this.requestHandler.resubscribe({ id: taskId, tenant: '' }, context);
```
bartek-gralewicz: Hardcoding an empty tenant will break the bidirectional communication between `v0.3` and `v1.0`. If a tenant-scoped request is given, we should return the same.

We could add an optional field with tenant to v0.3 where needed and when we go back to v1.0, we can include it in the response. Otherwise, tenant-scoped requests, return global responses.

Note: Same comments for other hardcoded `tenant: ''` occurrences.

src/server/express/rest_handler.ts
```diff
@@ -108,6 +109,15 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
 
+  // Mount the v0.3 compat router under `/v1`. This is the path-prefix
+  // dispatch point between v0.3 and v1.0: anything under `/v1/*` is
+  // handled by the legacy router with its own middleware (body parser,
+  // content-type setter, error handler); anything else falls through to
+  // the v1.0 middleware and routes below. The two route sets are
+  // disjoint by mount point, so neither version's middleware ever runs
+  // for the other version's requests.
+  router.use('/v1', legacyRestRouter(options));
+
```
bartek-gralewicz: Just like [here](https://github.com/a2aproject/a2a-js/pull/495/changes#r3301713698), this is not compatible with tenant semantics that are available in `v1.0`.

src/server/express/rest_handler.ts
```diff
@@ -108,6 +109,15 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
...
   const restTransportHandler = new RestTransportHandler(options.requestHandler);
 
+  // Mount the v0.3 compat router under `/v1`. This is the path-prefix
+  // dispatch point between v0.3 and v1.0: anything under `/v1/*` is
+  // handled by the legacy router with its own middleware (body parser,
+  // content-type setter, error handler); anything else falls through to
+  // the v1.0 middleware and routes below. The two route sets are
+  // disjoint by mount point, so neither version's middleware ever runs
+  // for the other version's requests.
+  router.use('/v1', legacyRestRouter(options));
```
bartek-gralewicz: Forcing wiring a `v0.3` handler inside a `v1.0` handler requires the compat layer. I believe we should treat the compat layer as something that we could decouple at some point. What do you think about an optional bind like `RestHandlerOptions.legacyCompat?: {enabled: boolean}`? That way, it is possible to use this SDK without the compat layer while sticking to v1.0 only.

With the approach above, one would need to fork and remove this line to remove compat requirement.

src/compat/v0_3/server/transports/rest/rest_transport_handler.ts
```diff
@@ -0,0 +1,399 @@
...
+   * Accepts optional `historyLength` (parsed/validated locally).
+   */
+  async getTask(
+    taskId: string,
+    context: ServerCallContext,
+    historyLength?: unknown
+  ): Promise<legacy.Task> {
+    const params: Parameters<A2ARequestHandler['getTask']>[0] = {
+      id: taskId,
+      historyLength: 0,
```
bartek-gralewicz: If `historyLength` is undefined, using 0 is not a proper equivalent. In `v0.3`, undefined `historyLength` meant "all history" while `historyLength = 0` meant "no history".

I believe that dropping this ` = 0` set here and setting after the `!== undefined` check below should be enough.

test/server/express/rest_handler.spec.ts
```diff
@@ -920,4 +932,266 @@ describe('restHandler', () => {
...
+      (mockRequestHandler.listTasks as Mock).mockResolvedValue({ tasks: [], nextPageToken: '' });
+
+      // Request falls through to the v1.0 router. Whether v1.0 matches it
+      // (under `/:tenant/tasks` with tenant='v1') or returns 404 is an
+      // implementation detail of the v1.0 router; what matters here is
+      // that the legacy router did NOT handle it.
+      await request(dualApp).get('/v1/tasks').set('A2A-Version', '1.0');
+
+      expect(legacyListSpy).not.toHaveBeenCalled();
+    });
```
bartek-gralewicz: This test directly states the issue of the `v1` tenant logic that is not desired. We should resolve over the version and not the just the path.

src/server/express/rest_handler.ts
JakubWorek: Added opt in flag, created #500 issue to mimic behavior in already merged jsonrpc transport handler

src/compat/v0_3/server/express/rest_handler.ts
JakubWorek: applied version checking to use legacy router (if legacyCompad enabled)

src/compat/v0_3/server/transports/rest/rest_transport_handler.ts
JakubWorek: added passing tenant in request to include it in response

bartek-gralewicz (approved):

PR #499, opened 2026-05-25
feat: implement v0.3 JSON-RPC client transport

### LINKED ISSUE #483: feat(compat): Implement JSON-RPC v0.3 transport
New `src/compat/v0_3/client/jsonrpc_transport.ts`

`client.listTasks(...)` agains 0.3 transport should throw an error (`MethodNotFoundError`). 
`ClientFactory` selects this transport when the matched `AgentInterface` advertises `protocolVersion === '0.3'`

JakubWorek:
# Description

This pull request introduces a unified protocol version utility and enhances the JSON-RPC transport layer to transparently support both v1.0 and legacy v0.3 agents. The main changes include updating the transport factory to dynamically select the correct transport implementation based on the agent's protocol version.

**JSON-RPC Transport Factory Enhancements:**

- Updated `JsonRpcTransportFactory` to use a new internal `pickMatchingInterface` function that selects the most appropriate `AgentInterface` for a given endpoint and protocol binding, including disambiguation by URL and preference for v1.0.
- The factory now dynamically loads and returns the legacy v0.3 transport (`LegacyJsonRpcTransport`) if the matched interface's protocol version is in the legacy range, otherwise it defaults to the standard v1.0 transport. This allows seamless support for both protocol versions without pulling in legacy code unless needed.

**Protocol Versioning Utilities:**

- Introduced a new `src/version_utils.ts` module containing helpers for protocol version comparison, such as `isLegacyVersion`, and constants for v0.3 and v1.0 protocol versions. This logic was previously duplicated in the compat layer and is now shared between core and compat code.
- The compat layer's `translate/versions.ts` now re-exports these helpers from the new shared module, ensuring a single source of truth for protocol versioning logic.

Closes #483 🦕

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -422,6 +422,22 @@ export class JsonRpcTransport implements Transport {
 
 export class JsonRpcTransportFactoryOptions {
   fetchImpl?: typeof fetch;
+  /**
+   * Enables the v0.3 protocol compatibility layer.
+   *
+   * When enabled, the factory inspects the matched
+   * `AgentInterface.protocolVersion` on every `create()` call; if it
+   * falls in `[0.3, 1.0)`, the v0.3 `LegacyJsonRpcTransport` is
+   * instantiated instead of the v1.0 `JsonRpcTransport`.
+   *
+   * Default: omitted (treated as disabled). To talk to v0.3 JSON-RPC
+   * agents, the agent card MUST declare a v0.3 `JSONRPC` interface in
+   * `supportedInterfaces`; see §3.6.2.
+   *
+   * When disabled, the v0.3 compat module is never loaded and v0.3
+   * agents are not contacted via the compat transport.
+   */
+  legacyCompat?: { enabled: boolean };
 }
 
 /**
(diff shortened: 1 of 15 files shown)


src/client/transports/json_rpc_transport.ts
```diff
@@ -420,16 +422,91 @@ export class JsonRpcTransport implements Transport {
...
     return PROTOCOL_NAME;
   }
 
-  async create(url: string, _agentCard: AgentCard): Promise<Transport> {
+  async create(url: string, agentCard: AgentCard): Promise<Transport> {
+    if (this.options?.legacyCompat?.enabled) {
+      const iface = pickMatchingInterface(agentCard, PROTOCOL_NAME, url);
+      if (iface && isLegacyVersion(iface.protocolVersion)) {
+        const { LegacyJsonRpcTransport } =
+          await import('../../compat/v0_3/client/transports/jsonrpc_transport.js');
```
bartek-gralewicz: We should keep imports non-conditional. Let's import this every time at the top of the file with the rest of imports.

If at some point, we move towards optional importing, I would propose to dedicate a small wrapper for that instead of spreading hidden imports in transport files.

src/compat/v0_3/client/transports/jsonrpc_transport.ts
```diff
@@ -0,0 +1,486 @@
...
+      case A2A_ERROR_CODE.CONTENT_TYPE_NOT_SUPPORTED:
+        return new ContentTypeNotSupportedError(errorMessage);
+      case A2A_ERROR_CODE.INVALID_AGENT_RESPONSE:
+        return new InvalidAgentResponseError(errorMessage);
+      case A2A_ERROR_CODE.EXTENDED_CARD_NOT_CONFIGURED:
+        return new ExtendedAgentCardNotConfiguredError(errorMessage);
+      default:
+        return new JSONRPCTransportError(response);
+    }
+  }
```
bartek-gralewicz: This is a duplication with `src/client/transports/json_rpc_transport.ts`. Let's move this mapping then to `errors.ts`.

src/compat/v0_3/client/transports/json_rpc_transport.ts
bartek-gralewicz, line 1: To align with `src/client/transports/json_rpc_transport.ts`, lets rename this file to `json_rpc_transport.ts`.

I know that `jsonrpc`/`json_rpc` is used inconsistently across the codebase but let's try to align core with compat at least for now.

Note: Please update the test filename as well.

bartek-gralewicz (approved):

PR #503, opened 2026-05-27
feat: implement v0.3 protocol compatibility for RestTransportFactory

JakubWorek:
# Description

This pull request introduces a compatibility layer for REST transports, allowing the client to transparently support both v1.0 and legacy v0.3 protocol versions based on the agent card's interface. It does this by adding a new dispatch policy that selects the appropriate transport implementation at runtime, and includes comprehensive tests for the new behavior. The main logic for matching the correct agent interface is now shared between protocol-specific factories.

Key changes include:

### Legacy protocol (v0.3) compatibility for REST transport

* Added a `legacyCompat` option to `RestTransportFactoryOptions`, enabling the factory to dynamically choose between `RestTransport` (v1.0) and `LegacyRestTransport` (v0.3) based on the matched interface's `protocolVersion`. When enabled, v0.3 agents are contacted using the compat transport; otherwise, only v1.0 is used.
* Updated the `RestTransportFactory.create()` method to use the new dispatch policy, loading the v0.3 transport module only when necessary.

### Interface selection logic refactor

* Moved the `pickMatchingInterface` helper into its own module (`pick_interface.ts`) so both JSON-RPC and REST factories share the same logic for matching agent interfaces by protocol binding, URL, and version.

Closes #484  🦕

PR #504, opened 2026-05-29
feat: add v0.3 compatibility layer for client and server agent-card handling

### LINKED ISSUE #485: feat(compat): Resolve AgentCard based on the provided protocol version
# New Files

- New `AgentCardResolver` logic defined in `src/compat/v0_3/client/`.
- New `AgentCardHandler` logic defined in `src/compat/v0_3/server/`.

# Client Side

"Compat-aware" `AgentCardResolver` that detects v0.3 cards, translates them to v1.0 shape, marks each synthesized interface with `protocolVersion: 0.3`so that proper compat transports can be selected by the factory.

# Server side

Extend the `agentCardHandler` with an opt-in compat negotiation that on 0.3 version, serves a compat card produced by `toCompatAgentCard(card)` helper method.

JakubWorek:
# Description

This pull request introduces a v0.3 protocol compatibility layer for both client and server handling of agent cards. The main goal is to allow negotiation and translation between legacy (v0.3) and modern (v1.0) agent card shapes, ensuring backward compatibility for clients and servers opting into this feature. The changes are opt-in and only affect behavior when the new `legacyCompat: { enabled: true }` option is set.

Closes #485  🦕

### DIFF
--- a/src/client/card-resolver.ts
+++ b/src/client/card-resolver.ts
@@ -1,9 +1,28 @@
 import { AGENT_CARD_PATH } from '../constants.js';
 import { AgentCard } from '../index.js';
+import { isLegacyAgentCard, parseLegacyAgentCard } from '../compat/v0_3/client/index.js';
 
 export interface AgentCardResolverOptions {
   path?: string;
   fetchImpl?: typeof fetch;
+  /**
+   * Enables the v0.3 protocol compatibility layer.
+   *
+   * When enabled, the resolver inspects each fetched agent-card
+   * payload; if its shape matches v0.3 (top-level `url` without
+   * `supportedInterfaces`, `preferredTransport`,
+   * `additionalInterfaces`, `supportsAuthenticatedExtendedCard`, or
+   * a `protocolVersion` in `[0.3, 1.0)`), it is translated to the
+   * v1.0 proto shape via `toCoreAgentCard`. Each synthesized
+   * `AgentInterface` is stamped with `protocolVersion: '0.3'` so
+   * that a {@link JsonRpcTransportFactory} configured with
+   * `legacyCompat: { enabled: true }` selects the compat transport
+   * automatically.
+   *
+   * Default: omitted (treated as disabled). When disabled, the v0.3
+   * compat module is never loaded.
+   */
+  legacyCompat?: { enabled: boolean };
 }
 
 export interface AgentCardResolver {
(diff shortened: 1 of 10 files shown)


src/client/card-resolver.ts
```diff
@@ -1,9 +1,28 @@
 import { AGENT_CARD_PATH } from '../constants.js';
 import { AgentCard } from '../index.js';
+import { isLegacyAgentCard, parseLegacyAgentCard } from '../compat/v0_3/client/index.js';
```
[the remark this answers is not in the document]
JakubWorek: as in other PRs, we've decided that for now we'd import statically in this cases (as mentioned [here](https://github.com/a2aproject/a2a-js/pull/499#discussion_r3309763191))

bartek-gralewicz: Let's import these helper methods directly from their source locations not to include the whole compat tree.

src/compat/v0_3/client/card_resolver.ts
```diff
@@ -0,0 +1,89 @@
...
+ * (`../translate/agent_card.ts`) and stamps `protocolVersion: '0.3'` on
+ * every synthesized interface — both the primary (line 200) and every
+ * entry of `additionalInterfaces` (via {@link toCoreAgentInterface}).
+ * A {@link JsonRpcTransportFactory} constructed with
+ * `legacyCompat: { enabled: true }` will then pick the compat transport
+ * automatically when the matched interface declares a legacy version.
+ *
+ * Kept in the compat layer (rather than in the core
+ * `src/client/card-resolver.ts`) so the v0.3 types and translators are
+ * only loaded when an operator has explicitly opted into compat.
```
bartek-gralewicz:
```suggestion
 * only called when an operator has explicitly opted into compat.
```

Similar as in the #503. I believe that due to such comments, there are multiple AI review suggestions about lazy imports.

src/compat/v0_3/client/card-resolver.ts
bartek-gralewicz, line 1: For unification with core card resolver, let's rename this file to `card-resolver`. I know that file naming is not unified but let's at least try to match naming between core-compat.

src/compat/v0_3/server/express/agent_card_handler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+      res.setHeader('Content-Type', LEGACY_JSON_CONTENT_TYPE);
+      res.status(HTTP_STATUS.OK).send(body);
+    } catch (error) {
+      console.error('Error fetching legacy agent card:', error);
+      res
+        .status(500)
+        .setHeader('Content-Type', LEGACY_JSON_CONTENT_TYPE)
+        .json({ error: 'Failed to retrieve agent card' });
+    }
```
bartek-gralewicz, line 156: Shouldn't we do the same addition as [here](https://github.com/a2aproject/a2a-js/pull/504/changes#diff-b157f9ced7ad3b8d37ecaad17880a9d6a985ebb06a897100dad8d1bb695818e1R133)?

bartek-gralewicz (approved):

PR #507, opened 2026-06-05
feat: implement versioned push notification serializers with legacy v0.3 compatibility support

### LINKED ISSUE #487: feat(compat): Push notification configs storing
Store push notification configs with the originating wire version so the sender can serialize back to v0.3 wire format.

JakubWorek:
# Description

This pull request introduces a version-aware push notification system, enabling the server to deliver webhooks in the correct wire format based on the protocol version used at registration. The changes add a serializer abstraction, track the wire version per webhook registration, and provide a factory for v0.3 compatibility. Below are the most important changes:

**Push Notification Versioning and Serialization:**

* Introduced the `PushNotificationSerializer` interface and `SerializedPushNotification` type, allowing the server to serialize push notification payloads differently based on the wire protocol version. Added the canonical `V1PushNotificationSerializer` for v1.0 and a `V03PushNotificationSerializer` for v0.3 compatibility. 
* Updated `DefaultPushNotificationSender` to support per-version serializers via a new `serializers` option, resolve the correct serializer for each outgoing notification, and fall back to v1.0 with a warning if an unknown version is encountered. 

**Wire Version Tracking in Push Notification Store:**

* Modified the `PushNotificationStore` interface and its in-memory implementation to persist the wire protocol version (`wireVersion`) alongside each config at registration, and return it on load. Introduced the `StoredPushNotificationConfig` type for this purpose.

**v0.3 Compatibility Layer:**

* Added a `createLegacyAwarePushNotificationSender` factory and re-exported it (along with the v0.3 serializer) from the compat layer, making it easy to set up a sender that delivers v0.3-shaped webhooks for legacy transports.

These changes ensure that webhooks registered over legacy transports receive notifications in the expected v0.3 format, while modern clients continue to receive the v1.0 payloads.

Closes #487 🦕

### DIFF
--- a/src/compat/v0_3/server/push_notification/v03_push_notification_serializer.ts
+++ b/src/compat/v0_3/server/push_notification/v03_push_notification_serializer.ts
@@ -1,0 +1,71 @@
+import {
+  PushNotificationSerializer,
+  SerializedPushNotification,
+} from '../../../../server/push_notification/push_notification_serializer.js';
+import { StreamResponse } from '../../../../index.js';
+import { LEGACY_JSON_CONTENT_TYPE } from '../../constants.js';
+import { toCompatTask } from '../../translate/tasks.js';
+import {
+  toCompatTaskArtifactUpdateEvent,
+  toCompatTaskStatusUpdateEvent,
+} from '../../translate/tasks.js';
+
+/**
+ * The v0.3 push notification serializer.
+ *
+ * Per the v0.3 spec example (§9.5), the push-notification HTTP body is the
+ * **bare event object** (a v0.3 JSON `Task`, `TaskStatusUpdateEvent`, or
+ * `TaskArtifactUpdateEvent` discriminated by its `kind` field) with
+ * `Content-Type: application/json`. Notably:
+ *
+ *  - It is **not** wrapped in a `StreamResponse` discriminator (no outer
+ *    `task` / `statusUpdate` / `artifactUpdate` key) — that wrapper is a
+ *    v1.0 addition (§4.3.3).
+ *  - It is **not** wrapped in a JSON-RPC envelope (no `jsonrpc`, `id`,
+ *    `result`) — push notifications are unsolicited and have no in-flight
+ *    request to correlate against; the JSON-RPC envelope only appears on
+ *    the streaming (SSE) path in v0.3.
+ *
+ * The canonical {@link StreamResponse} payload is translated to the v0.3
+ * JSON shape via the per-case `toCompat*` translators in
+ * `compat/v0_3/translate/tasks.ts`, which also set the legacy `kind`
+ * discriminator (`'task'`, `'status-update'`, or `'artifact-update'`) the
+ * v0.3 schema requires.
+ *
+ * `message` payloads are rejected (consistent with v1.0 behavior); push
+ * notifications are defined for task / status / artifact events only.
+ */
+export class V03PushNotificationSerializer implements PushNotificationSerializer {
+  serialize(streamResponse: StreamResponse): SerializedPushNotification {
+    const payload = streamResponse.payload;
+    if (!payload) {
+      throw new Error('StreamResponse payload is undefined');
+    }
+
+    let legacyEvent: unknown;
+    switch (payload.$case) {
+      case 'task':
+        legacyEvent = toCompatTask(payload.value);
+        break;
+      case 'statusUpdate':
+        legacyEvent = toCompatTaskStatusUpdateEvent(payload.value);
+        break;
+      case 'artifactUpdate':
+        legacyEvent = toCompatTaskArtifactUpdateEvent(payload.value);
+        break;
+      case 'message':
+        throw new Error('Push notification should not be sent for message payload.');
+      default: {
+        // Exhaustive check: keeps this switch in sync with the StreamResponse
+        // payload union at compile time.
+        const _exhaustive: never = payload;
+        throw new Error(`Unknown payload case: ${(_exhaustive as { $case: string }).$case}`);
+      }
+    }
+
+    return {
+      body: JSON.stringify(legacyEvent),
+      contentType: LEGACY_JSON_CONTENT_TYPE,
+    };
+  }
+}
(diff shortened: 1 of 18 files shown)


src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -29,16 +55,33 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
...
+    // this from the registry (not the local `builtinV1`) so a user who
+    // overrode '1.0' has their custom serializer used for fallback too.
+    this.fallbackSerializer = this.serializers.get(A2A_PROTOCOL_VERSION) ?? builtinV1;
   }
 
   async send(streamResponse: StreamResponse, context: ServerCallContext): Promise<void> {
     const taskId = this._getTaskId(streamResponse);
-    const pushConfigs = await this.pushNotificationStore.load(taskId, context);
-    if (!pushConfigs || pushConfigs.length === 0) {
+    const storedConfigs = await this.pushNotificationStore.load(taskId, context);
```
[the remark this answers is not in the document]
JakubWorek: Added streamResponse.payload?.$case !== 'message' to the _sendPushNotificationIfNeeded guard

src/server/push_notification/push_notification_store.ts
```diff
@@ -43,27 +73,36 @@ export class InMemoryPushNotificationStore implements PushNotificationStore {
...
     }
 
-    // Remove existing config with the same ID if it exists
-    const existingIndex = configs.findIndex((config) => config.id === pushNotificationConfig.id);
+    // Capture the wire version from the request context. ServerCallContext
+    // always populates this field (defaulting to A2A_LEGACY_PROTOCOL_VERSION
+    // when the A2A-Version header is absent, per §3.6.2), so the fallback to
+    // A2A_PROTOCOL_VERSION below is defensive only and applies if a caller
+    // somehow constructs an entry without going through the normal context.
+    const wireVersion = context.requestedVersion || A2A_PROTOCOL_VERSION;
```
bartek-gralewicz: In case the caller constructs the entry in a non-standard approach, isn't it safer to default to 0.3 just like the spec suggests? In general, if the version is absent in places where it is expected to be present, the recommended approach is to default to 0.3.

Same is instructed in the spec section that you refer to in this comment: [§3.6.2](https://a2a-protocol.org/latest/specification/#362-server-responsibilities)

src/server/request_handler/default_request_handler.ts
```diff
@@ -801,12 +802,24 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
   private async _sendPushNotificationIfNeeded(
     context: ServerCallContext,
     streamResponse: StreamResponse
   ): Promise<void> {
-    if (this.agentCard.capabilities?.pushNotifications && this.pushNotificationSender) {
+    if (
+      this.agentCard.capabilities?.pushNotifications &&
+      this.pushNotificationSender &&
+      streamResponse.payload?.$case !== 'message'
+    ) {
```
bartek-gralewicz: Is this really the case with messages? Looking at the [§4.3.3](https://a2a-protocol.org/latest/specification/#433-push-notification-payload), messages are one of the supported payload types.

src/server/push_notification/default_push_notification_sender.ts
```diff
@@ -15,12 +19,34 @@ export interface DefaultPushNotificationSenderOptions {
...
+   * ({@link V1PushNotificationSerializer}) at construction time; entries
+   * supplied here override that default and add support for additional
+   * versions (e.g. legacy v0.3 via the compat layer's
+   * `V03PushNotificationSerializer`).
+   *
+   * When a stored config carries a wire version with no registered
+   * serializer, the sender logs a warning and falls back to the `'1.0'`
+   * serializer for that dispatch.
+   */
+  serializers?: Record<string, PushNotificationSerializer>;
```
bartek-gralewicz: Perhaps a better solution would be to create (or reuse existing one if present) an enum with spec versions and use them instead of string based dict?

src/compat/v0_3/server/push_notification/v03_push_notification_serializer.ts
```diff
@@ -0,0 +1,71 @@
...
+  PushNotificationSerializer,
+  SerializedPushNotification,
+} from '../../../../server/push_notification/push_notification_serializer.js';
+import { StreamResponse } from '../../../../index.js';
+import { LEGACY_JSON_CONTENT_TYPE } from '../../constants.js';
+import { toCompatTask } from '../../translate/tasks.js';
+import {
+  toCompatTaskArtifactUpdateEvent,
+  toCompatTaskStatusUpdateEvent,
+} from '../../translate/tasks.js';
```
bartek-gralewicz: These import from the same file and could be gathered together.

src/server/push_notification/push_notification_store.ts
```diff
@@ -43,27 +73,36 @@ export class InMemoryPushNotificationStore implements PushNotificationStore {
...
+    bucket.set(taskId, entries);
   }
 
-  async load(taskId: string, context: ServerCallContext): Promise<TaskPushNotificationConfig[]> {
-    const configs = this._scopedStore.getBucket(context)?.get(taskId);
-    return configs || [];
+  async load(taskId: string, context: ServerCallContext): Promise<StoredPushNotificationConfig[]> {
+    const entries = this._scopedStore.getBucket(context)?.get(taskId);
+    return entries || [];
   }
```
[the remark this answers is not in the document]
bartek-gralewicz: This shallow copy solves the issue of sharing objects on the top level but items within the arrays are still not deep-copied. If an individual config field is mutated, the change is global for all references of this `entries`.

src/server/push_notification/push_notification_store.ts
```diff
@@ -43,27 +73,36 @@ export class InMemoryPushNotificationStore implements PushNotificationStore {
...
-    bucket.set(taskId, configs);
+    // Add the new/updated entry
+    entries.push({ config: pushNotificationConfig, wireVersion });
+    bucket.set(taskId, entries);
   }
 
-  async load(taskId: string, context: ServerCallContext): Promise<TaskPushNotificationConfig[]> {
-    const configs = this._scopedStore.getBucket(context)?.get(taskId);
-    return configs || [];
+  async load(taskId: string, context: ServerCallContext): Promise<StoredPushNotificationConfig[]> {
```
bartek-gralewicz: The change of the output type is the breaking change as this is a user-facing functionality. If we can avoid this, it would be the best option. If not, we need to explicitly mark this as a breaking change in the PR description and in the upcoming release.

JakubWorek: added it as optional method in the interface, output type of load() stays the same

src/server/push_notification/push_notification_store.ts
JakubWorek: fixed

bartek-gralewicz (approved):

PR #513, opened 2026-06-12
feat: implement v0.3 ITK cross-SDK validation workflows

### LINKED ISSUE #508: feat(compat): Expand ITK suite to test 0.3 compatibility

JakubWorek:
# Description

This pull request introduces significant improvements to the ITK agent and its test harness, focusing on full compatibility with both v1.0 and v0.3 agent protocols, and aligning the agent's task lifecycle and resubscribe/cancel behavior with the v0.3 reference implementations. Additionally, it introduces a new nightly workflow for ITK, and updates the main ITK test workflow for clarity and correctness.

Fixes #508  🦕

### DIFF
--- a/.github/workflows/run-itk-nightly.yaml
+++ b/.github/workflows/run-itk-nightly.yaml
@@ -6,14 +6,33 @@ on:
     # so the cross-SDK nightly artifacts upload roughly in parallel.
     - cron: '0 2 * * *'
   workflow_dispatch:  # Allow manual execution from the Actions tab
+  # Also fire on every push to `main` and `epic`. The `schedule:` and
+  # `workflow_dispatch:` triggers above only become live once this file
+  # lands on the repo's default branch (`main`) — until then the cron
+  # doesn't tick and the "Run workflow" button doesn't appear in the
+  # Actions UI. The `push:` trigger lets us validate the full nightly
+  # matrix (37 scenarios) and the rolling-metrics upload on the `epic`
+  # branch while it's still the active development branch, matching the
+  # PR workflow's `branches: ["main", "epic"]` configuration. Drop the
+  # `epic` entry once this is merged to `main` and the cron takes over.
+  push:
+    branches: ["main", "epic"]
+    paths:
+      - 'src/**'
+      - 'itk/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/run-itk-nightly.yaml'
 
 permissions:
   # Required for `softprops/action-gh-release` to publish the rolling
   # `nightly-metrics` prerelease asset.
   contents: write
 
-# Serialize nightly runs so a manual `workflow_dispatch` while the
-# scheduled run is still uploading doesn't race on the release asset.
+# Serialize nightly runs so a manual `workflow_dispatch`, a `push` to
+# main/epic, and the scheduled cron don't race on the release asset
+# upload (they all target the same `nightly-metrics` tag's
+# `itk_js.json` file).
 concurrency:
   group: itk-nightly
   cancel-in-progress: true
--- a/.github/workflows/run-itk-nightly.yaml
+++ b/.github/workflows/run-itk-nightly.yaml
@@ -1,0 +1,73 @@
+name: Run ITK (Nightly)
+
+on:
+  schedule:
+    # 02:00 UTC daily, matching a2a-python's and a2a-go's nightly windows
+    # so the cross-SDK nightly artifacts upload roughly in parallel.
+    - cron: '0 2 * * *'
+  workflow_dispatch:  # Allow manual execution from the Actions tab
+
+permissions:
+  # Required for `softprops/action-gh-release` to publish the rolling
+  # `nightly-metrics` prerelease asset.
+  contents: write
+
+# Serialize nightly runs so a manual `workflow_dispatch` while the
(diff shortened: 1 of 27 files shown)


src/server/express/rest_handler.ts
```diff
@@ -145,7 +145,7 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
       res.setHeader('Content-Type', A2A_CONTENT_TYPE);
       next();
     },
-    express.json({ type: [JSON_CONTENT_TYPE, A2A_CONTENT_TYPE] }),
+    express.json({ type: [JSON_CONTENT_TYPE, A2A_CONTENT_TYPE], strict: false }),
```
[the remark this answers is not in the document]
JakubWorek: Added guards

.github/workflows/run-itk-nightly.yaml
```diff
@@ -0,0 +1,92 @@
...
+  # Required for `softprops/action-gh-release` to publish the rolling
+  # `nightly-metrics` prerelease asset.
+  contents: write
+
+# Serialize nightly runs so a manual `workflow_dispatch`, a `push` to
+# main/epic, and the scheduled cron don't race on the release asset
+# upload (they all target the same `nightly-metrics` tag's
+# `itk_js.json` file).
+concurrency:
+  group: itk-nightly
```
bartek-gralewicz: I recommend adding a git ref to the concurrency group rule (just like in standard ITK workflow). Without it, a manual execution on any branch will cancel other jobs (even the cron).

src/compat/v0_3/client/transports/rest_transport.ts
```diff
@@ -245,27 +287,43 @@ export class LegacyRestTransport implements Transport {
...
     const path = `/v1/tasks/${encodeURIComponent(params.id)}:subscribe`;
-    yield* this._sendStreamingRequest(path, undefined, options);
+    // GET, not POST: the v0.3 proto's `google.api.http` annotation for
+    // `SubscribeToTask` maps to `get: "/v1/tasks/{id}:subscribe"` (no
+    // request body), and a2a-python's `A2ARESTFastAPIApplication` only
+    // registers the GET method. Some baselines (a2a-python's v0.3 compat
+    // adapter at `src/a2a/compat/v0_3/rest_adapter.py`) additionally
+    // accept POST for tolerance, but the canonical reference does not —
+    // sending POST yields 405 Method Not Allowed.
+    yield* this._sendStreamingRequest(path, 'GET', undefined, options);
```
bartek-gralewicz, line 297: According to both [spec for 1.0](https://a2a-protocol.org/v1.0.0/specification/#1132-task-operations) and [spec for 0.3](https://a2a-protocol.org/v0.3.0/specification/#356-method-mapping-reference-table), subscribe is the POST method, not GET.

test/compat/v0_3/client/transports/rest_transport.spec.ts
```diff
@@ -106,10 +106,11 @@ describe('LegacyRestTransport', () => {
...
         makeJsonResponse({
-          kind: 'task',
-          id: 't-1',
-          contextId: 'ctx-1',
-          status: { state: 'working' },
+          task: {
+            id: 't-1',
+            contextId: 'ctx-1',
+            status: { state: 'TASK_STATE_WORKING' },
+          },
```
bartek-gralewicz, line 113: The v0.3 data model uses `kind` discriminators. The change in this test most likely comes from proto and json conversions in `rest_handler.ts`.

Do you have any source of information that proves this change?

src/compat/v0_3/client/transports/rest_transport.ts
JakubWorek: Spec says POST, proto says GET:
https://github.com/a2aproject/A2A/blob/8d57eba286de756176892518a8fc39b0ac2ccefb/specification/grpc/a2a.proto#L68-L73
https://github.com/a2aproject/A2A/blob/173695755607e884aa9acf8ce4feed90e32727a1/specification/a2a.proto#L76-L83

Empirically: an earlier ITK iteration sent POST and python_v03 returned 405 Method Not Allowed; switching the client to GET fixed it. So I think we should follow proto convention. What do you think?

Our server side is tolerant — it registers both GET and POST for :subscribe, so v1.0 clients that send POST still work against our server. The client must use GET to interop with reference servers.

test/compat/v0_3/client/transports/rest_transport.spec.ts
JakubWorek: The change is intentional — it aligns the test with the code in src/compat/v0_3/server/express/rest_handler.ts, which emits proto-JSON of the v0.3 proto types for the REST surface (not the JSON-RPC body shape).

To the sources:
Empirical: ITK runs against python_v03 pass with this shape and fail with the JSON-RPC kind-discriminator shape

a2a-python v0.3.24 (the python_v03 ITK baseline):
https://github.com/a2aproject/a2a-python/blob/d2d38608b0262aeb67d94d11136a4147e65a7351/src/a2a/server/request_handlers/rest_handler.py#L86-L88

.github/workflows/run-itk-nightly.yaml
JakubWorek: fixed

bartek-gralewicz (approved):

PR #517, opened 2026-06-15
feat: implement TCK workflow and SUT for tests, adjust ci

### LINKED ISSUE #491: chore(compat): Enable TCK
Enable TCK test kit.

Existing branch: [bgralewicz/enable_tck_tests](https://github.com/a2aproject/a2a-js/tree/bgralewicz/enable_tck_tests) (needs updating to the epic/v1.0 state).

JakubWorek:
# Description

Replaces the experimental TCK SUT/workflow from bgralewicz/enable_tck_tests with a working setup across all three transports (JSON-RPC, REST/HTTP+JSON, gRPC) on a2a-tck 1.0.0.alpha2, and lands the SDK fixes that the conformance suite exposed.

TCK harness and its consequences:
- tck/agent/index.ts — rewritten as a messageId-prefix-routed SUT executor matching the canonical Gherkin scenarios (scenarios/core_operations.feature, scenarios/streaming.feature) and the reference Python SUT (a2a-tck/sut/a2a-python/sut_agent.py). Serves the agent card at /.well-known/agent-card.json, JSON-RPC at /, REST at /a2a/rest, and gRPC on a separate port — all three declared in supportedInterfaces so the conformance suite probes all bindings against the same agent.
- .github/workflows/run-tck.yaml — pinned to TCK release 1.0.0.alpha2; runs both MUST and SHOULD; uploads compatibility reports as artifacts; triggers on main and epic/**.
- SDK fixes driven by TCK failures

Closes #491  🦕

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -151,13 +151,17 @@ export function buildErrorInfo(
 export const A2A_ERROR_GRPC_STATUS: Record<string, string> = {
   TaskNotFoundError: 'NOT_FOUND',
   TaskNotCancelableError: 'FAILED_PRECONDITION',
-  PushNotificationNotSupportedError: 'FAILED_PRECONDITION',
-  UnsupportedOperationError: 'FAILED_PRECONDITION',
+  // Capability-gated rejections (PushNotifications, UnsupportedOperation,
+  // VersionNotSupported) surface as UNIMPLEMENTED rather than
+  // FAILED_PRECONDITION so a client can distinguish "agent does not
+  // implement this" from "current state forbids this operation".
+  PushNotificationNotSupportedError: 'UNIMPLEMENTED',
+  UnsupportedOperationError: 'UNIMPLEMENTED',
   ContentTypeNotSupportedError: 'INVALID_ARGUMENT',
   InvalidAgentResponseError: 'INTERNAL',
   ExtendedAgentCardNotConfiguredError: 'FAILED_PRECONDITION',
   ExtensionSupportRequiredError: 'FAILED_PRECONDITION',
-  VersionNotSupportedError: 'FAILED_PRECONDITION',
+  VersionNotSupportedError: 'UNIMPLEMENTED',
   RequestMalformedError: 'INVALID_ARGUMENT',
   GenericError: 'INTERNAL',
 };
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -151,10 +151,6 @@ export function buildErrorInfo(
 export const A2A_ERROR_GRPC_STATUS: Record<string, string> = {
   TaskNotFoundError: 'NOT_FOUND',
   TaskNotCancelableError: 'FAILED_PRECONDITION',
-  // Capability-gated rejections (PushNotifications, UnsupportedOperation,
-  // VersionNotSupported) surface as UNIMPLEMENTED rather than
-  // FAILED_PRECONDITION so a client can distinguish "agent does not
-  // implement this" from "current state forbids this operation".
   PushNotificationNotSupportedError: 'UNIMPLEMENTED',
   UnsupportedOperationError: 'UNIMPLEMENTED',
   ContentTypeNotSupportedError: 'INVALID_ARGUMENT',
--- a/src/server/express/rest_handler.ts
+++ b/src/server/express/rest_handler.ts
@@ -37,7 +37,7 @@ import {
   TaskPushNotificationConfig,
 } from '../../types/pb/a2a.js';
 import { ToProto } from '../../types/converters/to_proto.js';
-import { RequestMalformedError } from '../../errors.js';
+import { ContentTypeNotSupportedError, RequestMalformedError } from '../../errors.js';
 
 /**
  * Options for configuring the HTTP+JSON/REST handler.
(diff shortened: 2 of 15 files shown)


src/server/request_handler/default_request_handler.ts
```diff
@@ -1008,3 +1047,28 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+  let lastState: TaskState | undefined;
+  bus.on('event', (event) => {
+    if (event.kind === 'task' && event.data.status?.state !== undefined) {
+      lastState = event.data.status.state;
+    } else if (event.kind === 'statusUpdate' && event.data.status?.state !== undefined) {
+      lastState = event.data.status.state;
+    }
+  });
+  return () => lastState;
+}
```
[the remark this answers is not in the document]
JakubWorek: applied

src/server/request_handler/default_request_handler.ts
```diff
@@ -381,11 +389,40 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         );
       })
       .finally(() => {
-        eventBus.finished();
-        this.eventBusManager.cleanupByTaskId(taskId);
+        this._tearDownOrKeepBus(taskId, eventBus, stateTracker());
```
bartek-gralewicz: Since this section is in `.finally()`, do we really want to keep the event bus alive, even in case it was during some interrupted state?

src/server/request_handler/default_request_handler.ts
```diff
@@ -451,8 +491,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         eventBus.publish(AgentEvent.statusUpdate(errorTaskStatus));
       })
       .finally(() => {
-        eventBus.finished();
-        this.eventBusManager.cleanupByTaskId(taskId);
+        this._tearDownOrKeepBus(taskId, eventBus, stateTracker());
```
bartek-gralewicz: Same as [here](https://github.com/a2aproject/a2a-js/pull/517/changes#r3419714385).

src/errors.ts
```diff
@@ -151,13 +151,17 @@ export function buildErrorInfo(
...
   TaskNotFoundError: 'NOT_FOUND',
   TaskNotCancelableError: 'FAILED_PRECONDITION',
-  PushNotificationNotSupportedError: 'FAILED_PRECONDITION',
-  UnsupportedOperationError: 'FAILED_PRECONDITION',
+  // Capability-gated rejections (PushNotifications, UnsupportedOperation,
+  // VersionNotSupported) surface as UNIMPLEMENTED rather than
+  // FAILED_PRECONDITION so a client can distinguish "agent does not
+  // implement this" from "current state forbids this operation".
+  PushNotificationNotSupportedError: 'UNIMPLEMENTED',
+  UnsupportedOperationError: 'UNIMPLEMENTED',
```
bartek-gralewicz: nit: I would resist from keeping such comment blocks permanently. These are helpful when the code change is presented but have much smaller value for later if kept in the code. Also, these reduce the overall readability, especially in places like maps, enums etc like here.

src/server/express/rest_handler.ts
```diff
@@ -142,9 +169,18 @@ export function restHandler(options: RestHandlerOptions): RequestHandler {
 
   router.use(
     (_req: Request, res: Response, next: NextFunction) => {
-      res.setHeader('Content-Type', A2A_CONTENT_TYPE);
+      // §11.1 SHOULD use `application/a2a+json`, but we emit the
+      // broader `application/json` so every standard HTTP+JSON client
+      // (including those without A2A-specific media-type handling) can
+      // interoperate. Both content types are still accepted on input.
+      res.setHeader('Content-Type', JSON_CONTENT_TYPE);
```
bartek-gralewicz, line 176: I would try our best to be aligned with the spec, therefore, we SHOULD use `application/a2a+json` for requests and responses ([spec 11.1](https://a2a-protocol.org/latest/specification/#111-protocol-requirements)). Does this not work with TCK?

src/server/transports/rest/rest_transport_handler.ts
```diff
@@ -71,7 +72,7 @@ export function mapErrorToStatus(error: unknown): number {
   if (error instanceof TaskNotCancelableError) return HTTP_STATUS.BAD_REQUEST;
   if (error instanceof PushNotificationNotSupportedError) return HTTP_STATUS.BAD_REQUEST;
   if (error instanceof UnsupportedOperationError) return HTTP_STATUS.BAD_REQUEST;
-  if (error instanceof ContentTypeNotSupportedError) return HTTP_STATUS.BAD_REQUEST;
+  if (error instanceof ContentTypeNotSupportedError) return HTTP_STATUS.UNSUPPORTED_MEDIA_TYPE;
```
bartek-gralewicz: According to [spec 5.4](https://a2a-protocol.org/latest/specification/#54-error-code-mappings), `ContentTypeNotSupportedError` should map to 400 (Bad Request).

src/errors.ts
bartek-gralewicz, line 1: These changes does not align with the [spec 5.4](https://a2a-protocol.org/latest/specification/#54-error-code-mappings).

Can you cite some TCK specification that requires such change? Or is there any other source of truth for that?

src/server/request_handler/default_request_handler.ts
JakubWorek: Intentional. The bus is kept only for INPUT_REQUIRED / AUTH_REQUIRED so a follow-up message/send resumes via createOrGetByTaskId and resubscribers can still attach (spec 3.4.3).

That's the spec mechanism that requires our bus to stay alive after the executor returns in INPUT_REQUIRED: the follow-up message/send arrives with the same taskId and must reach the same execution context, which createOrGetByTaskId does by returning the still-live bus.

The keep-alive behavior is necessary for TCK (14 STREAM-* MUST tests break without it).

src/server/request_handler/default_request_handler.ts
JakubWorek: response in [here](https://github.com/a2aproject/a2a-js/pull/517#discussion_r3420456152)

src/errors.ts
JakubWorek: comment removed

src/server/express/rest_handler.ts
JakubWorek: None of sdks uses `a2a+json` to be complient with tck, which is an issue - missmatch between it and spec. Let's follow the pattern and wait till tck be spec complient.

PR #524, opened 2026-06-19
fix(server): merge Task history in ResultManager instead of overwriting

### LINKED ISSUE #532: fix(server): Merge Task history in ResultManager instead of overwriting
In `src/server/result_manager.ts`, replace wholesale Task replacement
with merge logic for the `task` event:

1. Load the existing persisted task.
2. If the incoming Task has empty `history` but the persisted Task has a
   non-empty history, preserve the persisted history.
3. Append new artifacts/messages instead of wholesale replace.

JakubWorek:
# Description

## What

`ResultManager.processEvent('task')` merges the incoming Task with the persisted
Task (preserving history + artifacts) instead of wholesale replacement.

## Why

Multi-turn conversations: after INPUT_REQUIRED → follow-up, the executor often
publishes a fresh Task event without history. Wholesale replacement lost the
entire conversation history.

Closes #532

### DIFF
--- a/src/server/result_manager.ts
+++ b/src/server/result_manager.ts
@@ -10,6 +10,49 @@ import { AgentExecutionEvent, assertUnreachableEvent } from './events/execution_
 import { TaskStore } from './store.js';
 
 /**
+ * Per-`taskId` write serializer shared across every `ResultManager`
+ * instance in this process.
+ *
+ * Multiple `ResultManager`s can run concurrently against the same task:
+ * - The AUTH_REQUIRED background drain holds one RM while a follow-up
+ *   `sendMessage` constructs a new one (spec §7.6).
+ * - INPUT_REQUIRED's `createOrGetByTaskId` bus reuse + a follow-up call.
+ * - `cancelTask` overlapping with the background drain.
+ * - Non-blocking `sendMessage` returning early while its drain continues.
+ *
+ * Each `processEvent` performs a load-merge-save sequence on the
+ * `TaskStore`. Without serialization, two RMs racing on the same row
+ * would each read a pre-sibling snapshot, merge against it, and the last
+ * writer would silently overwrite the other's contribution (classic
+ * lost-update). Wholesale-replace was bug-equivalent for both writers;
+ * merge semantics make the race observable as missing history /
+ * artifacts.
+ *
+ * Scope: in-process only. Cross-process serialization belongs in the
+ * `TaskStore` interface (compare-and-swap or transactional `update`) and
+ * is intentionally out of scope here.
+ */
+const taskWriteLocks = new Map<string, Promise<unknown>>();
+
+async function serializeByTaskId<T>(taskId: string, fn: () => Promise<T>): Promise<T> {
+  const prev = taskWriteLocks.get(taskId) ?? Promise.resolve();
+  // `.then(fn, fn)` ensures one RM's rejection doesn't block the chain —
+  // the next queued function still runs.
+  const next = prev.then(fn, fn);
+  taskWriteLocks.set(
+    taskId,
+    next.finally(() => {
+      // Only evict if no newer call has extended the chain. Without this
+      // guard, a fast successor would be orphaned by our cleanup.
+      if (taskWriteLocks.get(taskId) === next) {
+        taskWriteLocks.delete(taskId);
+      }
+    })
+  );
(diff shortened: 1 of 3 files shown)


src/server/result_manager.ts
```diff
@@ -27,25 +47,85 @@ export class ResultManager {
...
+        // warns on misses). We also re-load if the in-memory task is for
+        // a different id, otherwise we'd lose the persisted state for the
+        // new task id.
+        if ((!this.currentTask || this.currentTask.id !== taskEvent.id) && taskEvent.id) {
+          const loaded = await this.taskStore.load(taskEvent.id, this.serverCallContext);
+          if (loaded) {
+            this.currentTask = loaded;
+          } else if (this.currentTask && this.currentTask.id !== taskEvent.id) {
+            // The previously-tracked task is unrelated to the incoming
+            // event; drop it so we don't accidentally merge across ids.
```
bartek-gralewicz: A warning log in a case that such thing would occur would be advised.

bartek-gralewicz (commented):
What about recently added [PR #523](https://github.com/a2aproject/a2a-js/pull/523)? Is the idea of draining the event for AUTH_REQUIRED compatible with modifying the current task implemented now?

src/server/result_manager.ts
JakubWorek: Added `console.warn` via notePersistedTaskId — fires when an executor publishes events for a different taskId on the same RM

JakubWorek:
> What about recently added [PR #523](https://github.com/a2aproject/a2a-js/pull/523)? Is the idea of draining the event for AUTH_REQUIRED compatible with modifying the current task implemented now?

That PR enables two RMs writing to the same taskId (background drain + follow-up sendMessage); merge semantics turned that pre-existing race from bug-equivalent-overwrite into a lost-update. Added a per-taskId serializer in result_manager.ts so inside the lock we always reload from the store, so neither RM clobbers the other's write. Covered by the new concurrent ResultManagers on the same taskId test block, including the explicit AUTH_REQUIRED scenario.

src/server/result_manager.ts
```diff
@@ -1,132 +1,372 @@
...
+    taskId,
+    next.finally(() => {
+      // Only evict if no newer call has extended the chain. Without this
+      // guard, a fast successor would be orphaned by our cleanup.
+      if (taskWriteLocks.get(taskId) === next) {
+        taskWriteLocks.delete(taskId);
+      }
+    })
+  );
+  return next;
```
bartek-gralewicz, line 77: I believe that `=== next` will be never hit since the `.finally()` before that returns a `Promise`. Capturing the promise and comparing against it should work. 

```suggestion
  const wrapped = next.finally(() => {
    if (taskWriteLocks.get(taskId) === wrapped) {
      taskWriteLocks.delete(taskId);
    }
  );
  taskWriteLocks.set(taskId, wrapped);
  return next;
```

src/server/result_manager.ts
```diff
@@ -1,132 +1,372 @@
...
+ * writer would silently overwrite the other's contribution (classic
+ * lost-update). Wholesale-replace was bug-equivalent for both writers;
+ * merge semantics make the race observable as missing history /
+ * artifacts.
+ *
+ * Scope: in-process only. Cross-process serialization belongs in the
+ * `TaskStore` interface (compare-and-swap or transactional `update`) and
+ * is intentionally out of scope here.
+ */
+const taskWriteLocks = new Map<string, Promise<unknown>>();
```
bartek-gralewicz, line 47: The `TaskStore` operates on tenant-scoped basis but this locks mechanism in based purely on task ids. If two tenants reuse the same task id, this approach will serialize them against each other. I suggest to include the tenant in the lock as a top level mapping and then use IDs (just like the in memory task store).

JakubWorek: Added a `lockKey` helper that builds key, mirroring taskstore `tenant, owner, taskId`

src/server/result_manager.ts
JakubWorek: simplified it to store `next` directly in the map and schedule eviction via `.then(evict, evict)` which avoids .finally rejection re-throw that triggered unhandled-rejection warnings during testing

bartek-gralewicz (approved):

PR #525, opened 2026-06-19
fix(server): synthesize correct error Task id in blocking and streaming paths

### LINKED ISSUE #533: fix(server): Fix synthetic error Task in blocking and streaming executor paths
The executor error path is broken in two ways that share a root cause
(`requestContext.taskId` is not used as the synthetic Task id):

1. **Blocking path** (`src/server/request_handler/default_request_handler.ts:347`,
   `_runExecutor`): change

   ```ts
   id: requestContext.task?.id || uuidv4(),
   ```

   to:

   ```ts
   id: requestContext.taskId,
   ```

   The fabricated UUID didn't match the event bus registration key, so the
   client received an id that resolves to TaskNotFoundError on subsequent
   `getTask`.

2. **Streaming path** (`src/server/request_handler/default_request_handler.ts:453-457`,
   `_runStreamExecutor`): when `currentTask` is undefined after executor
   failure, fabricate a synthetic Task + statusUpdate(FAILED) using
   `requestContext.taskId` and publish them through `eventBus.publish(...)`
   before returning. Mirror the synthesis pattern from `_runExecutor`.

   Currently the streaming path silently returns, leaving the client with an
   empty stream and no indication of failure — asymmetric with the blocking
   path.

JakubWorek:
# Description

## What

Two coordinated fixes to the executor error path, both rooted in not using
`requestContext.taskId` for the synthetic error Task:

- Blocking `_runExecutor`: synthetic Task id is `requestContext.taskId`
  (was `requestContext.task?.id || uuidv4()`).
- Streaming `_runStreamExecutor`: when the executor throws before publishing
  any Task event, synthesize Task + statusUpdate(FAILED) instead of silently
  closing the stream.

## Why

- Blocking: the fabricated UUID didn't match the event bus registration key,
  so `getTask(id)` with the returned id raised TaskNotFoundError. The client
  had no way to learn the task failed.
- Streaming: silent empty stream is asymmetric with the blocking path (which
  always synthesizes the error Task) and makes production debugging impossible.

Closes #533

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -447,8 +447,13 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         // Publish a synthetic error event so the consumer's event loop
         // can settle the first-result promise and so any concurrent
         // resubscribers see the failure on the wire.
+        //
+        // The synthetic Task id MUST be `requestContext.taskId` — that's
+        // the id the bus is registered under and the id we hand back to
+        // the client. Fabricating a fresh `uuidv4()` here would make the
+        // returned Task unreachable via `getTask` (TaskNotFoundError).
         const errorTask: Task = {
-          id: requestContext.task?.id || uuidv4(),
+          id: requestContext.taskId,
           contextId: finalMessageForAgent.contextId!,
           status: {
             state: TaskState.TASK_STATE_FAILED,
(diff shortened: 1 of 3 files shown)


src/server/request_handler/default_request_handler.ts
```diff
@@ -540,23 +545,64 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+   * `throw` — which is exactly the typical executor pattern. See
+   * `trackLatestPublishedTask` and `_runExecutor` for the same
+   * rationale.
    */
   private _runStreamExecutor(
     taskId: string,
     eventBus: ExecutionEventBus,
     requestContext: RequestContext,
-    resultManager: ResultManager
+    _resultManager: ResultManager
```
bartek-gralewicz: Why ignoring an argument on a private method? Isn't removing it if not needed a better approach?

src/server/request_handler/default_request_handler.ts
```diff
@@ -540,23 +545,64 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
   ): void {
     const finalMessageForAgent = requestContext.userMessage;
     // See `_runExecutor` for why we snoop the bus directly instead of
-    // re-reading state from `resultManager` in the `.finally` block.
-    const stateTracker = trackLatestTaskState(eventBus);
+    // re-reading state from `resultManager` in the `.finally` block,
+    // and `trackLatestPublishedTask` for the analogous reasoning in
+    // the `.catch` block below.
+    const rawStateTracker = trackLatestTaskState(eventBus);
+    const publishedTaskTracker = trackLatestPublishedTask(eventBus);
```
bartek-gralewicz: Would be best to use a single listener for both `state` and `task` since they are connected to the same `eventBus` to avoid double listener calls per-execution.

src/server/request_handler/default_request_handler.ts
JakubWorek: removed the parameter and updated the call site

src/server/request_handler/default_request_handler.ts
JakubWorek: Collapsed both trackers into a single trackLatestTaskAndState listener returning {task, state}

bartek-gralewicz (approved):

PR #554, opened 2026-06-29
fix(samples): always publish Task snapshot + migrate to @genkit-ai/google-genai

JakubWorek:
# Description

Two sample-related fixes surfaced while validating samples and ITK.

## Sample fixes

- **4 sample executors** (`sample-agent`, `movie-agent`, `cancellable-agent`, `push-notification-agent`): Removed the `if (!existingTask)` guard around the initial Task publish. The guard broke follow-up turns on tasks in non-terminal states (e.g. after `INPUT_REQUIRED`) by emitting a `statusUpdate` first, which the SDK's stream-pattern validation rejects with `Received statusUpdate before initial 'Message'/'Task' event`. Each executor now publishes a Task snapshot every turn — replaying the existing task with the new user message appended for follow-ups.
- **`movie-agent` Genkit upgrade**: Migrated from `@genkit-ai/googleai@1.8` to `@genkit-ai/google-genai@1.38` (the official replacement plugin). Fixes `Function call is missing a thought_signature` errors with Gemini 2.5+/3.x during multi-turn tool calling (upstream: genkit-ai/genkit#4217). Also bumps the model to `gemini-3.1-flash-lite`.

### DIFF
--- a/src/samples/agents/movie-agent/genkit.ts
+++ b/src/samples/agents/movie-agent/genkit.ts
@@ -5,7 +5,7 @@ import { fileURLToPath } from 'url';
 
 export const ai = genkit({
   plugins: [googleAI()],
-  model: googleAI.model('gemini-flash-latest'),
+  model: googleAI.model('gemini-3.1-flash-lite'),
   promptDir: dirname(fileURLToPath(import.meta.url)),
 });
 
--- a/src/samples/agents/movie-agent/genkit.ts
+++ b/src/samples/agents/movie-agent/genkit.ts
@@ -1,11 +1,11 @@
-import { googleAI } from '@genkit-ai/googleai';
+import { googleAI } from '@genkit-ai/google-genai';
 import { genkit } from 'genkit';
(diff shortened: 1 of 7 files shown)


src/samples/agents/movie-agent/genkit.ts
```diff
@@ -1,11 +1,11 @@
-import { googleAI } from '@genkit-ai/googleai';
+import { googleAI } from '@genkit-ai/google-genai';
 import { genkit } from 'genkit';
 import { dirname } from 'path';
 import { fileURLToPath } from 'url';
 
 export const ai = genkit({
   plugins: [googleAI()],
-  model: googleAI.model('gemini-2.5-flash'),
+  model: googleAI.model('gemini-flash-latest'),
```
bartek-gralewicz: Let's set this to a fixed version as results vary when a model is updated. An update between some older flash models, broke this sample before.

bartek-gralewicz (approved):

PR #569, opened 2026-07-09
fix: support GET for resubscribe route

JakubWorek:
# Description

## Context
The v1.0 A2A spec has two normative sources — the proto file and the spec markdown (per §1.4 Normative Content (https://a2a-protocol.org/latest/specification/#14-normative-content)). They disagree on the HTTP method for SubscribeToTask:
- Proto (spec/a2a.proto): google.api.http annotation says get: "/tasks/{id=*}:subscribe"
- Spec markdown (§11.3.2 (https://a2a-protocol.org/latest/specification/#1132-task-operations)): documents it as POST

The v0.3 proto (google.api.http annotation on SubscribeToTask) also uses get:, and reference SDKs consistently followed the proto binding:
- a2a-python: ('/v1/tasks/{id}:subscribe', 'GET') in rest_adapter.py (https://github.com/a2aproject/a2a-python/blob/ead75f95b852810f1837248b9e9ffc5092c2d463/src/a2a/server/apps/rest/rest_adapter.py#L218)
- a2a-go: GET binding in its REST server
- a2a-js v1.0 (RestTransport): issues GET on the client side
- a2a-rust a2a-client-lf: issues GET on the client side

`@a2a-js/sdk@0.3.x` server was the outlier — only POST was registered. When any spec-compliant client (v0.3-native or v1.0-with-compat) sent GET, Express fell through to the more generic GET /v1/tasks/:taskId route, capturing :taskId = "<uuid>:subscribe" and throwing TaskNotFoundError: Task not found: <uuid>:subscribe — cross-SDK interop broken for resubscribe over REST.

## What this PR does
Adds a GET binding for /v1/tasks/:taskId:subscribe alongside the existing POST route:
- GET is the canonical binding per the proto's google.api.http annotation and is what the ecosystem's clients send.
- POST is retained for backward compatibility with clients that were built against the spec markdown or against previous `@a2a-js/sdk@0.3.x` behavior.
Both methods route to the same handler. The subscribe route is registered before the generic GET /v1/tasks/:taskId route so Express's greedy :taskId matcher doesn't swallow /v1/tasks/xyz:subscribe requests.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -27,7 +27,7 @@ exports[`TypeScript Strict Mode`] = {
       [248, 15, 4, "tsc: Expected 2 arguments, but got 1.", "2087764327"],
       [271, 15, 4, "tsc: Expected 2 arguments, but got 1.", "2087764327"]
     ],
-    "src/server/express/rest_handler.ts:2436698925": [
+    "src/server/express/rest_handler.ts:2980194902": [
       [204, 50, 17, "tsc: Argument of type \'unknown\' is not assignable to parameter of type \'Message | Task | TaskArtifactUpdateEvent | TaskStatusUpdateEvent\'.", "3749434707"]
     ],
     "src/server/request_handler/default_request_handler.ts:1208203039": [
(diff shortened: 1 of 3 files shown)


ishymko (approved):
nit: can we expand description a bit with why are we doing this?

PR #577, opened 2026-07-13
build: enable strict mode in tsconfig

### LINKED ISSUE #182: [Feat]: Enable strict mode in TSC conf
We should research which enforcements can be enabled to improve compile time checks. Very likely it'll require updating significant amount of code.

JakubWorek:
# Description

This PR removes betterer in favour of enabling strict mode in tsconfig for /src and base config based on previous one for other directories

Closes #182  🦕


ishymko (commented):
I actually agree with AI comments here, let's keep `tsconfig.json` strict and limit it to `src` and `tck` (as in the original betterer config, looks like `tck` is omitted now) and create a custom one for tests which is non-strict.

tsconfig.test.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.base.json",
+  "include": ["test/**/*.ts", "tck/**/*.ts"]
```
ishymko: Original betterer config had `tck` in strict mode, should we do it here as well?

ishymko (approved):
nit: I changed PR prefix from `feat` to `build` as `feat` would be added into a changelog which is no-op for package consumers.

PR #579, opened 2026-07-15
fix(server): await _handleProcessingError so blocking drains surface errors

RenKoya1:
# Description

`_processEvents` catches errors from the event-processing loop and delegates to
`_handleProcessingError`, which — per its own comment — re-throws on the blocking
path (no `firstResultRejector`) *"so the caller's await catches it"*:

```ts
} catch (error) {
  console.error(`Event processing loop failed for task ${taskId}:`, error);
  this._handleProcessingError(          // <-- not awaited
    error, resultManager, firstResultSent, taskId, options?.firstResultRejector,
  );
} finally {
  eventQueue.stop();
}
```

```ts
// Blocking: re-throw so the caller's await catches it.
if (!firstResultRejector) {
  throw error;
}
```

The call is **not awaited**. `_handleProcessingError` is `async`, so its `throw`
becomes a floating rejection instead of propagating out of `_processEvents`. The
generator resolves as if the drain had succeeded.

The one blocking caller that passes no `firstResultRejector` is `cancelTask`'s
drain:

```ts
await this._processEvents(
  taskId, new ResultManager(this.taskStore, context), eventQueue, context,
);
```

So when the drain throws (e.g. the task store fails while persisting the
cancellation), `cancelTask` silently swallows the real error, reloads the task,
sees a non-canceled state, and throws a misleading `TaskNotCancelableError` —
while the true error surfaces as an unhandled promise rejection.

## Fix

Await `_handleProcessingError`. The blocking re-throw now propagates through the
`finally` (which still detaches the queue) and out to the awaiting caller,
matching the documented intent.

The non-blocking paths are unaffected: `sendMessage`/`sendMessageStream` pass a
`firstResultRejector`, so `_handleProcessingError` rejects or persists a FAILED
status rather than throwing.

## Tests

Added a regression test asserting `cancelTask` surfaces a drain-time persistence
error instead of masking it with `TaskNotCancelableError`. Verified it fails
without the fix (throws `TaskNotCancelableError`) and passes with it.

Full suite green (1368 tests).

- [x] Follows the `CONTRIBUTING` guide
- [x] PR title uses Conventional Commits (`fix:`)
- [x] Tests and linter pass
- [ ] Docs updated (not necessary)

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -322,7 +322,11 @@ export class DefaultRequestHandler implements A2ARequestHandler {
       }
     } catch (error) {
       console.error(`Event processing loop failed for task ${taskId}:`, error);
-      this._handleProcessingError(
+      // Must be awaited: `_handleProcessingError` re-throws in the blocking
+      // (no `firstResultRejector`) case so the caller's `await` catches it.
+      // Without `await`, that throw escapes as a floating rejection and the
+      // caller (e.g. `cancelTask`'s drain) resolves as if nothing failed.
+      await this._handleProcessingError(
         error,
         resultManager,
         firstResultSent,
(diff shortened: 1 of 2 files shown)


JakubWorek (approved):
Hi @RenKoya1, thank you for spoting this out and fixing.

PR #581, opened 2026-07-15
feat(server)!: replace individual properties with SendMessageRequest in RequestContext

JakubWorek:
# Description

In order to simplify the interface of `RequestContext`, provide more data from the request and mimic [Python sdk](https://github.com/a2aproject/a2a-python/blob/0d82ab985ffbac381a38adb5bd160674fa05eb18/src/a2a/server/agent_execution/context.py#L32) it adds `request: SendMessageRequest` as a member of `RequestContextParams`.

### DIFF
--- a/src/server/agent_execution/request_context.ts
+++ b/src/server/agent_execution/request_context.ts
@@ -1,36 +1,54 @@
-import { Message, Task } from '../../index.js';
+import { Message, SendMessageRequest, Task } from '../../index.js';
 import { ServerCallContext } from '../context.js';
 
-export class RequestContext {
-  public readonly userMessage: Message;
-  public readonly taskId: string;
-  public readonly contextId: string;
-  public readonly task?: Task;
-  public readonly referenceTasks?: Task[];
-  public readonly context: ServerCallContext;
-  /**
-   * The request-level metadata from the originating `SendMessageRequest`,
-   * when provided. This is the spec's "flexible key-value map for passing
-   * additional context or parameters" and is distinct from
-   * `userMessage.metadata`.
-   */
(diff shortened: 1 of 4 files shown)


src/server/agent_execution/request_context.ts
```diff
@@ -1,36 +1,54 @@
-import { Message, Task } from '../../index.js';
+import { Message, SendMessageRequest, Task } from '../../index.js';
 import { ServerCallContext } from '../context.js';
 
+export interface RequestContextParams {
```
ishymko: nit: looks a bit confusing to me to have params for *almost* a plain DTO object constructor. Also if we do we need to reexport for the public API (which I'd also like to avoid) as `RequestContext` is exported.

src/server/agent_execution/request_context.ts
```diff
@@ -1,36 +1,54 @@
...
+
+/**
+ * Holds information about the current request being processed by the server.
+ *
+ * Wraps the incoming {@link SendMessageRequest} so agent executors can reach
+ * the full payload (message, configuration, metadata, tenant) via `request`.
+ */
 export class RequestContext {
-  public readonly userMessage: Message;
+  /** The incoming request payload, including `message`, `configuration`, `metadata`, and `tenant`. */
```
ishymko: nit: let's not list object properties in the comment, the code should be source of truth.

src/server/agent_execution/request_context.ts
ishymko, line 1: PR title should contain `!` as it's a public API breaking change (`feat(server)!`). Also maybe `replace individual properties with SendMessageRequest in RequestContext`?

ishymko (approved):

PR #582, opened 2026-07-16
fix(client): bound SSE event size to prevent client-side memory exhaustion

RenKoya1:
# Description

`parseSseStream` (shared by the JSON-RPC and REST client transports)
accumulates two buffers with **no size bound**:

- the line `buffer` — `buffer += value` on every chunk
- an event's joined `eventData` — consecutive `data:` lines are appended

Because A2A clients stream SSE from remote, potentially untrusted agent
servers, a malicious or broken server can exploit this as a **denial-of-service
against the client**:

- stream bytes that **never contain a `\n`** → the line `buffer` grows without bound
- stream endless `data:` lines with **no terminating blank line** → `eventData` grows without bound

Either way the client keeps allocating without bound. Today that can result in
a process-level OOM crash that the caller cannot handle as a normal stream
error.

## Fix

Cap both accumulation points at `maxEventSizeBytes` and throw when exceeded.
The throw runs the async generator's teardown, whose `finally` now cancels the
reader (#580), so the offending connection is also closed. Net effect: a
**process-level crash becomes a catchable `Error`** that flows into the
transports' existing SSE-parse error handling.

The same limit applies to the unterminated line buffer because a single SSE
field line is necessarily part of an event and should not be allowed to exceed
the maximum event budget on its own.

- Default is **4 MiB** (matching gRPC's default max message size): realistic
  A2A events (Message/Task JSON) are KB-scale, and large files should be
  referenced via `FileWithUri` parts rather than inlined.
- It is **configurable** via the new optional `maxEventSizeBytes` argument for
  callers that must inline larger payloads, and the error message points at
  both options.

## Open question — is this the right layer?

I'm **not sure `parseSseStream` is the right place** for this, and would
appreciate maintainer guidance. Alternatives worth considering:

- enforcing it at the transport layer instead of the shared parser,
- deriving/validating against `Content-Length` where present,
- or leaving DoS mitigation entirely to the runtime / a reverse proxy.

Happy to change the default, make it opt-in, or move it elsewhere based on
what you prefer.

## Tests

Adds regression tests for the unterminated-line, terminated-oversized-line
(single-chunk), and unterminated-event cases — all would grow unbounded
without the cap — plus a within-limit sanity check. `test/sse_utils.spec.ts`
is green.

- [x] Follows the `CONTRIBUTING` guide
- [x] PR title uses Conventional Commits (`fix:`)
- [x] Tests and linter pass
- [ ] Docs updated (not necessary)

### DIFF
--- a/src/sse_utils.ts
+++ b/src/sse_utils.ts
@@ -61,9 +61,11 @@ export function formatSSEErrorEvent(error: unknown): string {
 // ============================================================================
 
 /**
- * Upper bound (in UTF-16 code units, a conservative over-approximation of
- * byte length) on how large a single unterminated line or a single SSE event
- * may grow before {@link parseSseStream} aborts.
+ * Upper bound on how large a single unterminated line or a single SSE event
+ * may grow before {@link parseSseStream} aborts. Measured in UTF-16 code
+ * units (JS string length), which equals byte length for ASCII payloads such
+ * as JSON-encoded events; memory stays bounded within a small constant
+ * factor of this value either way.
  *
  * A2A clients stream from remote, potentially untrusted agent servers.
  * Without this cap a malicious or broken server can stream bytes that never
--- a/src/sse_utils.ts
+++ b/src/sse_utils.ts
@@ -104,6 +104,11 @@ export async function* parseSseStream(
     let lineEndIndex: number;
 
     while ((lineEndIndex = buffer.indexOf('\n')) >= 0) {
+      if (lineEndIndex > maxEventSizeBytes) {
+        throw new Error(
+          `SSE line exceeded the maximum allowed size of ${maxEventSizeBytes} bytes.`
+        );
+      }
       // Per the SSE spec lines may end with `\r\n`, `\r`, or `\n`. We
       // strip a trailing `\r` explicitly rather than calling `.trim()`,
       // which would also eat whitespace inside JSON-formatted `data:`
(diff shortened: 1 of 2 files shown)


RenKoya1:
Good catch — fixed in f16160c. The residual `buffer.length` check only fired for lines that never terminate; a large **terminated** line was drained by the split loop and slipped past. Now the limit is enforced inside the loop on `lineEndIndex` before extracting the line.

One note on the suggested test: with the 2-byte `createMockResponse` chunking, an oversized line trips the residual check *before* its newline arrives, so it would pass even without the fix. I added a regression test that delivers the whole terminated line in a **single chunk** — it fails on the pre-fix code and passes now, so it actually exercises the in-loop guard.

src/sse_utils.ts
```diff
@@ -60,13 +60,36 @@ export function formatSSEErrorEvent(error: unknown): string {
...
+ * Without this cap a malicious or broken server can stream bytes that never
+ * form a complete line — or `data:` lines whose blank-line terminator never
+ * arrives — growing an in-memory buffer without limit until the client
+ * process exhausts memory (CWE-400, uncontrolled resource consumption). The
+ * default is deliberately generous so that legitimate large JSON events
+ * (e.g. base64 file parts) still pass; callers handling larger payloads can
+ * raise it via the `maxEventSizeBytes` argument.
+ */
+export const DEFAULT_MAX_SSE_EVENT_SIZE_BYTES = 20 * 1024 * 1024; // 20 MiB
+
```
JakubWorek, line 83: A realistic size of events flowing through `parseSSEStream` is measured in KB not MB - JSONs with Message/Task/... . Comparable default for gRPC is 4MB and even less in nginx which is 1MB. A2A supports file with uri so large files are not inlined into events. 
We should consider lowering the limit and throwing more reasonable message on error - eg. `SSE event data exceeded the maximum allowed size of 4194304 bytes. Pass maxEventSizeBytes to raise the limit, or prefer FileWithUri parts for large payloads.`

RenKoya1: Agreed on both points — done in ee27bac. Default lowered to 4 MiB (gRPC's default max message size), and the error now reads: `... exceeded the maximum allowed size of 4194304 bytes. Pass maxEventSizeBytes to raise the limit, or prefer FileWithUri parts for large payloads.` Doc comment updated to reflect the FileWithUri guidance as well.

JakubWorek (approved):

Issue #350, opened 2026-03-10
[Feat]: Integrated SSRF and DNS Rebinding Protection across all Transports


alisa-alisa:
### Is your feature request related to a problem? Please describe.

The Gemini CLI team is integrating the A2A SDK to support remote agents. To meet production security standards, it is critical that the CLI is protected against Server-Side Request Forgery (SSRF) and DNS Rebinding attacks.

We believe these security guardrails should be a native part of the SDK to ensure all consumers are "secure by default."

### Describe the solution you'd like

Requested Features:
1. **Transport-Agnostic Connection Filtering:** A unified mechanism (e.g., a connectionInterceptor or lookup hook) in the ClientFactory that applies to all transports (REST, JSONRpc, and gRPC) to block access      to private/internal IP ranges.
1. **DNS Rebinding Protection (IP Pinning):** Support for "pinning" a connection to the specific IP address resolved during the initial Agent Card discovery phase. This is necessary to prevent      Time-of-Check/Time-of-Use (TOCTOU) attacks where an attacker swaps a hostname's DNS record between discovery and tool execution.
1. **Secure Card Validation:** Built-in validation in the DefaultAgentCardResolver to automatically reject agent cards if the card URL or any nested interface URLs point to restricted IP ranges.

### Describe alternatives you've considered

We attempted to implement these protections at the application layer, but found it impossible to achieve full coverage without "hacking" SDK internals—especially for gRPC, which does not respect the same connection-level hooks as HTTP-based transports.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chopmob-cloud:
We just landed the equivalent protection in the Go SDK ([a2a-go #374](https://github.com/a2aproject/a2a-go/pull/374)), so here is how the same design maps onto a2a-js, plus a working implementation and the two decisions worth settling before a PR.

**The core principle (why app-layer cannot cover it):** the check has to run against the IP that is actually dialled, at connect time, on every connection. Validating the hostname up front during card discovery does not stop rebinding, because the attacker re-answers DNS between discovery and the RPC call. In Go we did this with `net.Dialer.Control`. The Node equivalent is an undici `Agent` with a custom `connect.lookup`: the lookup resolves, validates every resolved address against a reserved-range block list, and only then hands the socket the pinned address. Redirects follow through the same dispatcher, so a redirect to an internal target is validated too.

That maps cleanly onto the three asks:

[...]

PR #586, opened 2026-07-17
fix(server): validate push-notification webhook URLs against SSRF per…

### LINKED ISSUE #584: [Bug]: push-notification webhook URL is not SSRF-validated (A2A spec §13.2 unimplemented)
### Summary

#350 asks for SSRF and DNS-rebinding protection on the **client** side (`ClientFactory`, agent-card
discovery, `DefaultAgentCardResolver`). This issue is the **server-side half of the same problem**, which
#350 doesn't cover: the SDK's push-notification sender POSTs to a client-supplied webhook URL with no
validation at all.

A2A spec §13.2 puts this on the agent:

> Agents **SHOULD** validate webhook URLs to prevent SSRF (Server-Side Request Forgery) attacks: Reject
> private IP ranges (127.0.0.0/8, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16); Reject localhost and
> link-local addresses; Implement URL allowlists where appropriate.

Verified at `cfeed0d` (v0.3.14). The same gap is tracked for the Python SDK in a2a-python#786.

### Where

No validation exists on any hop of the server-side path:

- `src/server/request_handler/default_request_handler.ts:229` (`sendMessage`) and `:331`
  (`sendMessageStream`) — save an inline `configuration.pushNotificationConfig`, gated **only** on
  `agentCard.capabilities.pushNotifications`. `:449` (`setTaskPushNotificationConfig`) — same, no check.
- `src/server/push_notification/push_notification_store.ts` — `save()` stores the URL as-is.
- `src/server/push_notification/default_push_notification_sender.ts` — `_dispatchNotification`:
  `const url = pushConfig.url; await fetch(url, { method: 'POST', ... })`.

### Attack scenario

- **Victim:** an operator running an agent built on `@a2a-js/sdk` with push notifications enabled
  (`capabilities.pushNotifications: true` + the SDK's `DefaultPushNotificationSender`), deployed
  somewhere with internal-only services (admin APIs, sidecars, cloud metadata).
- **Attacker:** any client the agent will serve. No elevated privilege required — and because the config
  can be supplied inline on `message/send`, **no pre-existing task is required either**; the attacker's
  first request is enough.
- **Result:** the agent POSTs the task payload to an attacker-chosen internal address, from inside the
  victim's network, carrying the attacker's `X-A2A-Notification-Token`.

### Reproduction

Wire an agent with the SDK's own components (`DefaultRequestHandler` + `InMemoryTaskStore` +
`InMemoryPushNotificationStore` + `DefaultPushNotificationSender`) and an agent card declaring
`capabilities: { pushNotifications: true }` — i.e. the wiring in your own
`test/server/push_notification_integration.spec.ts`. Start a listener on a loopback address to stand in
for an internal-only service, then send one message pointing the webhook at it:

```ts
await handler.sendMessage({
  message: { messageId: 'm1', role: 'user', parts: [{ kind: 'text', text: 'hi' }], kind: 'message' },
  configuration: {
    pushNotificationConfig: {
      id: 'cfg-1',
      url: 'http://127.0.0.1:<port>/internal-admin/keys', // §13.2 says reject 127.0.0.0/8
      token: 'attacker-chosen-token',
    },
  },
});
```

**Observed:** the loopback listener receives `POST /internal-admin/keys` with the task payload and
`x-a2a-notification-token: attacker-chosen-token`. **Expected per §13.2:** the URL is rejected.

The same happens via `sendMessageStream`. And `setTaskPushNotificationConfig` accepts every address class
the spec says to reject — no error for any of these:

```ts
'http://169.254.169.254/latest/meta-data/'  // link-local / cloud metadata
'http://10.0.0.1/admin'                     // private
'http://192.168.1.1/admin'                  // private
'http://172.16.0.1/admin'                   // private
'http://localhost:9999/x'                   // localhost by name
'file:///etc/passwd'                        // scheme unrestricted
```

### Impact

Any client of a push-enabled agent can make it issue attacker-directed POSTs from inside its trust
boundary: reaching internal-only services (including loopback/RFC-1918), cloud metadata
(`169.254.169.254`, environment-dependent), enumerating internal hosts/ports via timing/error
differences, and delivering task state to an attacker-chosen endpoint. Since the vulnerable sender and
store are the SDK's **default** components, every push-enabled agent built on `@a2a-js/sdk` inherits it.

### Note

`fetch()` is called with no `redirect` option, so it defaults to `follow`. A webhook URL that passes a
naive string check can `307`-redirect to an internal address and the SDK follows it, POST and body
preserved. So validating the URL string alone isn't enough — the resolved IP needs to be validated and
pinned. (This is the same DNS-rebinding concern #350 raises for the client side.)

### Proposed direction (looking for input from reviewers)

The intent is **not** to block private/loopback unconditionally — those are legitimate targets for
internal agents, sidecars, and dev setups. (a2a-python#895 was closed for validating unconditionally; the
maintainers' follow-up design is a2a-python#1023.) Mirroring that:

1. A **composable, opt-out** validator: reject private/loopback/link-local by default, with a host/CIDR
   allowlist.
2. Restrict scheme to HTTP(S) (or HTTPS-only, configurable).
3. **Validate, then connect to the pinned resolved IP**; set `redirect: 'error'` on the dispatch `fetch`.
4. Make it injectable so integrators can supply their own policy.

Since #350 wants the client-side half, one shared validator could serve both — which is what a2a-python#1023
scopes (`UrlValidator` + per-domain wrappers). Happy to implement it here if that direction looks right.

---

### Code of Conduct

Varun-S10:
# Description

This PR fixes issue #584 by adding URL validation for push notification webhooks as required by A2A Protocol Specification §13.2 to prevent Server-Side Request Forgery (SSRF) attacks.

### Changes Included:
- **URL Validation**: Added `validateWebhookUrl` to check webhook URLs upon receipt and before sending notifications.
- **SSRF Protection**: Rejects loopback (`127.0.0.1`, `localhost`), cloud metadata (`169.254.169.254`), private IP ranges (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`), and non-HTTP(S) schemes.
- **Redirect Protection**: Configured `redirect: 'error'` on fetch calls to prevent HTTP redirect bypasses.
- **Configurable Options**: Added optional settings (`allowLoopback`, `allowedHosts`) for internal testing and local setups.
- **Tests**: Added unit tests in `url_validator.spec.ts`. All 420 tests pass cleanly.

- [ ] Appropriate docs were updated (if necessary)

Fixes #584

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -32,6 +32,8 @@ export { DefaultPushNotificationSender } from './push_notification/default_push_
 export type { DefaultPushNotificationSenderOptions } from './push_notification/default_push_notification_sender.js';
 export type { PushNotificationStore } from './push_notification/push_notification_store.js';
 export { InMemoryPushNotificationStore } from './push_notification/push_notification_store.js';
+export { validateWebhookUrl, UrlValidationError } from './push_notification/url_validator.js';
+export type { UrlValidationOptions } from './push_notification/url_validator.js';
 
 export type { User } from './authentication/user.js';
 export { UnauthenticatedUser } from './authentication/user.js';
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -1,6 +1,7 @@
 import { Task, PushNotificationConfig } from '../../types.js';
 import { PushNotificationSender } from './push_notification_sender.js';
 import { PushNotificationStore } from './push_notification_store.js';
+import { validateWebhookUrl, UrlValidationOptions } from './url_validator.js';
 
 export interface DefaultPushNotificationSenderOptions {
   /**
@@ -11,12 +12,24 @@ export interface DefaultPushNotificationSenderOptions {
    * Custom header name for the token. Defaults to 'X-A2A-Notification-Token'.
    */
   tokenHeaderName?: string;
+  /**
+   * URL validation options to prevent SSRF per A2A Spec §13.2.
+   * By default, rejects private IP ranges, loopback/localhost, link-local addresses, and non-HTTP(S) schemes.
+   */
+  urlValidationOptions?: UrlValidationOptions;
+  /**
+   * Custom validator function to validate webhook URLs.
+   */
+  urlValidator?: (url: string) => void;
 }
 
 export class DefaultPushNotificationSender implements PushNotificationSender {
   private readonly pushNotificationStore: PushNotificationStore;
   private notificationChain: Map<string, Promise<unknown>>;
-  private readonly options: Required<DefaultPushNotificationSenderOptions>;
+  private readonly options: DefaultPushNotificationSenderOptions & {
+    timeout: number;
+    tokenHeaderName: string;
+  };
 
   constructor(
     pushNotificationStore: PushNotificationStore,
@@ -69,6 +82,14 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
     pushConfig: PushNotificationConfig
   ): Promise<void> {
     const url = pushConfig.url;
+
+    // Validate URL per A2A spec §13.2 before sending request
+    if (this.options.urlValidator) {
+      this.options.urlValidator(url);
+    } else {
+      validateWebhookUrl(url, this.options.urlValidationOptions);
+    }
+
     const controller = new AbortController();
     // Abort the request if it takes longer than the configured timeout.
     const timeoutId = setTimeout(() => controller.abort(), this.options.timeout);
@@ -87,6 +108,7 @@ export class DefaultPushNotificationSender implements PushNotificationSender {
         headers,
         body: JSON.stringify(task),
         signal: controller.signal,
+        redirect: 'error',
       });
 
       if (!response.ok) {
--- a/src/server/push_notification/url_validator.ts
+++ b/src/server/push_notification/url_validator.ts
@@ -1,0 +1,163 @@
+import net from 'net';
+
+export interface UrlValidationOptions {
+  /**
+   * Whether to allow loopback addresses (localhost, 127.0.0.0/8, ::1).
+   * Defaults to false.
+   */
+  allowLoopback?: boolean;
+
+  /**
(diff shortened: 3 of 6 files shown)


chopmob-cloud:
Reviewed this against the PR head. The validation is a solid improvement, and I want to flag one gap in `validateWebhookUrl` that is central to the threat model in #584 and #350: it checks the URL hostname as a string but never resolves it, so it does not cover the DNS case.

`validateWebhookUrl` is synchronous and only inspects `parsedUrl.hostname`. Literal IPs go through `net.isIP` plus the range checks, and `localhost` is rejected, but for an ordinary hostname `net.isIP(cleanHost)` returns 0, so no address check runs. A webhook whose hostname resolves (via attacker-controlled DNS) to `127.0.0.1`, `169.254.169.254`, or a `10.0.0.0/8` address passes validation, and the sender then fetches the internal target. Since an attacker supplies a hostname rather than a literal IP, this is the common path, and it is the DNS-rebinding case that #350 names in its title and that #584 inherits as the server-side half.

This reproduces directly against the PR head: `validateWebhookUrl('https://localtest.me/notify')` returns without throwing, yet `localtest.me` is a public name that resolves to `127.0.0.1`. The literal-IP cases (`127.0.0.1`, `10.0.0.1`, `169.254.169.254`) are correctly rejected, which is what makes the hostname path the gap.

The gemini suggestions (IPv6 `::`, CGNAT `100.64.0.0/10`) are worth taking, but they extend the literal-IP denylist and do not close this, because the hostname is never resolved in the first place.

A pure string check cannot cover it, since resolution happens later at connect time and a name can rebind between the check and the fetch (a TOCTOU window). The robust shape is to validate the resolved IP at connection time and pin the connection to the address that was validated. In Go, a2aproject/a2a-go #374 took this approach with a `DialContext` / `Control` hook that runs a resolved-address range check (`isBlockedIP`) on every hop (CWE-918). The Node equivalent is a custom `lookup` or agent that checks each resolved address before the socket connects; keeping `redirect: 'error'` on top remains necessary so a redirect cannot bypass it.

Happy to share the Node dial-time pattern if it helps.

PR #587, opened 2026-07-17
feat!: unify A2AError hierarchy with transport specific subclasses

### LINKED ISSUE #317: [Feat]: Passthrough http status code in error

### LINKED ISSUE #583: [Feat]: Unified A2AError hierarchy with transport-specific subclasses
## Problem

SDK errors are ~11 disjoint `Error` subclasses (plus v0.3's separate `A2AError`) wired together by 6+ parallel `error.name`-keyed lookup tables and duplicated `instanceof` chains in each transport. Consequences:

- No shared base — client and server rebuild the same mappings.
- Transport-native context is lost: HTTP status/headers, `cause`, and gRPC `grpc-status-details-bin` metadata never reach the caller (see #317).
- Adding an error class means touching 6+ tables and 3 mapping chains.

The two dimensions callers need to discriminate on — **what went wrong** (semantic) and **which wire carried it** (transport) — are orthogonal. JS single-inheritance means a class tree has to pick one dimension as parent. The proposal below picks _semantic_ as the class axis and models _transport_ as interfaces + user-defined type guards, so both dimensions remain reachable.

## Proposal

Class axis is semantic. `A2AError` is the shared base; each spec-aligned error extends it; each transport publishes a concrete per-transport subclass of every semantic error. Transport-native context is expressed as an **interface** implemented by all subclasses of a given transport, and narrowed at catch time via **user-defined type guards** (`isRestError`, `isGrpcError`, `isJsonRpcError`).

```mermaid
flowchart TD
    Error --> A2AError

    A2AError --> TNF[TaskNotFoundError]
    A2AError --> UOE[UnsupportedOperationError]
    A2AError --> Etc["…8 more semantic errors"]

    TNF --> RTNF[RestTaskNotFoundError]
    TNF --> GTNF[GrpcTaskNotFoundError]
    TNF --> JTNF[JsonRpcTaskNotFoundError]

    RE[[RestA2AError<br/><i>interface</i>]] -.implements.-> RTNF
    GE[[GrpcA2AError<br/><i>interface</i>]] -.implements.-> GTNF
    JE[[JsonRpcA2AError<br/><i>interface</i>]] -.implements.-> JTNF
```

- **`A2AError`** — extends `Error`; carries spec-aligned reason/code and structured metadata. Base for all semantic errors.
- **Semantic errors** (`TaskNotFoundError`, `UnsupportedOperationError`, `TaskNotCancelableError`, `ContentTypeNotSupportedError`, `InvalidAgentResponseError`, `PushNotificationNotSupportedError`, `ExtendedAgentCardNotConfiguredError`, `VersionNotSupportedError`, `RequestMalformedError`, `ExtensionSupportRequiredError`) — extend `A2AError`; carry error-specific fields (e.g., `taskId`).
- **Per-transport concrete subclasses** (`RestTaskNotFoundError`, `GrpcTaskNotFoundError`, `JsonRpcTaskNotFoundError`, and analogues for every semantic error) — extend the corresponding semantic class _and_ implement the transport interface.
- **Transport interfaces** (`RestA2AError`, `GrpcA2AError`, `JsonRpcA2AError`) — declare transport-native context:
  - `RestA2AError`: `statusCode`, `headers`, `cause`
  - `GrpcA2AError`: `status`, `statusDetailsBin` (no `cause` on the wire)
  - `JsonRpcA2AError`: `envelopeCode`, `data`
- **Type guards** (`isRestError(e)`, `isGrpcError(e)`, `isJsonRpcError(e)`) — user-defined type predicates that narrow an `A2AError` to the transport interface.

### Catch-site ergonomics

```ts
try {
  await client.getTask({ id });
} catch (e) {
  if (e instanceof TaskNotFoundError) {
    // Transport-agnostic: e.taskId is available.
    if (isRestError(e)) {
      // Narrowed: e.statusCode, e.headers, e.cause are typed.
      if (e.statusCode === 429) backoff(e.headers['retry-after']);
    }
  }
}
```

The TS compiler intersects the class type with the interface type inside the guard, so both semantic fields (`taskId`) and transport fields (`statusCode`) are simultaneously accessible.

Closes #317.

JakubWorek:
# Description

**Error Handling Refactor and Unification**
* All error classes now form a shared, transport-agnostic hierarchy with A2AError as the base and semantic subclasses (e.g., TaskNotFoundError, RequestMalformedError). Per-transport variants (e.g., RestTaskNotFoundError, GrpcTaskNotFoundError, JsonRpcTaskNotFoundError) extend their semantic parent and carry transport-native context (HTTP status/headers/cause, gRPC status/status-details-bin, JSON-RPC envelope code/data). Type guards isRestError / isGrpcError / isJsonRpcError narrow at catch time so callers keep instanceof TaskNotFoundError and simultaneously get typed access to transport fields. Closes #317.
* Errors live at two dedicated subpaths: @a2a-js/sdk/errors (pb-free, Workers-safe — base + semantic + REST + JSON-RPC) and @a2a-js/sdk/errors/grpc (requires @bufbuild/protobuf for grpc-status-details-bin encode/decode). The SDK root and /client / /server no longer re-export errors, so non-gRPC consumers don't pull in the pb peer dep.
* All transport implementations (rest, grpc, json-rpc, and legacy variants) now use centralized error mapping functions from the new modules (fromRestErrorBody, fromGrpcError, fromJsonRpcErrorResponse, toRestErrorBody, toJsonRpcError, buildGrpcErrorMetadata, restStatusFor, grpcStatusFor) instead of custom or scattered logic. Adding a new error is one row in A2A_ERROR_SPECS — all wire mappings derive from it.
* The v0.3 compat layer replaces LegacyA2AError with a thin facade over the new hierarchy that keeps the classic A2AError.taskNotFound(id) / new A2AError(code, msg, data?) API. Wire codes without a v1.0 semantic twin (PARSE_ERROR, INVALID_REQUEST, METHOD_NOT_FOUND) are preserved via JsonRpc*Error.envelopeCode, so v0.3 clients keep seeing the same numeric codes on the wire.

**Transport Implementation Simplification**
* Removed old error mapping methods from RestTransport, GrpcTransport, JsonRpcTransportHandler, and their v0.3 compat counterparts in favor of the centralized helpers, reducing code duplication and eliminating 6+ parallel error.name-keyed lookup tables.
* Added helpers to collect HTTP headers and provide richer error context in REST transport errors.
* Deleted src/errors.ts (354 lines) and src/server/grpc/error_details.ts (67 lines); folded into the new src/errors/ module.

**Build-tests Fix**
* The existing test-build script (esbuild --platform=neutral) never actually enforced Workers-safe boundaries — platform=neutral bundles Node-only modules silently, and CI's npm ci installs devDependencies (which include @grpc/grpc-js and @bufbuild/protobuf), so the check was a no-op. Added scripts/checkWorkersSafeBundles.js that fails if any Workers-safe bundle inlines @grpc/grpc-js or @bufbuild/protobuf. TDD-verified (goes RED when errors/grpc is re-exported from the pb-free barrel, GREEN with the split).

**Documentation and Migration Guide Updates**
* Updated the migration guide to explain the new error class structure, the two subpath entrypoints (@a2a-js/sdk/errors, @a2a-js/sdk/errors/grpc), and the transport-specific catch-site pattern with type guards, with code examples for the new patterns.

Closes #583 #317  🦕


src/compat/v0_3/client/transports/grpc/grpc_transport.ts
```diff
@@ -423,38 +418,9 @@ export class LegacyGrpcTransport implements Transport {
   /**
    * Decodes `google.rpc.ErrorInfo` from `grpc-status-details-bin` when
    * present (this SDK's `legacyGrpcService` emits it); otherwise returns
-   * a generic `Error` preserving the gRPC code and details.
+   * `GrpcGenericError` preserving the gRPC code and details.
    */
   private static _mapToError(error: grpc.ServiceError, method?: string): Error {
```
bartek-gralewicz: nit: We could just remove this local method since it only calls the imported `fromGrpcError` and use it directly where relevant.

src/index.ts
```diff
@@ -33,3 +33,8 @@ export {
   parseSseStream,
   type SseEvent,
 } from './sse_utils.js';
+
+// Transport-agnostic error hierarchy: base class, semantic subclasses,
+// per-transport variants (Rest*/Grpc*/JsonRpc*), and `is*Error` type
+// guards. Also re-exported from `./client` and `./server`.
```
bartek-gralewicz: The re-exports were removed in this PR so this comment is misleading. I would suggest to remove the last sentence from it.

bartek-gralewicz (approved):
Would be worth adding some tests that target newly added logic, like actually testing `instanceof`.

src/errors/grpc/index.ts
ishymko, line 1: As it depends on `@grpc/grpc-js` shouldn't it be somewhere under `grpc`? We now have [this](https://github.com/a2aproject/a2a-js/tree/epic/1.0_breaking_changes#for-grpc-usage) in the readme but I guess re-exporting all errors from `index.ts` hits it now for all clients?

JakubWorek: sure, I'll restructure the directories and exports, it actually breaks the contract of no grpc deps in index

JakubWorek: done, PR desc also adjusted

.github/workflows/build-tests.yml
```diff
@@ -15,16 +21,22 @@ on:
...
+        with:
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
+      - run: npm ci
+      - run: npm run build
+      # Strip the optional peer deps so esbuild sees the same module graph
+      # a non-gRPC consumer would. Any Workers-safe entrypoint that leaks
+      # a gRPC / pb import fails to resolve here.
+      - run: npm uninstall --no-save @grpc/grpc-js @bufbuild/protobuf
```
ishymko: Should we uninstall `express` as well? So that we uninstall all three optional dependencies before running these.

package.json
```diff
@@ -56,6 +62,16 @@
...
+    "./errors": {
+      "types": "./dist/errors/index.d.ts",
+      "import": "./dist/errors/index.js",
+      "require": "./dist/errors/index.cjs"
+    },
+    "./errors/grpc": {
+      "types": "./dist/errors/grpc/index.d.ts",
+      "import": "./dist/errors/grpc/index.js",
+      "require": "./dist/errors/grpc/index.cjs"
+    },
```
ishymko, line 74: nit: we likely need to update [this readme section](https://github.com/a2aproject/a2a-js/blob/epic/1.0_breaking_changes/README.md#for-grpc-usage) with this new entrypoint which will require grpc.

package.json
```diff
@@ -154,7 +170,9 @@
     "coverage": "vitest run --coverage",
     "generate": "curl https://raw.githubusercontent.com/google-a2a/A2A/refs/heads/main/specification/json/a2a.json > spec.json && node scripts/generateTypes.js && rm spec.json",
     "generate:compat": "curl https://raw.githubusercontent.com/a2aproject/A2A/v0.3.0/specification/json/a2a.json > compat_spec.json && node scripts/generateCompatTypes.js && rm compat_spec.json",
-    "test-build": "esbuild ./dist/client/index.js ./dist/server/index.js ./dist/index.js ./dist/compat/v0_3/index.js ./dist/compat/v0_3/client/index.js ./dist/compat/v0_3/server/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist",
+    "test-build:workers-safe": "esbuild ./dist/index.js ./dist/errors/index.js ./dist/client/index.js ./dist/server/index.js ./dist/compat/v0_3/index.js ./dist/compat/v0_3/client/index.js ./dist/compat/v0_3/server/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist",
+    "test-build:grpc": "esbuild ./dist/errors/grpc/index.js --bundle --platform=neutral --outdir=dist/tmp-checks --outbase=./dist",
```
ishymko: Should it actually be platform neutral? Maybe it doesn't cause problems for pure errors but overall grpc requires node APIs and won't work in browsers for instance. So real client/server usage won't build and outside of it there is not much value in pure errors.

(unless it validates it in some other way)

src/errors/base.ts
```diff
@@ -0,0 +1,403 @@
...
+  TASK_NOT_FOUND: -32001,
+  TASK_NOT_CANCELABLE: -32002,
+  PUSH_NOTIFICATION_NOT_SUPPORTED: -32003,
+  UNSUPPORTED_OPERATION: -32004,
+  CONTENT_TYPE_NOT_SUPPORTED: -32005,
+  INVALID_AGENT_RESPONSE: -32006,
+  EXTENDED_CARD_NOT_CONFIGURED: -32007,
+  EXTENSION_SUPPORT_REQUIRED: -32008,
+  VERSION_NOT_SUPPORTED: -32009,
+} as const;
```
ishymko: Can JSON-RPC codes be put in `json_rpc.ts`?

src/errors/base.ts
```diff
@@ -0,0 +1,403 @@
...
+  CREATED: 201,
+  ACCEPTED: 202,
+  NO_CONTENT: 204,
+  BAD_REQUEST: 400,
+  UNAUTHORIZED: 401,
+  NOT_FOUND: 404,
+  CONFLICT: 409,
+  INTERNAL_SERVER_ERROR: 500,
+  NOT_IMPLEMENTED: 501,
+} as const;
```
ishymko: Can HTTP statuses be in `rest.ts`?

src/errors/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * subclasses, plus the REST and JSON-RPC transport variants (both
+ * Workers-safe, no pb runtime dependency).
+ *
+ * gRPC errors live at `@a2a-js/sdk/errors/grpc` because their
+ * encode/decode helpers pull `@bufbuild/protobuf`.
+ */
+
+export * from './base.js';
+export * from './rest.js';
+export * from './json_rpc.js';
```
ishymko: Consider explicit exports name by name. Yeah that's a lot of boilerplate but makes it clear what's the contract of a certain entry point. Also it allows sharing some utility functions within it: i.e. putting a helper in `base.ts` and exporting it to consume from transport specific file will also export it with current `export *`.

src/errors/base.ts
```diff
@@ -0,0 +1,403 @@
...
+export interface A2AErrorSpec {
+  name: string;
+  reason: string;
+  code: number;
+  grpcStatus: number;
+  httpStatus: number;
+  defaultMessage: string;
+}
+
+const specs: A2AErrorSpec[] = [
```
ishymko, line 98: What I find confusing a bit here is that `A2AErrorSpec` sits in the `base.ts` but holds fields for all transports.

Can we somehow limit it to error names and descriptions (as in [this section](https://a2a-protocol.org/latest/specification/#326-service-parameters) of the spec which is a "core" transport agnostic section) and put transport specific details in the respective files?

src/errors/base.ts
```diff
@@ -0,0 +1,403 @@
...
+  TaskNotCancelableError,
+  PushNotificationNotSupportedError,
+  UnsupportedOperationError,
+  ContentTypeNotSupportedError,
+  InvalidAgentResponseError,
+  ExtendedAgentCardNotConfiguredError,
+  ExtensionSupportRequiredError,
+  VersionNotSupportedError,
+  RequestMalformedError,
+  GenericError,
```
ishymko: `GenericError` feels like an outlier here as it's not an actual spec defined error (unlike others). Is it possible to use `A2AError` directly? I see it's not abstract.

test/compat/v0_3/constants.spec.ts
```diff
@@ -29,7 +29,15 @@ import {
   isLegacyJsonRpcMethod,
   isV1JsonRpcMethod,
 } from '../../../src/compat/v0_3/constants.js';
-import { A2AError } from '../../../src/compat/v0_3/server/error.js';
+import { A2AError as _LegacyFacade } from '../../../src/compat/v0_3/server/error.js';
+import { A2AError, isJsonRpcError, UnsupportedOperationError } from '../../../src/errors/index.js';
+void _LegacyFacade;
```
ishymko: Why do we need `void _LegacyFacade;`?

.github/workflows/build-tests.yml
JakubWorek: All three optional peers stripped

package.json
JakubWorek: Updated

package.json
JakubWorek: Now using platform=node for grpc and express specific paths, keeping neutral only for non(grpc/express) paths

src/errors/base.ts
JakubWorek: done

src/errors/base.ts
JakubWorek: done

src/errors/base.ts
JakubWorek: Done, A2AErrorSpec now holds only {name, reason, defaultMessage}

src/errors/index.ts
JakubWorek: Done. Every public symbol from the non protobuf entrypoint is now listed by name.

src/errors/base.ts
JakubWorek: Updated, now src uses A2AError as base error class

test/compat/v0_3/constants.spec.ts
JakubWorek: we don't, now it is removed

.github/workflows/build-tests.yml
```diff
@@ -15,16 +25,26 @@ on:
...
-    - run: npm run build
-    - run: npm run test-build
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
+      - run: npm ci
+      - run: npm run build
```
bartek-gralewicz: nit: Could you restore indents so that we do not create unnecessary diff? Later on, it hides when lines were actually modified

src/errors/index.ts
ishymko, line 1: Let's limit public API to just errors (transport agnostic and transport specific).

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -310,43 +309,14 @@ export class GrpcTransport implements Transport {
...
-    return undefined;
-  }
-
   /**
-   * Maps a gRPC `ServiceError` to an SDK error class via
-   * `google.rpc.ErrorInfo` from `grpc-status-details-bin` metadata.
-   * Falls back to a generic `Error` carrying the gRPC code and details
+   * Maps a gRPC `ServiceError` to a semantic {@link import('../../../errors/index.js').GrpcA2AError}
+   * via `google.rpc.ErrorInfo` from `grpc-status-details-bin`. Falls
+   * back to `GrpcGenericError` preserving the gRPC status and details
```
ishymko: Seems like now it's `GrpcA2AErrorImpl`, also let's check other transports.

src/client/transports/grpc/grpc_transport.ts
```diff
@@ -310,43 +309,14 @@ export class GrpcTransport implements Transport {
...
    */
   private static mapToError(error: grpc.ServiceError, method?: keyof A2AServiceClient): Error {
-    const fromErrorInfo = GrpcTransport.mapFromErrorInfo(error);
-    if (fromErrorInfo) return fromErrorInfo;
-
-    const methodContext = method ? ' for ' + String(method) : '';
-    return new Error('gRPC error' + methodContext + ': ' + error.code + ' ' + error.details, {
-      cause: error,
-    });
+    return fromGrpcError(error, method ? String(method) : undefined);
```
ishymko: nit
```suggestion
    return fromGrpcError(error, method);
```

ishymko (approved):

src/client/transports/grpc/grpc_transport.ts
JakubWorek: removed the whole method in favour of direct calling fromGrpcError

PR #595, opened 2026-07-22
ci: point next dist-tag at stable releases too

JakubWorek:
# Description

New pointing logic:
```
If stable release:
    release under latest and next tag
else:
    release under next tag
```

Also bumps node version to `20` in the workflow

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -17,16 +17,16 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
           registry-url: 'https://registry.npmjs.org'
           cache: 'npm'
       - run: npm ci
       - run: npm run build
-      
+
       # Now configure with the publish service for install.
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -38,3 +38,11 @@ jobs:
           fi
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
bartek-gralewicz, line 42:
```suggestion
- name: Publish to NPM
run: |
         if [ "${{ github.event.release.prerelease }}" = "true" ]; then
           npm publish --provenance --access public --tag next
         else
           npm publish --provenance --access public
           VERSION=$(node -p "require('./package.json').version")
           npm dist-tag add "@a2a-js/sdk@${VERSION}" next
         fi
       env:
NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```

Could be done in a single step since we already have this `if` check and the required `env` token.

bartek-gralewicz (approved):

ishymko (approved):

PR #604, opened 2026-07-27
fix(compat/v0_3): call context builder for compatibility layer

### LINKED ISSUE #600: [Feat]: Call Context Builder also for Compatibility Layer

I was trying out A2A SDK v1.0 as it promised to solve some of the issue we still faced with the 0.3.0 version.
The V1 promises that one can develop with v.1 code but still supports v.0.3.0 message formats. The ContextBuilder breaks that idea as it's not executed when calling `a2a/rest/v1/message:send` with A2A-Version 0.3 while a task store extension still receives the new context property.

We had the hope, that we could mitigate our custom tenant logic and place it in the contextBuilder as we then would have access to the tenant in the our custom task store and custom push notification store. We need the tenant information to store the data into the database.

Is there any chance to support the contextBuilder also for the compatibility layer?

_No response_

### Additional context

My hope to the compatibility layer mainly was, that i can develop an Agent using the v1 code base entirely, but at the same time being ably to support the message format of v.0.3. So the stupid assumption i had was the compatibility layer transforms incoming messages to the v1 format - calls the v1 logic, and before sending a response the layer will transform the message back to the v0.3 format.

### Code of Conduct

JakubWorek:
# Description

The v0.3 compat REST router and gRPC service hard-coded `new
ServerCallContext(...)` and never invoked an operator-supplied `contextBuilder`. State that `contextBuilder` would inject (tenant, auth, request-id) was silently missing on v0.3 REST and gRPC requests.

JSON-RPC was already correct: the v1.0 `jsonRpcHandler` builds the context once and then dispatches to either the v1.0 or v0.3 transport handler based on the JSON-RPC method name, so an injected `contextBuilder` reached v0.3 traffic on that path. This change brings REST and gRPC to parity.

- Add `contextBuilder?: ServerCallContextBuilder` to `LegacyRestHandlerOptions` and `LegacyGrpcServiceOptions`.
- Route both compat handlers through `options.contextBuilder ?? defaultServerCallContextBuilder`.
- Pass `headers` through in both paths so the default builder's `context.state[STATE_HEADERS_KEY]` behavior isn't silently lost on v0.3 requests.
- Fix `LegacyGrpcServiceOptions` docstring that falsely claimed "Same shape as v1.0 `GrpcServiceOptions`".

The v1.0 Express `restHandler` auto-forwards its `contextBuilder` into the compat sub-router via structural typing (no v1.0-side change).
gRPC operators pass the same builder to `legacyGrpcService` next to `grpcService`.

Fixes #600


ishymko (approved):

bartek-gralewicz (approved):

RonaldMarske:
Awesome, thanks a lot :)

PR #618, opened 2026-08-03
fix(compat/v0_3/translate): add null and undefined checks to array and string fields

### LINKED ISSUE #616: [Bug]: Compat Client fails with a native 0.3 request wich does not provide optional fields
### What happened?

I'm currently exploring with A2A SDK v1.0.1 and try to replace our 0.3.0 implementation with 1.0.1.
Sadly the Client Compat Layer expects `referenceTaskIds` to be set even though the specification of 0.3.0 mark this field as optional.

Consequence, the app crashes with a native JS error message.

The relevant code parts:

```
function toCompatMessage(coreMsg) {
  // ... 
  const referenceTaskIds = nonEmptyArray(coreMsg.referenceTaskIds);
}

function nonEmptyArray(value) {
  return value.length === 0 ? void 0 : [...value];
}
```

Makes sense as referenceTaskIds are optional, meaning undefined or null `.length` fails.

FYI: I discovered also the same pattern for AgentCards, if the card does not have the correct format, the app just crashes accessing properties which are not there.

Maybe it's nice to add a safetynet with proper validation messages to consumers, to prevent users from digging in the code to find out where the problem comes from.

### Relevant log output

```shell
The Error Message:

TypeError: Cannot read properties of undefined (reading 'length')
    at nonEmptyArray (node_modules/@a2a-js/sdk/dist/client/index.cjs:2399:16)
    at toCompatMessage (node_modules/@a2a-js/sdk/dist/client/index.cjs:2440:28)
    at toCompatSendMessageRequest (node_modules/@a2a-js/sdk/dist/client/index.cjs:2563:14)
    at toCompatSendStreamingMessageRequest (node_modules/@a2a-js/sdk/dist/client/index.cjs:2573:17)
    at _LegacyJsonRpcTransport.sendMessageStream (node_modules/@a2a-js/sdk/dist/client/index.cjs:2701:22)
    at sendMessageStream.next (<anonymous>)
    at Client.sendMessageStream (node_modules/@a2a-js/sdk/dist/client/index.cjs:5437:22)
```

### Code of Conduct

JakubWorek:
# Description

The v0.3 compat translators dereferenced V1/legacy array/object fields without any guard, so a partial object crashed with TypeError: Cannot read properties of undefined (reading 'length') inside nonEmptyArray.

Two-part fix:
1. Optional v0.3 fields (referenceTaskIds, extensions, history, artifacts, signatures, securitySchemes, securityRequirements, examples/inputModes/outputModes, acceptedOutputModes) — silently coerced to []/undefined when missing. nonEmptyArray/nonEmptyString helpers now accept T[] | undefined | null / string | undefined | null.
2. Required v0.3 fields (parts, tags, capabilities, defaultInputModes, defaultOutputModes, skills) — new requireArray / requireObject helpers in _validate.ts throw A2AError.invalidParams with the field path (e.g. message.parts is required and must be an array) instead of a native TypeError. Existing strict validation in toCoreMessage preserved.

Fixes #616 🦕

### DIFF
--- a/src/compat/v0_3/translate/_validate.ts
+++ b/src/compat/v0_3/translate/_validate.ts
@@ -1,0 +1,15 @@
+import { A2AError } from '../server/error.js';
+
+export function requireArray<T>(value: T[] | undefined | null, path: string): T[] {
+  if (!Array.isArray(value)) {
+    throw A2AError.invalidParams(`${path} is required and must be an array`);
+  }
+  return value;
+}
+
+export function requireObject<T>(value: T | undefined | null, path: string): T {
+  if (value == null || typeof value !== 'object') {
+    throw A2AError.invalidParams(`${path} is required`);
+  }
+  return value;
+}
--- a/src/compat/v0_3/translate/agent_card.ts
+++ b/src/compat/v0_3/translate/agent_card.ts
@@ -101,7 +101,7 @@ export function toCompatAgentCapabilities(core: V1AgentCapabilities): legacy.Age
   const result: legacy.AgentCapabilities1 = {};
   if (core.streaming !== undefined) result.streaming = core.streaming;
   if (core.pushNotifications !== undefined) result.pushNotifications = core.pushNotifications;
-  if (core.extensions.length > 0) {
+  if (core.extensions && core.extensions.length > 0) {
     result.extensions = core.extensions.map(toCompatAgentExtension);
   }
   return result;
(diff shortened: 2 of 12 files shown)


bartek-gralewicz (commented):
I believe that we should reassess if this approach is the right one for this issue. The issue author mentions that our SDK crashed when an optional field was missing. This is a true bug and should be addressed but the changes in this PR, make required fields optional which should not be the case. Now, for missing required fields, there is no error, no warning, just silent accept with default values.

There is even a test `'does not crash when every declared-required field is missing at once'`. This IMO is something we should revert and only fix handling missing optional fields.

Another comment from the issue author was that it would be nice to get some validation error when a required field is missing. If we want to touch required fields in this PR, I would suggest adding a helper method that takes the field or throws a user-friendly informative message saying which required field was missing. This could be a follow up as it doesn't fix the issue directly and is just a "nice to have". I am leaving the separation of these functionalities to your judgement.

bartek-gralewicz (approved):

PR #621, opened 2026-08-04
test: add public API change detection

### LINKED ISSUE #178: [Feat]: Research options for build-time breaking changes detection
Similar to https://github.com/VerifyTests/Verify from .NET, it would be beneficial to highlight breaking changes introduced in exported types (function signatures, interfaces, etc.)

Given that Typescript type system is Turing Complete, it may be possible only to some extent, however anything will help.

https://api-extractor.com/ looks promising.

Coooder-Crypto:
## Summary

- add API Extractor-based reports for every typed entry in `package.json#exports`
- check in API baselines for all 14 public package entry points
- fail CI on API drift, missing reports, or stale reports left by removed entry points
- document how contributors update API reports for intentional changes

## Why

Exported TypeScript signatures are part of the SDK's compatibility contract, but they were not previously reviewed as a single build artifact. The checked-in reports make API changes visible in pull request diffs and prevent accidental changes from passing CI unnoticed.

## Developer impact

The build workflow now runs `npm run api-report` after generating declarations. Contributors who intentionally change the public API can run:

```sh
npm run build
npm run api-report:update
```

The report runner discovers package entry points automatically, so newly added typed exports require a baseline and removed exports remove their stale baseline during a local update.

## Validation

- `npm run build`
- `npm run api-report`
- `npm run lint:ci`
- `npm run test-build`
- `npm test` — 67 test files and 1,438 tests passed
- verified that an added public declaration fails the API check
- verified that a stale report fails CI and is removed by the update command

Closes #178


.github/workflows/build-tests.yml
bartek-gralewicz, line 1: This workflow has a `- '**.md'` disable rule. It would be possible to modify `etc/api/**.md` files without triggering this workflow what would result in a broken repo state. Adding `etc/api` directory to a separate trigger list would help but I believe that having an explicit, separate workflow for this is a cleaner approach. `build-test` doesn't represent the nature of this updated test scope this workflow now delivers.

WDYT about `.github/workflows/api-report.yml`?

scripts/apiReport.js
ishymko, line 1: > verified that an added public declaration fails the API check

Would be super helpful to attach artifacts of e2e testing (e.g. from running on fork) which demonstrate the workflow, e.g. a PR which fails the check and is red, a PR which is supposed to add public API so has md's updated.

PR #629, opened 2026-08-10
fix: reject empty task IDs with RequestMalformedError in getTask/cancelTask

ez-lbz:
### 1. Reject empty/whitespace task IDs with `RequestMalformedError` in `getTask`/`cancelTask`

**Problem:** `getTask` and `cancelTask` passed `params.id` straight to `taskStore.load`. An empty or whitespace-only task ID therefore surfaced as `TaskNotFoundError` (-32001 / HTTP 404) instead of the protocol-correct `RequestMalformedError` (-32602 / HTTP 400). Python and Go return `InvalidParams` for this input; JS diverged.

- `getTask` now rejects empty/whitespace-only `params.id` with `RequestMalformedError('Task ID is required')` before touching the store.
- `cancelTask` applies the same validation.

## Testing

- `npx tsc --noEmit -p tsconfig.test.json` — passes.
- New regression tests: `getTask` with `id: ''` and `id: ' '` reject with `RequestMalformedError`; `cancelTask` with `id: ''` and `id: ' \t '` reject with `RequestMalformedError`.

Behavior change: empty/whitespace task IDs on `tasks/get` and `tasks/cancel` now return -32602/400 instead of -32001/404. All other inputs behave as before.

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -713,6 +713,9 @@ export class DefaultRequestHandler implements A2ARequestHandler {
 
   async getTask(params: GetTaskRequest, context: ServerCallContext): Promise<Task> {
     const taskId = params.id;
+    if (!taskId || taskId.trim() === '') {
+      throw new RequestMalformedError('Task ID is required');
+    }
     const task = await this.taskStore.load(taskId, context);
     if (!task) {
       throw new TaskNotFoundError(`Task not found: ${params.id}`);
@@ -744,6 +747,9 @@ export class DefaultRequestHandler implements A2ARequestHandler {
 
   async cancelTask(params: CancelTaskRequest, context: ServerCallContext): Promise<Task> {
     const taskId = params.id;
+    if (!taskId || taskId.trim() === '') {
+      throw new RequestMalformedError('Task ID is required');
+    }
     const task = await this.taskStore.load(taskId, context);
     if (!task) {
       throw new TaskNotFoundError(`Task not found: ${params.id}`);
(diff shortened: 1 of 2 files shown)


JakubWorek:

ez-lbz:
Hi @JakubWorek, thanks for the heads-up — the branch was based on a chain that included the #625 commits. I've rebased it onto the current `main` and removed the unrelated commits; it now contains only the empty-task-ID validation change (plus the prettier/tracking-id cleanup). All 1494 tests pass locally.

src/server/request_handler/default_request_handler.ts
JakubWorek, line 942: Thank you for the contribution, wdyt about covering also other methods: `resubscribe` and four `*TaskPushNotificationConfig` ?

ez-lbz:
Thanks for the suggestion @JakubWorek — done. `resubscribe` and all four `*TaskPushNotificationConfig` methods (`create`/`get`/`list`/`delete`) are now covered. The whitespace/empty guard is deduplicated into a shared `_requireValidTaskId` helper used by every task-ID entry point of `DefaultRequestHandler`, so they all surface `RequestMalformedError` instead of `TaskNotFoundError`.

While extending coverage I also fixed a pre-existing test ("Push Notification methods should throw error if task does not exist") that was still passing legacy `name:`-path-style params, i.e. no real `taskId` field at all. It now uses the current request shapes against a genuinely non-existent task id and keeps asserting `TaskNotFoundError`.

All checks are green on the updated head.

JakubWorek (approved):

PR #630, opened 2026-08-10
fix: return 404/-32001 for missing push notification configs

ez-lbz:
### 1. Map missing push-notification config to 404 / -32001

**Problem:** `getTaskPushNotificationConfig` threw a bare `A2AError` when no config existed for the task (or the requested config id was unknown). A bare `A2AError` has no transport mapping, so REST returned 500 (`INTERNAL_ERROR` / `-32603` on JSON-RPC) instead of the semantic "not found" response. Python returns 404 for this case; Go had the same 500 bug.

- Both not-found branches in `getTaskPushNotificationConfig` now throw `TaskNotFoundError`, which maps to HTTP 404 on REST and `-32001` on JSON-RPC (the `TaskNotFound`-class code the plan targets), matching Python's behavior.

## Testing

- `npm test` — 68 test files, 1508 tests passed.
- `npx tsc --noEmit` — no errors in `src/`; `npx tsc --noEmit -p tsconfig.test.json` passes.
- New regression tests: `getTaskPushNotificationConfig` with no stored configs, and with an unknown config id, both reject with `TaskNotFoundError`. The existing REST test (`GET /tasks/:taskId/pushNotificationConfigs/:configId` → 404 with `TASK_NOT_FOUND` detail) now exercises the real handler path.

Behavior change: missing push-notification configs now return 404 (REST) / `-32001` (JSON-RPC) instead of 500 / `-32603`.


JakubWorek:

ez-lbz:
Hi @JakubWorek, thanks for the heads-up — the branch was based on a chain that included the #625 commits. I've rebased it onto the current `main` and removed the unrelated commits; it now contains only the missing-push-config 404 change (plus the prettier/tracking-id cleanup). All 1492 tests pass locally.

JakubWorek (approved):
LGTM, changes are according to spec: https://a2a-protocol.org/v1.0.0/specification/#318-get-push-notification-config

PR #631, opened 2026-08-10
fix: reject invalid ListTasks status filters with RequestMalformedError

ez-lbz:
### 1. Reject invalid `status` filters on `ListTasks` instead of silently converting them to `UNRECOGNIZED`

**Problem:** The generated `taskStateFromJSON` (from `a2a.proto`) maps any unknown status string or out-of-range number to `UNRECOGNIZED` (-1). The REST transport used it directly, so `GET /tasks?status=INVALID_VALUE` silently applied an `UNRECOGNIZED` filter that matches nothing and returned an empty list — the client believed the filter worked. Python's `ParseDict` rejects invalid values; JS diverged. (The pb file is generated via `buf generate` from the upstream A2A proto, so the fix lives in the transport parsing, not in the generated file.)

**Fix (src/server/transports/rest/rest_transport_handler.ts):**
- `listTasks` now throws `RequestMalformedError('Invalid status filter')` when `taskStateFromJSON` yields `UNRECOGNIZED`, returning 400 `INVALID_ARGUMENT` instead of silently filtering.

**Fix (src/server/transports/jsonrpc/jsonrpc_transport_handler.ts):**
- The `ListTasks` case validates the parsed `ListTasksRequest.status` and rejects `UNRECOGNIZED` with `RequestMalformedError` (-32602) before dispatching.

## Testing

- New regression tests: REST `listTasks` with `status='bogus-status'` and `status='99'` reject with `Invalid status filter`; a valid `TASK_STATE_COMPLETED` still passes through; JSON-RPC `ListTasks` with `status='bogus-status'` returns `INVALID_PARAMS` (-32602) and does not reach the handler.

Behavior change: an invalid `status` on `ListTasks` now returns 400 (REST) / `-32602` (JSON-RPC) instead of 200 with an empty result set. Existing test `GET /tasks > should treat unrecognized status values as UNRECOGNIZED (-1)` was updated to assert the new 400 contract.

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -737,6 +737,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
 
   async getTask(params: GetTaskRequest, context: ServerCallContext): Promise<Task> {
     const taskId = params.id;
+    this._requireValidTaskId(taskId);
     const task = await this.taskStore.load(taskId, context);
     if (!task) {
       throw new TaskNotFoundError(`Task not found: ${params.id}`);
@@ -776,6 +777,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
 
   async cancelTask(params: CancelTaskRequest, context: ServerCallContext): Promise<Task> {
     const taskId = params.id;
+    this._requireValidTaskId(taskId);
     const task = await this.taskStore.load(taskId, context);
     if (!task) {
       throw new TaskNotFoundError(`Task not found: ${params.id}`);
@@ -860,6 +862,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
       throw new PushNotificationNotSupportedError();
     }
     const taskId = params.taskId;
+    this._requireValidTaskId(taskId);
     const task = await this.taskStore.load(taskId, context);
     if (!task) {
       throw new TaskNotFoundError(`Task not found: ${taskId}`);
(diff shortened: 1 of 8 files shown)


JakubWorek:

ez-lbz:
Hi @JakubWorek, thanks for the heads-up — the branch was based on a chain that included the #625 commits. I've rebased it onto the current `main` and removed the unrelated commits; it now contains only the invalid-status-filter rejection change (plus the prettier/tracking-id cleanup). All 1495 tests pass locally.

JakubWorek (commented):
Overall changes are correct, however I'd move them to request handler. Currently grpc listTasks method is not covered by them and this violates [Functional Equivalence Requirement](https://a2a-protocol.org/v1.0.0/specification/#51-functional-equivalence-requirements).

ez-lbz:
Thanks for catching this @JakubWorek — you're right, validating at the REST layer left `grpc` `listTasks` uncovered and broke the §5.1 functional-equivalence guarantee. I've moved the validation into `DefaultRequestHandler.listTasks` (next to the existing `pageSize`/`statusTimestampAfter` checks): any state outside the named `TaskState` values raises `RequestMalformedError` there, so JSON-RPC, REST, **and** gRPC now reject invalid filters identically — gRPC matters most since it decodes `status` as a raw enum number without going through `taskStateFromJSON`.

A side effect worth calling out: an unknown status previously deserialized to protobufjs' synthetic `UNRECOGNIZED` (-1) sentinel and silently matched nothing, on every transport. That silent-empty behavior is now a malformed request (`-32602` / HTTP 400) everywhere.

The transport handlers are back to plain parse-and-forward; their specs now assert pass-through of the parsed value plus correct mapping of a handler-raised `RequestMalformedError`, and handler-level specs cover both unrecognized strings and out-of-range numbers. All checks green on the updated head.

JakubWorek (approved):

PR #632, opened 2026-08-10
fix: reject concurrent sendMessage while the task executor is running

ez-lbz:
### 1. Reject a second `sendMessage` while the task's executor is still running

**Problem:** `sendMessage`/`sendMessageStream` only checked for terminal states before starting an executor. A second `message/send` on a task still in `WORKING` started a *second* executor for the same task — the two ran concurrently and raced on the shared task state. Go has `ErrExecutionInProgress` and Python V2 has an `ActiveTaskRegistry`; JS had neither.

- Added an in-flight registry (`executingTaskIds: Set<string>`) whose lifetime is tied to the *executor* (added before the executor starts, removed in the executor's `.finally` when it settles) — not to the event bus, so an `INPUT_REQUIRED` / `AUTH_REQUIRED` pause (executor settled, bus kept alive for follow-ups) still allows the next turn.
- `sendMessage` and `sendMessageStream` check the registry twice: once synchronously at the top (before the first `await`, closing the race window between check and executor start) and once after `_createRequestContext`. A conflicting send throws `UnsupportedOperationError` with a clear message.
- The streaming path deliberately does **not** clear the mark when the consumer disconnects early — only the executor's `.finally` clears it, so an early client disconnect cannot unlock a second executor.

## Testing

- `npm test` — 68 test files, 1507 tests passed.
- New regression test: with a gated executor, a second `sendMessage` on the same WORKING task rejects with `UnsupportedOperationError` and does not start a second executor (`execute` called once); after the first executor settles at `INPUT_REQUIRED`, a follow-up send succeeds (`execute` called twice).

Behavior change: `message/send` / `message/stream` on a task whose executor is still running now returns `-32004` (REST 400) instead of starting a second executor. Multi-turn `INPUT_REQUIRED` / `AUTH_REQUIRED` flows are unaffected.

### DIFF
--- a/itk/itk_agent.ts
+++ b/itk/itk_agent.ts
@@ -17,6 +17,8 @@ import {
   RequestContext,
   ExecutionEventBus,
   AgentEvent,
+  InMemoryPushNotificationStore,
+  DefaultPushNotificationSender,
 } from '../src/server/index.js';
 import {
   agentCardHandler,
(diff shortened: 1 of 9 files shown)


JakubWorek:
Hi @ez-lbz, looks like this PR contains commit not related to the description, taken from https://github.com/a2aproject/a2a-js/pull/625. Could you please update the branch to contain only relevant changes and resolve merge conflicts?

ez-lbz:

Thanks again!

PR #633, opened 2026-08-10
fix: handle gRPC client cancellation on streaming calls

ez-lbz:
### 1. Handle gRPC client cancellation on streaming calls

**Problem:** `wrapStreaming` (used by `sendStreamingMessage` and `subscribeToTask`) had no `'cancelled'` event listener on the gRPC call. When a client disconnected mid-stream, the server kept draining the async generator and holding its resources (execution event queue, store reads, executor bus) even though nobody consumed the writes, and the write loop never ended promptly.

**Fix (src/server/grpc/grpc_service.ts):**
- `wrapStreaming` now registers a `'cancelled'` listener on the call.
- The write loop races `iterator.next` against a cancellation signal, so it always terminates when the client disconnects.
- On cancellation it also calls `iterator.return` best-effort, which closes the generator so its `finally` blocks (e.g. `ExecutionEventQueue.stop`) run — with a note that a generator parked on an await that never resolves cannot be closed, but the loop still exits and `call.end` still runs.
- The listener is removed in a `finally` when the stream ends normally or fails.

## Testing

- `npm test` — 68 test files, 1507 tests passed.
- New regression test: with a gated generator, firing `'cancelled'` mid-stream stops further writes (only the first event is written), closes the generator (its `finally` runs), ends the call, and does not emit an error.

Behavior change: none for uncancelled streams — the existing streaming tests (multi-write + error paths) still pass unchanged.

### DIFF
--- a/itk/itk_agent.ts
+++ b/itk/itk_agent.ts
@@ -17,6 +17,8 @@ import {
   RequestContext,
   ExecutionEventBus,
   AgentEvent,
+  InMemoryPushNotificationStore,
+  DefaultPushNotificationSender,
 } from '../src/server/index.js';
 import {
   agentCardHandler,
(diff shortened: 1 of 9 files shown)


JakubWorek:

ez-lbz:

Thanks again!

PR #634, opened 2026-08-10
fix: reject empty/blank and CRLF push auth credentials

ez-lbz:
### 1. Fail loudly on empty push-notification auth credentials

**Problem:** `_buildAuthHeaders` used a truthiness check (`scheme && credentials`), so a configured auth `scheme` with an empty/whitespace `credentials` value silently omitted the `Authorization` header. The webhook then received an unauthenticated push and returned 401/403, with no server-side signal about the misconfigured credential.

**Fix (src/server/push_notification/default_push_notification_sender.ts):**
- When `authentication.scheme` is set, empty or blank `credentials` now throw an error (logged by the sender's dispatch error handler) instead of silently dropping the header.

### 2. Reject CR/LF in push-notification auth credentials

**Problem:** The `Authorization` header was built via `scheme + " " + credentials` with no sanitization. Credentials containing CR/LF characters could inject additional HTTP headers into the outbound push request (header injection, CWE-113) — JDK clients validate by default, but Node's `fetch`/undici and custom HTTP clients may not.

**Fix (src/server/push_notification/default_push_notification_sender.ts):**
- Credentials containing `\r` or `\n` are rejected before the header is constructed.

## Testing

- `npm test` — 68 test files, 1509 tests passed.
- `npx tsc --noEmit -p tsconfig.test.json` passes.
- New regression tests (integration-level, against a real HTTP server): a `Bearer` scheme with `credentials: ''` and `credentials: ' '` produce no outbound request and log an error containing `scheme 'Bearer' requires non-empty credentials`; credentials `'token\r\nX-Injected: 1'` produce no outbound request and log `must not contain CR/LF characters`. The existing happy-path auth test (`Authorization: Bearer token-xyz`) still passes unchanged.

Behavior change: misconfigured auth (empty/blank credentials, or CR/LF in credentials) now errors instead of silently sending an unauthenticated (or header-injecting) push.

### DIFF
--- a/itk/itk_agent.ts
+++ b/itk/itk_agent.ts
@@ -17,6 +17,8 @@ import {
   RequestContext,
   ExecutionEventBus,
   AgentEvent,
+  InMemoryPushNotificationStore,
+  DefaultPushNotificationSender,
 } from '../src/server/index.js';
 import {
   agentCardHandler,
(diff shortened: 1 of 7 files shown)


JakubWorek:

ez-lbz:

Thanks again!

PR #636, opened 2026-08-10
fix: guard terminal state transitions and make cancelTask atomic

ez-lbz:
### 1. Terminal states can no longer be overwritten

**Problem:** `ResultManager.applyStatusUpdate` assigned `task.status = structuredClone(updateEvent.status)` unconditionally. A late status update could roll a `COMPLETED`/`FAILED`/`CANCELED`/`REJECTED` task back into a non-terminal state (or a different terminal state) — the same blind-overwrite gap as Python's `task_manager.py:203-211`.

**Fix (src/server/result_manager.ts):**
- `applyStatusUpdate` now checks the persisted state: if it is terminal and the incoming update has a *different* state, the update is ignored (with a warning). Same-state terminal updates (e.g. a final `COMPLETED` message arriving after the `COMPLETED` status) are still applied.

### 2. Make `cancelTask`'s check-then-act atomic

**Problem:** `cancelTask` loaded the task, checked its state, set `CANCELED`, and saved — with no lock between check and save. A `COMPLETED` status written by a racing executor in that window could be clobbered by the `CANCELED` write (or the cancel would win and corrupt the completed task), a TOCTOU race shared by most SDKs.

- The no-bus cancel path now routes the `CANCELED` status update through a fresh `ResultManager.processEvent(...)` instead of mutating the loaded task and saving directly. That applies the per-(tenant, owner, taskId) write lock and the new terminal-state guard: if the executor already completed the task, the `CANCELED` update is rejected and `cancelTask` surfaces `TaskNotCancelableError` instead of corrupting the task.

## Testing

- New regression tests:
  - `ResultManager`: a `WORKING` status update on a `COMPLETED` task is ignored; a `CANCELED` update on a `COMPLETED` task is ignored; a same-state `COMPLETED` update with a final message is still applied.
  - `cancelTask` vs a racing executor that completes the task between the initial load and the locked write: `TaskNotCancelableError` is thrown and no `CANCELED` write reaches the store.
- Test updates required by the behavior change: two mock stores (`default_request_handler.spec.ts`, `auth_required.spec.ts`) returned live references from `load`; they now deep-clone like the real `InMemoryTaskStore`, so the guard sees the true persisted state. Their assertions (FAILED persisted after a COMPLETED save failure) are unchanged.

Behavior change: status updates that would move a terminal task to a different state are dropped; canceling a task that a racing executor just completed fails with `TaskNotCancelableError` instead of overwriting `COMPLETED` with `CANCELED`.

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -29,8 +29,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-node@v4
+    - uses: actions/checkout@v7
+    - uses: actions/setup-node@v7
       with:
         node-version: 20
         registry-url: 'https://registry.npmjs.org'
--- a/.github/workflows/coverage-comment.yml
+++ b/.github/workflows/coverage-comment.yml
@@ -18,7 +18,7 @@ jobs:
       github.event.workflow_run.conclusion == 'success'
     steps:
       - name: Download Coverage Artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v7
         with:
           run-id: ${{ github.event.workflow_run.id }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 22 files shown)


JakubWorek:
Hi @ez-lbz, looks like this PR contains commit not related to the description, taken from https://github.com/a2aproject/a2a-js/pull/625. Could you please update the branch to contain only relevant changes and resolve merge conflicts?

ez-lbz:

Thanks again!

JakubWorek (approved):

PR #638, opened 2026-08-10
fix: harden SSE streaming (keep-alive, event ids, idle timeout) and cap request body size

ez-lbz:
### 1. SSE keep-alive heartbeats + monotonic event IDs (+ optional idle timeout)

**Problem:** The REST and JSON-RPC SSE write loops emitted bare `data:` frames with no keep-alive traffic and no event IDs. Proxies/load balancers can terminate idle SSE connections, and clients cannot implement resumable subscriptions (`Last-Event-ID`) without IDs. No SDK had heartbeats or IDs consistently.

**Fix (src/sse_utils.ts):**
- Added `writeSseStream(res, frames, options)`, a shared server-side SSE writer that:
  - prefixes every data frame with a monotonically increasing `id:` field (counter),
  - emits `: keep-alive` comment frames while the stream is idle (`heartbeatIntervalMs`, default 15s), which clients ignore per the SSE spec,
  - supports an optional inactivity timeout (`idleTimeoutMs`, default 0 = disabled) that stops the stream, bounding how long a half-dead client holds server resources.
- The helper never ends the response itself, so callers can still emit an SSE error frame on stream failure.

**Fix (src/server/express/rest_handler.ts):**
- `sendStreamResponse` now routes its write loop through `writeSseStream`.
- `RestHandlerOptions` gains `sseIdleTimeoutMs` (default 0 = disabled).

**Fix (src/server/express/json_rpc_handler.ts):**
- Both SSE loops (the v0.3 always-stream path and the v1.0 peek-then-flush path) now route through `writeSseStream`.
- `JsonRpcHandlerOptions` gains `sseIdleTimeoutMs` (default 0 = disabled).

### 2. Explicit request body size limit

**Problem:** `express.json` has no default body-size limit; a client could stream an arbitrarily large JSON payload into server memory (CWE-400).

**Fix (src/server/express/rest_handler.ts, src/server/express/json_rpc_handler.ts):**
- Both `express.json` calls now set `limit: '10mb'` (large A2A files should be referenced via `FileWithUri` parts, so 10 MB is generous). Oversized bodies surface as 413 via body-parser and are handled by the existing error middleware.

## Testing

- `npm test` — 68 test files, 1509 tests passed.
- New regression tests (test/sse_utils.spec.ts): `writeSseStream` prefixes frames with `id: 1`, `id: 2`, …; emits `: keep-alive` comment frames while idle (fake timers); stops when no event arrives within `idleTimeoutMs`.
- Existing REST/JSON-RPC SSE tests still pass unchanged (`data:` payload assertions are unaffected by the added `id:` lines).

Behavior change: SSE streams now carry `id:` fields and periodic keep-alive comment frames (clients ignore both per spec); request bodies over 10 MB are rejected with 413. `sseIdleTimeoutMs` defaults to 0, so existing long-running subscriptions are unaffected unless the option is set.

### DIFF
--- a/itk/itk_agent.ts
+++ b/itk/itk_agent.ts
@@ -17,6 +17,8 @@ import {
   RequestContext,
   ExecutionEventBus,
   AgentEvent,
+  InMemoryPushNotificationStore,
+  DefaultPushNotificationSender,
 } from '../src/server/index.js';
 import {
   agentCardHandler,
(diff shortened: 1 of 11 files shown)


JakubWorek:

ez-lbz:

Thanks again!

Issue #643, opened 2026-08-10
[Bug]: multiple content oneof arms are silently accepted instead of rejected


ishan-1010:
### What happened?

More than one member of a oneof being present is malformed input, and proto3 JSON treats
it as an error. `a2a-python` does:

```python
from a2a.types import Part
from google.protobuf.json_format import ParseDict

ParseDict({"text": "hello", "url": "https://example.com/x"}, Part())
# ParseError: Message type "lf.a2a.v1.Part" should not have multiple "content" oneof fields
```

This SDK accepts it and picks one:

```js
import { Part } from '@a2a-js/sdk';

Part.toJSON(Part.fromJSON({ text: 'hello', url: 'https://example.com/x' }));
// { text: 'hello' }     url is gone, no error
```

### Why it matters

Which arm survives is decided by the order fields are assigned in the generated
`fromJSON` body, not by anything in the input. So a malformed message from a buggy peer
gets silently reinterpreted rather than refused, and the two SDKs can end up disagreeing
about what the sender meant. Silently is the part that bothers me: whoever sent it gets
no signal that half their Part was discarded.

### What I'd expect

`fromJSON` throws when more than one member of `content` is present, matching what
`a2a-python` and the proto3 JSON mapping do.

### How this was found

Round-tripping a shared corpus through this SDK and `a2a-python` and diffing the JSON.

Reproduced on `@a2a-js/sdk` 1.0.1 from npm and on `main` at `1c6eb32`, against
`a2a-python` at `cff6727`.

Generated code again, so let me know where you'd want the fix.

Varun-S10:
Hi @ishan-1010, Thank you for raising the issue. We have reproduced the issue and raised a PR to fix it. The PR will be reviewed. Thank you for your patience and understanding.

Issue #642, opened 2026-08-10
[Bug]: Part.data set to null drops the content oneof


ishan-1010:
### What happened?

`Part` has a oneof named `content` over `text`, `raw`, `url` and `data`. `data` is a
`google.protobuf.Value`, and `null` is a perfectly good `Value`.

When `data` is null, this SDK parses the Part as having no content arm at all:

```js
import { Part } from '@a2a-js/sdk';

const p = Part.fromJSON({ data: null, mediaType: 'application/json' });
// { filename: '', mediaType: 'application/json' }   no content case

Part.toJSON(p);
// { mediaType: 'application/json' }
```

`a2a-python` treats the arm as set and round-trips it:

```python
from a2a.types import Part
from google.protobuf.json_format import MessageToDict, ParseDict

m = Part(); ParseDict({"data": None, "mediaType": "application/json"}, m)
m.HasField('data')   # True
MessageToDict(m)     # {'data': None, 'mediaType': 'application/json'}
```

### Why it matters

"This is a data part whose payload is null" and "this part has no content" are different
statements, and the second one isn't valid. A data part sent by a Python agent arrives at
a JS peer as a Part with nothing in it, and gets forwarded on in that state.

Other `Value` shapes are fine. Objects, arrays, strings, numbers, booleans and empty
containers all round-trip identically between the two SDKs. Null is the only one that
falls through.

### What I'd expect

`{"data": null}` keeps the `data` arm set, and `toJSON` emits `data: null` rather than
omitting the key.

### How this was found

Round-tripping a shared corpus through this SDK and `a2a-python` and diffing the JSON.

Reproduced on `@a2a-js/sdk` 1.0.1 from npm and on `main` at `1c6eb32`, against
`a2a-python` at `cff6727`.

This is in the generated `src/types/pb/a2a.ts`, so happy to send the fix wherever you'd
prefer it to live.

Varun-S10:
Hi @ishan-1010,  Thank you for raising the issue. We have reproduced the issue and raised a PR to fix it. The PR will be reviewed. Thank you for your patience and understanding.

Issue #641, opened 2026-08-10
[Bug]: Timestamp fields are unvalidated strings and are never normalized to UTC


ishan-1010:
### What happened?

`google.protobuf.Timestamp` fields are passed straight through as strings. The generated
code is:

```ts
timestamp: isSet(object.timestamp) ? globalThis.String(object.timestamp) : undefined,
```

and `toJSON` copies it back out unchanged. Nothing parses it and nothing checks it, which
causes two separate problems.

**Malformed timestamps are accepted and forwarded.** Every one of these is stored and
re-emitted verbatim:

```js
import { TaskStatus } from '@a2a-js/sdk';
const ts = (t) => TaskStatus.toJSON(TaskStatus.fromJSON({ state: 'TASK_STATE_WORKING', timestamp: t }));

ts('not-a-timestamp');  // { state: 'TASK_STATE_WORKING', timestamp: 'not-a-timestamp' }
ts('');                 // { ..., timestamp: '' }
ts(12345);              // { ..., timestamp: '12345' }   a JSON number becomes a string
```

`a2a-python` rejects all of them. So a JS server accepts a `Task` whose
`status.timestamp` is arbitrary text, hands it downstream, and the Python peer is the one
that throws. The component that let the bad data in isn't the component that reports it,
which makes this annoying to track down in a mixed deployment.

**Offsets are never normalized.** proto3 JSON output is always Z-normalized.
`a2a-python` does that; this SDK keeps whatever offset arrived:

```
input   "2026-01-01T05:30:00+05:30"
python  "2026-01-01T00:00:00Z"
js      "2026-01-01T05:30:00+05:30"
```

Same instant, different bytes. That's fine until something compares bytes rather than
instants, which is what canonicalization does. `src/signature.ts:188` canonicalizes via
`AgentCard.toJSON(AgentCard.fromJSON(card))`, so this codec sits on the signing path.

Storing the field unparsed also means any timestamp comparison or ordering inside the SDK
is operating on a string whose format isn't guaranteed.

### What I'd expect

`fromJSON` parses and validates RFC 3339, rejecting input that isn't a timestamp, and
`toJSON` emits the Z-normalized form.

### How this was found

Round-tripping a shared corpus through this SDK and `a2a-python` and diffing the JSON.
Both generate from the same `.proto`.

Reproduced on `@a2a-js/sdk` 1.0.1 from npm and on `main` at `1c6eb32`, against
`a2a-python` at `cff6727`.

[...]

Issue #640, opened 2026-08-10
[Bug]: unknown enum values become the string "UNRECOGNIZED", which a2a-python rejects


ishan-1010:
### What happened?

`ts-proto` maps any enum value it doesn't recognise to the sentinel `UNRECOGNIZED`, and
`toJSON` emits that as a literal string. proto3 JSON keeps an unknown enum value as its
integer instead, so that a peer running an older version can still pass a newer value
through untouched.

Right now the value isn't just degraded, it's gone. `"UNRECOGNIZED"` carries nothing about
what the original was, so nothing downstream can recover it.

```js
import { TaskStatus } from '@a2a-js/sdk';

TaskStatus.toJSON(TaskStatus.fromJSON({ state: 99 }));
// { state: 'UNRECOGNIZED' }
```

`a2a-python` on the same input:

```python
from a2a.types import TaskStatus
from google.protobuf.json_format import MessageToDict, ParseDict

m = TaskStatus(); ParseDict({"state": 99}, m)
MessageToDict(m)                       # {'state': 99}

ParseDict({"state": "UNRECOGNIZED"}, TaskStatus())
# ParseError: Invalid enum value UNRECOGNIZED for enum type lf.a2a.v1.TaskState
```

So the round trip through a JS hop is lossy in a way the other end can't parse:

```
python emits    {"state": 99}
js re-emits     {"state": "UNRECOGNIZED"}
python rejects  ParseError
```

### Why it matters

`TaskState` currently defines 0 through 8, so 9 is the next value the spec would add. On
that day, any 1.0 JS SDK sitting between two newer peers as a proxy, relay or gateway
turns a forward-compatible number into a string that a conformant parser refuses. The
same applies to `Role`.

It's latent today and guaranteed to bite on the next enum addition, which is the awkward
kind: nothing is visibly wrong until a version bump makes it wrong everywhere at once.

### What I'd expect

Unrecognised enum members keep their numeric value through `fromJSON` and `toJSON` rather
than collapsing to a sentinel.

### How this was found

Round-tripping a shared corpus of protocol objects through this SDK and `a2a-python` and
diffing the JSON. Both generate from the same `.proto`, so a disagreement is an interop
divergence rather than a matter of preference.

Reproduced on `@a2a-js/sdk` 1.0.1 from npm and on `main` at `1c6eb32`, against
`a2a-python` at `cff6727`.

I'm happy to send a fix, but this one lives in `src/types/pb/a2a.ts`, which is generated
[...]

PR #644, opened 2026-08-10
test: add proto3 JSON conformance spec for the generated codecs

ishan-1010:
# Description

## What

Adds `test/proto_json_conformance.spec.ts`, covering how the generated codecs handle
proto3 JSON. Ten cases: five that pass today and pin behaviour that's already correct,
and five marked `it.fails` for the divergences in #640, #641, #642 and #643.

No source changes. This documents current behaviour and gives the fixes a regression net
to land against.

## Why

These messages go over the wire to SDKs generated from the same `.proto` with a different
toolchain, so the JSON has to agree. I built a differential harness that round-trips a
shared corpus through this SDK and `a2a-python` and diffs the output, and it turned up
four disagreements:

- unknown enum values become `"UNRECOGNIZED"`, which `a2a-python` rejects
- `Timestamp` is an unvalidated string and is never normalized to UTC
- `Part.data` set to `null` drops the content oneof
- multiple content oneof arms are accepted instead of rejected

Each is filed separately with a full repro. This PR is just the tests.

## On `it.fails`

`it.fails` asserts that the body fails, so the suite is green today and goes red the
moment a bug is fixed. That's deliberate: the fix can't land without flipping the marker,
so none of these can be quietly resolved and forgotten. I checked the inversion is strict
with a control test whose body passes, and vitest reports it as a failure.

The five passing cases are there as controls. If a change breaks known-good behaviour like
bytes or non-ASCII round-tripping, those go red on their own.

## Where the fixes should go

All four live in `src/types/pb/a2a.ts`, which is generated and marked `DO NOT EDIT`, and I
couldn't find a regeneration path in the repo. Two of them, the enum sentinel and the
oneof leniency, are `ts-proto` behaviour rather than anything configurable, so I didn't
want to guess. Happy to send the fixes once you've said whether you'd prefer them patched
into the generated output, handled in a normalization layer at the transport boundary, or
raised upstream.

## Verification

```
npx vitest run                                        69 files, 1497 tests, all pass
npx vitest run test/proto_json_conformance.spec.ts    10 tests, all pass
npx tsc --noEmit -p tsconfig.test.json                clean
npx eslint test/proto_json_conformance.spec.ts        clean
```

Reproduced against `@a2a-js/sdk` 1.0.1 from npm as well as `main` at `1c6eb32`, so none of
this is a main-only artifact. Python side pinned at `a2a-python` `cff6727`.

Parts of this were put together with AI assistance. Every case was verified by hand
against both SDKs before filing.

Refs #640, #641, #642, #643

### DIFF
--- a/test/proto_json_conformance.spec.ts
+++ b/test/proto_json_conformance.spec.ts
@@ -1,0 +1,88 @@
+import { describe, it, expect } from 'vitest';
+import { Part, TaskStatus } from '../src/index.js';
+
+/**
+ * proto3 JSON conformance for the generated codecs.
+ *
+ * These same messages go over the wire to SDKs that generate from the same
+ * .proto with a different toolchain, so the JSON has to match. Cases were
+ * found by round-tripping a shared corpus through this SDK and a2a-python and
+ * diffing the output.
+ *
+ * `it.fails` marks a case that is known-broken today. When one gets fixed the
+ * marker starts failing, so the fix can't land without flipping it.
+ */
+
+describe('proto3 JSON conformance', () => {
+  describe('enums', () => {
+    // Baseline: both spellings of a known value work, so the harness itself is sound.
+    it('accepts a known value by name', () => {
+      const out = TaskStatus.toJSON(TaskStatus.fromJSON({ state: 'TASK_STATE_WORKING' })) as any;
+      expect(out.state).toBe('TASK_STATE_WORKING');
+    });
+
+    it('accepts a known value by number', () => {
+      const out = TaskStatus.toJSON(TaskStatus.fromJSON({ state: 2 })) as any;
+      expect(out.state).toBe('TASK_STATE_WORKING');
+    });
+
+    // TaskState currently stops at 8. When the spec adds 9, a 1.0 SDK sitting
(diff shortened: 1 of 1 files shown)


ishan-1010:
Bumping this, been about 5 weeks with no review. Still green, happy to adjust anything if it's not quite what you're looking for. @ishymko or @bartek-gralewicz, tagging you since you've been reviewing most recent PRs here.

bartek-gralewicz:
> Bumping this, been about 5 weeks with no review. Still green, happy to adjust anything if it's not quite what you're looking for. @ishymko or @bartek-gralewicz, tagging you since you've been reviewing most recent PRs here.

Hey, sorry for keeping this on a side lane. This is actually a nice addition that tracts a few of the open issues.

You mentioned where fixes (if made) should land. `src/types/pb/a2a.ts` is a no-edit file but recently, I split the generation part into types (which still live in `src/types/pb/a2a.ts`) and `toJSON`/`fromJSON` methods which now live in manually maintained `src/types/protojson.ts`. Any JSON conversion fixes should land directly there.

bartek-gralewicz (approved):
Thank you for adding this. Sorry for the delay on the review.

PR #645, opened 2026-08-11
fix(server): validate push-notification URLs before dispatch (SSRF)

SashaMIT:
## Summary

`DefaultPushNotificationSender` POSTed to whatever webhook URL a client stored, with no scheme or resolved-IP checks. Create paths (`createTaskPushNotificationConfig` and embedded `taskPushNotificationConfig` on send) also stored those URLs unchecked.

This adds fail-closed validation (sibling of a2a-python#1164 / #1173 and a2a-go push SSRF hardening):

- http/https only
- DNS resolve (or literal IP) must be public unicast (incl. CGNAT / benchmarking / multicast blocks)
- Applied at dispatch and at config creation / send embed
- `allowPrivatePushUrls` opt-in on sender and `DefaultRequestHandler` for trusted private/loopback webhooks
- `fetch` uses `redirect: 'error'` so redirects cannot bypass the initial-URL check

## Attacker model

Any client that can register a push config can make the agent server send authenticated POSTs to internal targets. Validation runs before store and before fetch.

## Test plan

- [x] `test/server/push_url_validation.spec.ts` (blocked IPs, dispatch drop, create reject, allowPrivate opt-in, revert-tested)
- [x] push config / sender serializer / integration suites
- [ ] CI

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -46,6 +46,10 @@ export { validateVersion, getSupportedVersions } from './version.js';
 export type { PushNotificationSender } from './push_notification/push_notification_sender.js';
 export { DefaultPushNotificationSender } from './push_notification/default_push_notification_sender.js';
 export type { DefaultPushNotificationSenderOptions } from './push_notification/default_push_notification_sender.js';
+export {
+  isBlockedPushIp,
+  pushUrlValidationError,
+} from './push_notification/push_url_validation.js';
 export type {
   PushNotificationStore,
   StoredPushNotificationConfig,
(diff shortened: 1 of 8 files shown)


SashaMIT:
Follow-up on this branch: also reject unsafe push webhook URLs at config creation and when `sendMessage` / `sendMessageStream` embed `taskPushNotificationConfig` (sibling of a2a-python#1173). Same `allowPrivatePushUrls` opt-in now exists on `DefaultRequestHandler` and is passed through to the default sender.

PR #650, opened 2026-08-13
ci(itk): switch to v2 itk launcher

JakubWorek:
# Description

Switches itk scripts to use new version of launcher

Confirmation and testing done in fork: https://github.com/JakubWorek/a2a-js/actions/runs/31665791892

### DIFF
--- a/.github/workflows/run-itk-nightly.yaml
+++ b/.github/workflows/run-itk-nightly.yaml
@@ -52,8 +52,11 @@ jobs:
         run: npm ci
 
       - name: Clone a2a-itk
-        working-directory: itk
-        run: git clone -b main --depth 1 https://github.com/a2aproject/a2a-itk.git a2a-itk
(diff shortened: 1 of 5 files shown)


.github/workflows/run-itk-nightly.yaml
```diff
@@ -51,6 +51,30 @@ jobs:
       - name: Install Node dependencies
         run: npm ci
 
+      - name: Clone a2a-itk
+        working-directory: itk
+        run: git clone -b main --depth 1 https://github.com/a2aproject/a2a-itk.git a2a-itk
```
bartek-gralewicz: Why not using the `checkout` action?

Same in `run-itk.yaml`.

JakubWorek: Fair point, i'll switch to use checkout

bartek-gralewicz (approved):

PR #657, opened 2026-08-14
fix(types): preserve unknown enum integer values in codecs

### LINKED ISSUE #640: [Bug]: unknown enum values become the string "UNRECOGNIZED", which a2a-python rejects

Varun-S10:
# Description

Fixes #640 🦕

---

### What was the bug?
In the proto3 JSON specification, enums (such as `TaskState` and `Role`) must remain forward-compatible. When an SDK receives an unknown or newly added numeric enum value from a newer peer (e.g. `state: 99` or `role: 42`), it must preserve that integer value so it can be safely passed through proxies, gateways, and relays without data loss.

Previously in `@a2a-js/sdk`:
- `taskStateFromJSON` and `roleFromJSON` mapped any unrecognized numeric value to `UNRECOGNIZED` (`-1`).
- `taskStateToJSON` and `roleToJSON` then serialized this into the literal string `"UNRECOGNIZED"`.
- When an `a2a-python` peer received `{"state": "UNRECOGNIZED"}`, it threw a `ParseError: Invalid enum value UNRECOGNIZED`. The original numeric value `99` was destroyed, breaking interoperability.

---

### How is it fixed?
1. **`src/types/pb/a2a.ts`**:
   - Updated `taskStateFromJSON` and `roleFromJSON` default switch cases to preserve unknown integer numbers (`typeof object === "number" ? object : UNRECOGNIZED`).
   - Updated `taskStateToJSON` and `roleToJSON` to return `string | number` so that unknown enum integers are emitted directly as numbers instead of `"UNRECOGNIZED"`.
2. **`src/client/transports/rest_transport.ts` & `src/samples/cli.ts`**:
   - Wrapped `taskStateToJSON` in `String(...)` when formatting URL query parameters and CLI logs.
3. **`test/types/enums.spec.ts`**:
   - Added unit tests for `TaskState` and `Role` covering known string names, known numbers, unknown numeric integer preservation, and message round-tripping.

---

### Verification
- Unknown enum numbers (`state: 99`, `role: 42`) round-trip cleanly as integers (`99`, `42`) instead of collapsing into `"UNRECOGNIZED"`.
- Known enum values continue to serialize as string names (e.g. `"TASK_STATE_WORKING"`, `"ROLE_USER"`).
- Unit tests in `test/types/enums.spec.ts` passed (`10/10`).
- Full test suite (`npm test`) passed (`69/69` test files, `1,497` tests).
- ESLint and TypeScript checks (`npm run lint`) passed with 0 errors.
- Build and bundle checks (`npm run build && npm run test-build`) passed cleanly.


bartek-gralewicz:
Hello @Varun-S10,

Thank you for looking into this and proposing a fix. The issue is real but the fix proposed in this PR is difficult to maintain. The changes made to the `src/types/pb/a2a.ts` should be reverted as this file is autogenerated and should not be modified manually (https://github.com/a2aproject/a2a-js/blob/main/src/types/pb/a2a.ts#L1). Any modification to this file would need to be reapplied with each new from-proto-generation.

A more sustainable approach would be to create codecs file that would reexport from original `pb/a2a.ts` with fixed methods where relevant. Additionally, `src/index` would export these codecs instead of `pb/a2a.ts`. Direct imports from `pb/a2a.ts` would be restricted using `eslint.config.mjs`.

What do you think about such approach?

Varun-S10:
> Hello @Varun-S10,
> 
> Thank you for looking into this and proposing a fix. The issue is real but the fix proposed in this PR is difficult to maintain. The changes made to the `src/types/pb/a2a.ts` should be reverted as this file is autogenerated and should not be modified manually (https://github.com/a2aproject/a2a-js/blob/main/src/types/pb/a2a.ts#L1). Any modification to this file would need to be reapplied with each new from-proto-generation.
> 
> A more sustainable approach would be to create codecs file that would reexport from original `pb/a2a.ts` with fixed methods where relevant. Additionally, `src/index` would export these codecs instead of `pb/a2a.ts`. Direct imports from `pb/a2a.ts` would be restricted using `eslint.config.mjs`.
> 
> What do you think about such approach?

Hi @bartek-gralewicz, Thanks for the feedback. I agree with this approach. I reverted all changes to src/types/pb/a2a.ts, created src/types/codecs.ts with the fixed enum serialization, and re-exported it via src/index.ts. I also added the ESLint restriction rule for pb/a2a in src/index.ts and merged the latest main. Could you please review it again?

eslint.config.mjs
```diff
@@ -25,6 +25,22 @@ export default defineConfig(
       ],
     },
   },
+  {
+    files: ['src/index.ts'],
```
bartek-gralewicz, line 29: This lint check shouldn't be scoped to `src/index.ts` only. It should be made for all files (`src/**/*.ts`) so that the change is effective.

This will also require updating all current `a2a.js` imports to `codecs.js`.

src/types/codecs.ts
```diff
@@ -0,0 +1,295 @@
...
+      obj.task = Task.toJSON(message.payload.value);
+    } else if (message.payload?.$case === 'message') {
+      obj.message = Message.toJSON(message.payload.value);
+    }
+    return obj;
+  },
+};
+
+export type StreamResponse = pb.StreamResponse;
+export const StreamResponse: MessageFns<pb.StreamResponse> = {
```
bartek-gralewicz, line 265: Let's add an arm for artifacts as well here to be complete. Despite no real use today, this is an easy place to miss in case of future updates to the proto.

src/types/codecs.ts
```diff
@@ -0,0 +1,295 @@
...
+  toJSON(message: pb.SendMessageResponse): unknown {
+    const obj: Record<string, unknown> = {};
+    if (message.payload?.$case === 'task') {
+      obj.task = Task.toJSON(message.payload.value);
+    } else if (message.payload?.$case === 'message') {
+      obj.message = Message.toJSON(message.payload.value);
+    }
+    return obj;
+  },
+};
```
bartek-gralewicz, line 262: What is the value of that? We just do what original `a2a.js` did.

src/types/codecs.ts
bartek-gralewicz: Maybe we can drop this completely for the same reason as [this](https://github.com/a2aproject/a2a-js/pull/657/changes#r3967947574)?

bartek-gralewicz (commented):
There are multiple similar PRs opened for these files (`a2a.js` - related fixes using `codecs.js`). Since all of these touch similar files, there will be a lot of merge conflicts once these are ready for merging. To avoid having the `main` branch with partial fixes only, I propose the following options:

1. Merging all `a2a.js / codecs.js` changes into a single PR and resolving multiple fixes at once. Easier to work on as changes happen for all cases as well as there are no merge conflicts but the PR will be significant.
2. Developing each change separately till all are approved and merging all together in a short period of time (resolving conflicts between each merge). Cumbersome to merge but atomic in changes.

Let me know what do you think about propositions above. From my perspective, option 1 looks much more stable and easier to develop+merge then option 2 but requires a lot of rework so that current spread-out PRs are grouped into one. A possibility would be to create an `epic` branch now for this and merge each PR into the epic, then merge the epic into main once complete.

bartek-gralewicz:
Hey, to allow more flexibility with applying this and other `a2a.ts`-related fixes, I changed the approach how we handle proto generations in our SDK. I extracted `toJSON`/`fromJSON` methods to a manually maintained file in this PR: https://github.com/a2aproject/a2a-js/pull/719.

I propose that we put a pause on this topic here until that path is complete. Then, pushing fixes to the `protojson.ts` file will be much simpler to develop and test.

bartek-gralewicz:
> Hey, to allow more flexibility with applying this and other `a2a.ts`-related fixes, I changed the approach how we handle proto generations in our SDK. I extracted `toJSON`/`fromJSON` methods to a manually maintained file in this PR: #719.
> 
> I propose that we put a pause on this topic here until that path is complete. Then, pushing fixes to the `protojson.ts` file will be much simpler to develop and test.

This and the follow-up https://github.com/a2aproject/a2a-js/pull/720 are both merged to main now. Can you update your changes so that they target the `protojson.ts` directly and drop the now obsolete `codecs` workaround?

Issue #624, opened 2026-08-06
Push notification body shape change between 0.3.14 and 1.0 with legacyCompat: { enabled: true }


BennetKunze:
Hi, thanks a lot for the ongoing work on the SDK and for the legacyCompat layer.

We're currently migrating from 0.3.14 to 1.0 and found a behaviour change (with `legacyCompat: { enabled: true }`) that we think breaks the compat contract for v0.3 clients. Neither the migration guide nor the compatibility guide mentions it explictily.

Given an AgentExecutor that publishes a `status-update` event (or an `artifact-update` event) to the event bus, the shape of the outgoing push notification diverges between the two SDK versions:

- `0.3.14`: the SDK loads the full Task from the TaskStore and pushes `JSON.stringify(task)`, a single consistent shape (kind: "task") for every event kind. Delta events are recognised by the SDK but consolidated into a full-task push. e.g.: https://github.com/a2aproject/a2a-js/blob/v0.3.14/src/server/request_handler/default_request_handler.ts#L628
- `1.0 with legacyCompat: { enabled: true } and V03PushNotificationSerializer`: the SDK maps each executor event to a StreamResponse with a matching `$case` (task, statusUpdate, artifactUpdate, message), and the compat serializer emits a matching v0.3 wire body (kind: "task", kind: "status-update", kind: "artifact-update"). Delta events are pushed as deltas, not consolidated. e.g. https://github.com/a2aproject/a2a-js/blob/v1.0.1/src/server/request_handler/default_request_handler.ts#L1006

For downstream receivers built against the 0.3.14-actual shape (in our case both internal test doubles and business-application callbacks), this is an unannounced breaking change on the wire.

The [migration guide](https://github.com/a2aproject/a2a-js/blob/main/docs/migration-guide.md) documents the interface change from `send(task: Task)` to `send(streamResponse: StreamResponse, context)`. The [compat guide](https://github.com/a2aproject/a2a-js/blob/main/docs/compatibility-v0_3.md#push-notifications-routed-per-webhook) describes the compat push body as "the bare event object (v0.3 Task, TaskStatusUpdateEvent, or TaskArtifactUpdateEvent)". Neither points out that this is a deviation from what `0.3.14` actually emitted, so a `0.3.14`-upgrader has no way to anticipate the wire-shape change.

We have a few questions:

[...]

PR #658, opened 2026-08-14
fix(server): send full Task on every v0.3 push trigger

RonaldMarske:
## What & why

As stated in Bug #624 the v0.3 spec (§9.5, and confirmed against the `v0.3.14` reference implementation), a v0.3 push notification must deliver the **full `Task`** on *every* trigger — task, status, and artifact updates alike. The v1 core emits per-event `StreamResponse`s, so status/artifact triggers were being pushed as bare `TaskStatusUpdateEvent` / `TaskArtifactUpdateEvent` bodies. This restores the v0.3 contract while leaving v1 untouched (v1 push keeps mirroring the client stream verbatim, `historyLength` trim included).

The switch is gated on the a2a request's version (`isLegacyVersion(context.requestedVersion)`).

## Why the code lives in `DefaultRequestHandler`

The snapshot has to be captured **synchronously and in order**, right after `resultManager.processEvent(event)` — the drain loop is the only place that holds the ordered event, the authoritative merged `Task`, and the wire version at that moment. It also carries the **least legacy into the v1 code base**: the sender and serializers stay version-agnostic, and the single v0.3 gate lives here.

It's also simply the most convenient place: in `sendMessageStream` the mapper already feeds **both** the client SSE `yield` and the push, and v0.3 streaming must stay event-shaped while only the push becomes a `Task`. So the mapper takes a `deliveryTarget` and is called once per target — raw events for the stream, full `Task` for the legacy push.

## Why we don't read the Task in the sender

The sender is fire-and-forget, so loading the Task at dispatch time races the event stream — by the time the async read resolves, a later event may have advanced the store past the triggering state. We tried exactly this in our own code base with a DB-backed `TaskStore`, and it failed consistently in our tests through that race. Freezing the snapshot by value in the drain loop side-steps it.

## Alternatives we considered

- **Per-config serializer plumbing** — thread the frozen snapshot through `send` → `_dispatchNotification` → `serialize(streamResponse, { task })` and let `V03PushNotificationSerializer` pick the `Task`. It's the *more correct* option (it keys the Task-vs-event choice off each webhook's registered `wireVersion`), but the downsides outweigh it for a v0.3-only fix:
  - carries too much legacy through the whole code base,
  - keeps the `Task` bubbling through every layer,
  - would push the `Task` to the v1 serializer, where it's useless,
  - adds unused code just to hold the interface contracts.

## `_mapEventToStreamResponse`: dropped `taskStore.load`

The `task` case previously did `await taskStore.load(taskId, context)`. We now read `resultManager.getCurrentTask()` (deep-cloned) instead — mainly a performance improvement for anyone running an alternative `TaskStore` (as we do with a DB): it drops a per-event store round-trip on the hot event path in favor of the in-memory task the `ResultManager` already holds for this request.

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -54,7 +54,6 @@ import { PushNotificationSender } from '../push_notification/push_notification_s
 import { DefaultPushNotificationSender } from '../push_notification/default_push_notification_sender.js';
 import { ServerCallContext } from '../context.js';
 import { DEFAULT_PAGE_SIZE } from '../../constants.js';
-import { isLegacyVersion } from '../../version_utils.js';
 import {
   AUTH_REQUIRED_STATE_LIST,
   INTERRUPTED_STATE_LIST,
@@ -74,12 +73,6 @@ export interface DefaultRequestHandlerOptions {
    * the default list.
    */
   keepBusAliveStates?: TaskState[];
-}
-
-/** Destination a mapped {@link StreamResponse} is headed for. */
-enum StreamResponseDeliveryTarget {
-  ClientStream = 'clientStream',
-  PushNotification = 'pushNotification',
 }
 
 /**
@@ -278,13 +271,12 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         await resultManager.processEvent(event);
 
         try {
-          const streamResponse = this._mapEventToStreamResponse(
-            event,
+          const streamResponse = this._mapEventToStreamResponse(event, context, resultManager);
+          await this._sendPushNotificationIfNeeded(
             context,
-            resultManager,
-            StreamResponseDeliveryTarget.PushNotification
+            streamResponse,
+            structuredClone(resultManager.getCurrentTask())
           );
-          await this._sendPushNotificationIfNeeded(context, streamResponse);
         } catch (error) {
           console.error(`Error sending push notification: ${error}`);
         }
@@ -720,12 +712,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
 
         await resultManager.processEvent(event);
 
-        const streamResponse = this._mapEventToStreamResponse(
-          event,
-          context,
-          resultManager,
-          StreamResponseDeliveryTarget.ClientStream
-        );
+        const streamResponse = this._mapEventToStreamResponse(event, context, resultManager);
         if (streamResponse.payload?.$case === 'task') {
           this._applyHistoryLengthSemantics(
             streamResponse.payload.value,
@@ -733,16 +720,11 @@ export class DefaultRequestHandler implements A2ARequestHandler {
           );
         }
 
-        // v0.3 push carries the full Task on every trigger; v1 mirrors the stream.
-        const pushNotificationResponse = isLegacyVersion(context.requestedVersion)
-          ? this._mapEventToStreamResponse(
-              event,
-              context,
-              resultManager,
-              StreamResponseDeliveryTarget.PushNotification
-            )
-          : streamResponse;
-        await this._sendPushNotificationIfNeeded(context, pushNotificationResponse);
+        await this._sendPushNotificationIfNeeded(
+          context,
+          streamResponse,
+          structuredClone(resultManager.getCurrentTask())
+        );
         yield streamResponse;
       }
     } finally {
@@ -1012,8 +994,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
   private _mapEventToStreamResponse(
     event: AgentExecutionEvent,
     context: ServerCallContext,
-    resultManager: ResultManager,
-    deliveryTarget: StreamResponseDeliveryTarget
+    resultManager: ResultManager
   ): StreamResponse {
     switch (event.kind) {
       case 'task': {
@@ -1025,15 +1006,6 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         return { payload: { $case: 'message', value: event.data } };
       case 'statusUpdate':
       case 'artifactUpdate': {
-        const isLegacyPushNotification =
-          deliveryTarget === StreamResponseDeliveryTarget.PushNotification &&
-          isLegacyVersion(context.requestedVersion);
-
-        const currentTask = isLegacyPushNotification ? resultManager.getCurrentTask() : undefined;
-        if (currentTask) {
-          return { payload: { $case: 'task', value: structuredClone(currentTask) } };
-        }
-
         return event.kind === 'statusUpdate'
           ? { payload: { $case: 'statusUpdate', value: event.data } }
           : { payload: { $case: 'artifactUpdate', value: event.data } };
@@ -1051,10 +1023,11 @@ export class DefaultRequestHandler implements A2ARequestHandler {
    */
   private async _sendPushNotificationIfNeeded(
     context: ServerCallContext,
-    streamResponse: StreamResponse
+    streamResponse: StreamResponse,
+    task?: Task
   ): Promise<void> {
     if (this.agentCard.capabilities?.pushNotifications && this.pushNotificationSender) {
-      this.pushNotificationSender.send(streamResponse, context).catch((error) => {
+      this.pushNotificationSender.send(streamResponse, context, task).catch((error) => {
         console.error(`Failed to send push notification:`, error);
       });
     }
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -54,6 +54,7 @@ import { PushNotificationSender } from '../push_notification/push_notification_s
 import { DefaultPushNotificationSender } from '../push_notification/default_push_notification_sender.js';
 import { ServerCallContext } from '../context.js';
 import { DEFAULT_PAGE_SIZE } from '../../constants.js';
+import { isLegacyVersion } from '../../version_utils.js';
 import {
   AUTH_REQUIRED_STATE_LIST,
   INTERRUPTED_STATE_LIST,
@@ -73,6 +74,12 @@ export interface DefaultRequestHandlerOptions {
    * the default list.
    */
   keepBusAliveStates?: TaskState[];
+}
+
+/** Destination a mapped {@link StreamResponse} is headed for. */
+enum StreamResponseDeliveryTarget {
+  ClientStream = 'clientStream',
+  PushNotification = 'pushNotification',
 }
 
 /**
@@ -271,7 +278,12 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         await resultManager.processEvent(event);
 
         try {
-          const streamResponse = await this._mapEventToStreamResponse(event, context);
+          const streamResponse = this._mapEventToStreamResponse(
+            event,
+            context,
+            resultManager,
+            StreamResponseDeliveryTarget.PushNotification
+          );
           await this._sendPushNotificationIfNeeded(context, streamResponse);
         } catch (error) {
           console.error(`Error sending push notification: ${error}`);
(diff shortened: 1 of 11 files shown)


bartek-gralewicz:
Hey, thank you for submitting a PR on this issue. I have a comment regarding the proposal:

The code changes in this PR are taking the information about the version from the trigger and not from the moment the webhook was registered. Hence, if a v0.3 registered webhook sends a v1.0 trigger, it would receive a v1.0 `StreamResponse{statusUpdate}` when the v0.3 `Task` should be returned. Same goes the other way around (v1.0 webhook with v0.3 trigger).
Note: Expectation that the webhook registration version is the controlling component is even mentioned in the original issue (3rd clause).

I believe that `src/server/push_notification/push_notification_serializer.ts` might be the best location for such logic but I am open for other proposals as well.

RonaldMarske:
> Hey, thank you for submitting a PR on this issue. I have a comment regarding the proposal:
> 
> The code changes in this PR are taking the information about the version from the trigger and not from the moment the webhook was registered. Hence, if a v0.3 registered webhook sends a v1.0 trigger, it would receive a v1.0 `StreamResponse{statusUpdate}` when the v0.3 `Task` should be returned. Same goes the other way around (v1.0 webhook with v0.3 trigger). Note: Expectation that the webhook registration version is the controlling component is even mentioned in the original issue (3rd clause).
> 
> I believe that `src/server/push_notification/push_notification_serializer.ts` might be the best location for such logic but I am open for other proposals as well.

Hey @bartek-gralewicz , thanks for reviewing. I moved the decision to the Serializer now by handling the task all the way through as an additional parameter. This is the option i wanted to avoid but the only one i saw which still holds the snap shot of the exact moment.

Hope this is fine now. Sorry for the force push, rebased my branch to latest main.

bartek-gralewicz:
> > Hey, thank you for submitting a PR on this issue. I have a comment regarding the proposal:
> > The code changes in this PR are taking the information about the version from the trigger and not from the moment the webhook was registered. Hence, if a v0.3 registered webhook sends a v1.0 trigger, it would receive a v1.0 `StreamResponse{statusUpdate}` when the v0.3 `Task` should be returned. Same goes the other way around (v1.0 webhook with v0.3 trigger). Note: Expectation that the webhook registration version is the controlling component is even mentioned in the original issue (3rd clause).
> > I believe that `src/server/push_notification/push_notification_serializer.ts` might be the best location for such logic but I am open for other proposals as well.
> 
> Hey @bartek-gralewicz , thanks for reviewing. I moved the decision to the Serializer now by handling the task all the way through as an additional parameter. This is the option i wanted to avoid but the only one i saw which still holds the snap shot of the exact moment.
> 
> Hope this is fine now. Sorry for the force push, rebased my branch to latest main.

Hey, thanks for following up on this. Are you sure that you included all of your changes after the rebase? It seems like your comment above misaligns with the PR changes at the moment.

RonaldMarske:
> Hey, thanks for following up on this. Are you sure that you included all of your changes after the rebase? It seems like your comment above misaligns with the PR changes at the moment.

Thanks for pointing out, the changes got lost on their way up - now they landed.

src/server/request_handler/default_request_handler.ts
```diff
@@ -976,29 +985,31 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
         return { payload: { $case: 'message', value: event.data } };
       case 'statusUpdate':
-        return { payload: { $case: 'statusUpdate', value: event.data } };
-      case 'artifactUpdate':
-        return { payload: { $case: 'artifactUpdate', value: event.data } };
+      case 'artifactUpdate': {
+        return event.kind === 'statusUpdate'
+          ? { payload: { $case: 'statusUpdate', value: event.data } }
+          : { payload: { $case: 'artifactUpdate', value: event.data } };
+      }
```
bartek-gralewicz: What is the reason for this change? I would advise restoring the original per `case` split.

test/server/push_notification_integration.spec.ts
```diff
@@ -276,15 +277,20 @@ describe('Push Notification Integration Tests', () => {
...
       assert.deepEqual(
         secondNotification.body,
         StreamResponse.toJSON({
           payload: {
             $case: 'statusUpdate',
             value: {
-              taskId: taskId,
-              contextId: contextId,
+              taskId,
+              contextId,
```
bartek-gralewicz: nit: Is there a reason for this change?

test/server/push_notification_integration.spec.ts
```diff
@@ -303,8 +309,8 @@ describe('Push Notification Integration Tests', () => {
...
```
bartek-gralewicz: nit: Is there a reason for this change?

src/server/request_handler/default_request_handler.ts
```diff
@@ -976,29 +985,31 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+  ): StreamResponse {
     switch (event.kind) {
       case 'task': {
-        const taskId = event.data.id;
-        const fullTask = await this.taskStore.load(taskId, context).catch((error): Task | null => {
-          console.warn('Failed to load full task from store, falling back to event data:', error);
-          return null;
-        });
-        return { payload: { $case: 'task', value: fullTask || event.data } };
+        const currentTask = resultManager.getCurrentTask();
```
bartek-gralewicz: Why changing the `taskStore.load()` to `resultManager.getCurrentTask()` here? Such change seems to be out of scope of this PR.

src/server/request_handler/default_request_handler.ts
```diff
@@ -271,8 +271,12 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         await resultManager.processEvent(event);
 
         try {
-          const streamResponse = await this._mapEventToStreamResponse(event, context);
-          await this._sendPushNotificationIfNeeded(context, streamResponse);
+          const streamResponse = this._mapEventToStreamResponse(event, context, resultManager);
+          await this._sendPushNotificationIfNeeded(
+            context,
+            streamResponse,
+            structuredClone(resultManager.getCurrentTask())
```
bartek-gralewicz, line 278: Do we really need all these `structureClone()` calls everywhere? Now, the task is deeply copied twice when processed. Once in `_mapEventToStreamResponse` and once on the call site. Additionally, these copies seem unnecessary for v1.0.

src/server/request_handler/default_request_handler.ts
RonaldMarske: reverted to the previous state.

test/server/push_notification_integration.spec.ts
RonaldMarske: reverted

test/server/push_notification_integration.spec.ts
RonaldMarske: reverted

src/server/request_handler/default_request_handler.ts
RonaldMarske: left the change basically in as a slight performance improvement. ResultManager does load the task from the store, updates and saves. Then `_mapEventToStreamResponse` loads the task again from the store.

Might not be a big deal if you use the DefaultInMemoryStore - a DB store is a different story.

Reverted for now - will create a new PR for this only.

src/server/request_handler/default_request_handler.ts
RonaldMarske: Good to point that out, i traced the path down again and ResultManager and TaskStore ensure a safe to copy element already. Therefore removed the structuredClone from the call.

src/server/request_handler/default_request_handler.ts
```diff
@@ -997,8 +1008,9 @@ export class DefaultRequestHandler implements A2ARequestHandler {
         return { payload: { $case: 'message', value: event.data } };
       case 'statusUpdate':
         return { payload: { $case: 'statusUpdate', value: event.data } };
-      case 'artifactUpdate':
+      case 'artifactUpdate': {
         return { payload: { $case: 'artifactUpdate', value: event.data } };
+      }
```
bartek-gralewicz:
```suggestion
      case 'artifactUpdate':
        return { payload: { $case: 'artifactUpdate', value: event.data } };
```

src/server/request_handler/default_request_handler.ts
```diff
@@ -976,9 +985,11 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
 
   /**
-   * Maps an {@link AgentExecutionEvent} to a `StreamResponse`. For Task
-   * events the full task is loaded from the store to include accumulated
-   * history and artifacts.
+   * Maps an {@link AgentExecutionEvent} to a `StreamResponse`, resolving
+   * Task events from the {@link ResultManager} snapshot. For a legacy
+   * (v0.3) push notification, status/artifact updates are replaced by the
+   * full Task, which v0.3 always sends; the client stream and v1.0 keep
+   * the raw event.
```
bartek-gralewicz: This docstring is out of date now.

src/server/request_handler/default_request_handler.ts
```diff
@@ -1012,10 +1024,11 @@ export class DefaultRequestHandler implements A2ARequestHandler {
    */
   private async _sendPushNotificationIfNeeded(
     context: ServerCallContext,
-    streamResponse: StreamResponse
+    streamResponse: StreamResponse,
+    task?: Task
   ): Promise<void> {
     if (this.agentCard.capabilities?.pushNotifications && this.pushNotificationSender) {
-      this.pushNotificationSender.send(streamResponse, context).catch((error) => {
+      this.pushNotificationSender.send(streamResponse, context, task).catch((error) => {
```
bartek-gralewicz, line 1028:
```suggestion
      const taskSnapshot = task ? structuredClone(task) : undefined;
      this.pushNotificationSender.send(streamResponse, context, taskSnapshot).catch((error) => {
```

I believe that a clone here is needed as for v0.3 registered webhook, if `historyLength` is set, the `task` could be mutated.

bartek-gralewicz (commented):
Added a few small suggestions. The rest looks good

src/server/request_handler/default_request_handler.ts
RonaldMarske: That's a good catch - in the streaming case the `_sendPushNotificationIfNeeded` is called after the history entries where modified. This is a direct mutation, so cloning here is to late as the history entries are gone already.

will capture the snapshot before the mutation happens.

RonaldMarske: I checked this again more deeply - the sendMessageStream does only modify streamResponse from `_mapEventToStreamResponse` - as this loads the task fresh from the Store, the task from the resultManager stays as is.

so in both cases a structuredClone is not required. This changes with #671 if this get's accepted. It would make sense to block #671 until this one is merged and then the clone should be applied before the history modification.

src/server/request_handler/default_request_handler.ts
RonaldMarske: fixed

src/server/request_handler/default_request_handler.ts
bartek-gralewicz: I agree on the `sendMessageStream` path but I think that the `sendMessage` is the trouble here.
You can even take a look at the test that was added in this PR, specifically lines [push_notification_integration.spec.ts:363-364](https://github.com/RonaldMarske/a2a-js/blob/bbb89358fed210f1100ad19873b514c483f38dac/test/server/push_notification_integration.spec.ts#L363-L364). The test works only because it doesn't challenge the history.

You can add the following lines to that test and see it fail:
```
const historyLengths = receivedNotifications.map((n) => n.body.history?.length ?? 0);
assert.deepEqual(historyLengths, [1, 1, 1]);
```

The test result is `[1, 1, 0]` instead of `[1, 1, 1]`. If you apply the clone like in my suggestion above, the test will pass. I tested this locally to make sure.

bartek-gralewicz: Btw, while we are at it, please include the lines above in the mentioned test so that history is checked as well.

RonaldMarske: Very interesting find. I added your check to the test to ensure the history length is given.
Turns out that this is a race condition and your fix prevented this. 

I still think that `_sendPushNotificationIfNeeded` is the wrong spot for this clone - so i placed it in `_processEvents`. I added also the condition that the clone only happens in case of a task and it turns out the test keeps failing. This ends up in always cloning the task before sending it to the PushNotificationSender - t least in the `sendMessage` case.

I added also the same test for the streaming case to ensure the same behaviour for the streaming case and found another bug which was hidden due to the modification of the history in `sendMessageStream`. So the v03Serializer now always returns the provided task vor statusUpdate, ArtefactUpdate and Task.

If a modification of the Task happens, the test will make this visible.

bartek-gralewicz (approved):
Thanks for contributing with this change.

Note: we currently have issues with ITK testing suite which are being resolved. Disregard the failing CI check. It should be fixed shortly.

PR #662, opened 2026-08-19
fix: clone push notification configs before storing

anxkhn:
`InMemoryPushNotificationStore.save` retained the caller's `TaskPushNotificationConfig` object. Mutating that object after saving could silently change the stored webhook URL, token, or authentication data.

This change stores a deep clone in `src/server/push_notification/push_notification_store.ts` while preserving the generated ID on the caller's object. `DefaultRequestHandler.sendMessage` and `sendMessageStream` also clone inline configurations before passing them to a store, preventing aliasing in custom implementations.

Regression coverage verifies input isolation, generated-ID behavior, and inline `sendMessage` configuration handling.

Tested with:

```sh
npx --no-install vitest run test/server/push_notification_store.spec.ts test/server/request_handler/push_notification_config.spec.ts
```

### DIFF
--- a/src/server/push_notification/push_notification_store.ts
+++ b/src/server/push_notification/push_notification_store.ts
@@ -94,7 +94,10 @@ export class InMemoryPushNotificationStore implements PushNotificationStore {
       entries.splice(existingIndex, 1);
     }
 
-    entries.push({ config: pushNotificationConfig, wireVersion });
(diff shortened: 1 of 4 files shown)


JakubWorek (approved):
Hi @anxkhn, thank you for your contribution, lgtm, approved

PR #665, opened 2026-08-20
fix(deps): replace `uuid` package with `crypto.randomUUID()`

ishymko:
This eliminates `uuid` dependency completely.


JakubWorek (approved):
LGTM

PR #666, opened 2026-08-20
fix: make spec-optional object properties optional in TypeScript types

### LINKED ISSUE #660: [Bug]: Some optional properties are required
### What happened?

The `filename` property of the `Part` interface is clearly documented as optional, yet it's a required property, which doesn't even accept `undefined`. Within the same interface, there's also `metadata`, which is again documented as optional, but it's in fact a required property (although in this case at least it does accept undefined). In fact, according to [the spec for Part,](https://a2a-protocol.org/v1.0.1/specification/#416-part) **ALL** properties should be optional.

There are many other similar cases, for example in the `Message` interface, where `contextId`, `taskId` and `metadata` are all documented as optional, yet required in code.

I'm aware the code is generated from the `a2a.proto` from the spec repo, but that's also the case for the Python SDK, for example, which doesn't have this issue.

On top of that, I don't understand the need to use `Part.content` as a wrapper over the actual content of the part. Why not use `Part.text` / `raw` / etc, as is done in the Python library?

EDIT:
There are also examples of the opposite kind of problem, having properties which should always be required as possibly undefined, like in the case of `TaskArtifactUpdateEvent.artifact`

### Relevant log output

```shell

```

### Code of Conduct

Varun-S10:
# Description

According to the A2A v1.0.1 specification, properties such as `filename`, `mediaType`, and `metadata` on `Part`, `contextId`, `taskId`, and `extensions` on `Message`, and optional fields on `Task`, `Artifact`, `SendMessageConfiguration`, and `AgentCard` are optional.

Previously, these properties were typed as required in the TypeScript interfaces, causing compiler errors (`TS2739`) when omitting them.

This PR makes these spec-optional properties optional (`property?: type | undefined`) across the TypeScript types and updates the serialization and conversion helpers to handle undefined values safely.

Fixes #660 🦕


src/types/pb/a2a.ts
```diff
@@ -294,50 +294,50 @@ export interface Artifact {
...
 /** A task delta where an artifact has been generated. */
 export interface TaskArtifactUpdateEvent {
   /** The ID of the task for this artifact. */
   taskId: string;
   /** The ID of the context that this task belongs to. */
-  contextId: string;
+  contextId?: string | undefined;
   /** The artifact that was generated or updated. */
-  artifact:
+  artifact?:
```
wpp-vlad-ardelean: Artifact should be mandatory here, according to the spec (`undefined` shouldn't be among the allowed types)

Varun-S10: Fixed. artifact is now mandatory (artifact: Artifact) without undefined as per the spec. Pushed the update.

PR #669, opened 2026-08-20
feat: make event bus keep-alive states configurable

### LINKED ISSUE #620: [Clarification]: Evetbus is closed directly after executor has finished
We're using the A2A SDK in a different approach that it was actually designed for. Our Executor is basically a proxy to other Agents and in some cases we talk to target Agents via push notification to get their response.

Additionally, we overwrote the DefaultEventBus and replaces it with a Database event bus which adds additional time onto the loop. I don't want to go to much in detail how we did this.

While upgrading to A2A SDK v1 we figured out that we have non deterministic behaviour that sometimes the task was completed, and sometimes it got stuck entirely.

After investigation i found out that you call `_settleBus` in a finally call after the Executor has finished which then sends a finished event to the event bus, and closes the event bus for this task.

In normal circumstances i guess this is fine, to assume that when the Executor is finished, the Task/Loop should be finished as well.  In our case never the less, due to the DB loop the event bus was closed before we could read the messages from the DB and everything breaks down. On Top, we have communication with other Agents over Push notification, which results in a very short living Executor because it's only sending the Call to the other agent and finishes - which does not mean the Task is finished.

With the previous A2A SDK 0.3.14 this was not an issue as the loop was closed based on the events status.

Our plan for a mitigation would be to override `_settleBus` and basically do nothing and check in our PersistentEventBus, which we override anyway for the finished events and call the EventBusManager to close the bus for the desired task. This is the easiest way to do this, to not override the DefaultRequestHandler entirely.

1. Do you see any issues with following this approach as we're basically overriding a `_` "private" method. Also in regards of future versions.
2. Would it be an option for you to reconsider your approach as other consumer might run into the same issue when adding more delay or using a DB EventBus as we do.
3. Maybe it can be turned on/off using specific settings/configurations

So, not really a bug, or Feature Request, more like a clarification with you guys about your plans, and talking about options.

Thanks a lot for you Support and awesome work.

russlan23:
# Description

`DefaultRequestHandler` always cleaned up the execution event bus unless the final state was `INPUT_REQUIRED` or `AUTH_REQUIRED`. That prevents applications with persistent event buses from retaining a bus for additional states.

This change:

- adds an exported `DefaultRequestHandlerOptions` type with `keepBusAliveStates`
- preserves the existing interrupted-state list as the default
- copies the configured states into a `Set` and uses it when settling the bus
- covers a configured terminal state while retaining the existing default-cleanup test

- [x] Make the Pull Request title follow the Conventional Commits specification.
- [x] Appropriate docs were updated (the exported option includes API documentation).

Tests:

- `npm test` (1,491 tests)
- `npm run lint:ci`
- `npm run build`
- `npm run test-build`

Fixes #620

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -68,6 +68,9 @@ export interface DefaultRequestHandlerOptions {
   /**
    * Task states that keep the execution event bus alive after the agent
    * executor returns. Defaults to INPUT_REQUIRED and AUTH_REQUIRED.
+   * To add another keep-alive state while preserving those defaults, include
+   * both default states and the additional state. Any custom value overrides
+   * the default list.
    */
   keepBusAliveStates?: TaskState[];
 }
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -64,6 +64,14 @@ import {
 import { AgentCardSignatureGenerator } from '../../signature.js';
 import { extractErrorMessage } from '../../errors/index.js';
 
+export interface DefaultRequestHandlerOptions {
+  /**
+   * Task states that keep the execution event bus alive after the agent
+   * executor returns. Defaults to INPUT_REQUIRED and AUTH_REQUIRED.
+   */
+  keepBusAliveStates?: TaskState[];
+}
+
 /**
  * Default implementation of the A2A request handler.
  *
@@ -82,6 +90,7 @@ export class DefaultRequestHandler implements A2ARequestHandler {
   private readonly pushNotificationSender?: PushNotificationSender;
   private readonly extendedAgentCardProvider?: AgentCard | ExtendedAgentCardProvider;
   private readonly agentCardSignatureGenerator?: AgentCardSignatureGenerator;
+  private readonly keepBusAliveStates: Set<TaskState>;
 
   constructor(
     agentCard: AgentCard,
@@ -91,7 +100,8 @@ export class DefaultRequestHandler implements A2ARequestHandler {
     pushNotificationStore?: PushNotificationStore,
     pushNotificationSender?: PushNotificationSender,
     extendedAgentCardProvider?: AgentCard | ExtendedAgentCardProvider,
-    agentCardSignatureGenerator?: AgentCardSignatureGenerator
+    agentCardSignatureGenerator?: AgentCardSignatureGenerator,
+    options: DefaultRequestHandlerOptions = {}
   ) {
     this.agentCard = agentCard;
     this.taskStore = taskStore;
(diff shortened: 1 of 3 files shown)


src/server/request_handler/default_request_handler.ts
```diff
@@ -64,6 +64,14 @@ import {
 import { AgentCardSignatureGenerator } from '../../signature.js';
 import { extractErrorMessage } from '../../errors/index.js';
 
+export interface DefaultRequestHandlerOptions {
+  /**
+   * Task states that keep the execution event bus alive after the agent
+   * executor returns. Defaults to INPUT_REQUIRED and AUTH_REQUIRED.
```
bartek-gralewicz, line 70:
```suggestion
  /**
   * Task states that keep the execution event bus alive after the agent
   * executor returns. Defaults to INPUT_REQUIRED and AUTH_REQUIRED.
   * If a user wants an additional state to keep the bus alive, both default states
   * and new user's state need to be provided. Any custom value in this field
   * will override default states.
```

test/server/request_handler/streaming_errors.spec.ts
```diff
@@ -328,6 +328,66 @@ describe('DefaultRequestHandler streaming error synthesis (_runStreamExecutor)',
     expect(eventBusManager.getByTaskId(observedRequestTaskId)).toBeUndefined();
   });
 
+  it('keeps the event bus alive when the final state is configured', async () => {
```
bartek-gralewicz: I would suggest adding another test that actually extends the `keepBusAliveStates`, not reduces them as this is the main purpose of this update.

src/server/request_handler/default_request_handler.ts
russlan23: Updated the option documentation to clarify that a custom list replaces the defaults, and that callers should include INPUT_REQUIRED and AUTH_REQUIRED when extending them.

test/server/request_handler/streaming_errors.spec.ts
russlan23: Added a separate extension case with INPUT_REQUIRED, AUTH_REQUIRED, and COMPLETED while retaining the replacement case. The focused suite now passes 9 tests, and the full suite passes 1,492 tests.

test/server/request_handler/streaming_errors.spec.ts
```diff
@@ -328,6 +328,81 @@ describe('DefaultRequestHandler streaming error synthesis (_runStreamExecutor)',
...
+      messageId: 'msg-stream-err-configured-override',
+    },
+    {
+      description: 'when configured states extend the defaults',
+      keepBusAliveStates: [
+        TaskState.TASK_STATE_INPUT_REQUIRED,
+        TaskState.TASK_STATE_AUTH_REQUIRED,
+        TaskState.TASK_STATE_COMPLETED,
+      ],
+      messageId: 'msg-stream-err-configured-extended',
```
bartek-gralewicz: Both these cases challenge the same case. They do the same test in fact.

What I meant in my original comment was extending the test suite for non-terminal statuses like `TASK_STATE_WORKING`. Sorry for not being clear enough here.

The goal would be to hit the #620 scenario (non-terminal state -> executor returns -> late publish still reaches a consumer and gets persisted).

russlan23: Updated the regression to use configured WORKING alongside the default states. It now lets the executor return, publishes COMPLETED afterward through the retained bus, verifies the original stream receives it, and confirms the task store persists COMPLETED. The focused 9-test suite and all 1,492 tests pass, along with lint, build, and bundle checks.

bartek-gralewicz (approved):
Thanks for the PR!

PR #670, opened 2026-08-21
feat(compat): expose v0.3 to v1.0 translators as a subpath export

### LINKED ISSUE #646: Expose v0.3 to v1.0 conversion helpers as public API
Hi, thanks again for the ongoing work on the SDK and the legacyCompat layer.

We're on 1.0.1, migrating from 0.3.14. We have a receiving endpoint that is not an SDK client: agents reply to us via push notification against our own Express callback route. We receive a raw v0.3 body and have to convert it to v1.0 ourselves before handing it to the rest of our v1.0 pipeline.

The SDK already contains exactly the converters we need, internally named functions in `dist/compat/v0_3/client/index.js` and the server compat `index.js`. They are just not part of the public exports. From `@a2a-js/sdk/compat/v0_3/client` only `LegacyJsonRpcTransport`, `LegacyRestTransport`, `isLegacyAgentCard`, and `parseLegacyAgentCard` are exported, and no v0.3 to v1.0 conversion helpers are re-exported from the top-level entry either.

As a result we had to hand-reimplement the v0.3 wire to core conversion (role mapping, task-state mapping, part $case mapping including file bytes/uri, status/artifact/task/statusUpdate/artifactUpdate shapes), duplicating logic that already exists and is maintained inside the SDK.

Would you be open to exposing the v0.3 to v1.0 converters as public API from `@a2a-js/sdk/compat/v0_3`?

Thanks again for the great work on the SDK.

harshitwandhare:
# Description

`src/compat/v0_3/translate/` has bidirectional v0.3 <-> v1.0 payload translators, but nothing published re-exports them. A receiver that isn't an SDK client (a webhook route taking raw push bodies, for example) gets a v0.3-shaped body on the wire and has no supported way to turn it into a v1.0 value without reimplementing translation logic that already handles the non-obvious cases: `PushNotificationAuthenticationInfo.schemes` truncation, `final` recomputation, and the inverted `returnImmediately`/`blocking` polarity.

`compat/v0_3/README.md` states the translators are intentionally private and may change without a major bump, so this was a deliberate call rather than an oversight, and needed a maintainer decision on shape before a PR (see the [issue thread](https://github.com/a2aproject/a2a-js/issues/646)).

This change:

- adds `@a2a-js/sdk/compat/v0_3/translate` as its own subpath rather than widening the existing `compat/v0_3` entry. Measured both under `test-build:workers-safe`: widening the existing entry takes its bundled size from 22,677 to 62,971 bytes with no `sideEffects: false` to tree-shake it back out, so a Workers consumer that only wants `LEGACY_METHOD_MESSAGE_SEND` would carry the whole translation layer. The dedicated subpath leaves `compat/v0_3` untouched and costs 57,830 bytes only on import, same reasoning the README already gives for the other five subpaths
- re-exports `types/types.ts` (the v0.3 wire types the translators accept) as a `legacy` namespace, since its 123 exported names collide with the v1.0 ones at the package root (`Task`, `Message`, `AgentCard`); matches how the translate modules already refer to it internally
- updates `compat/v0_3/README.md`, which still described the translators as fully private
- adds `test/compat/v0_3/public_api.spec.ts`, covering the three v0.3 push-body shapes from #646 (task, status-update, artifact-update) through the published entry point rather than the internal module

- [x] Make the Pull Request title follow the Conventional Commits specification.
- [x] Appropriate docs were updated (README table + policy note).

Tests:

- `npm test` (1,528 tests, re-run after syncing onto `main` at #687)
- `npm run lint:ci`
- `npm run test-build:workers-safe`

Fixes #646


harshitwandhare:
The red ITK check on this PR is a stale transient, not a failure of the change, and I cannot re-run it myself. Details so nobody has to dig:

The job died 2 seconds into `Run ITK Tests (PR scenarios)`, before any scenario ran, at the proto-generation step:

```
++ ../node_modules/.bin/buf generate
Failure: permission_denied: 403 Forbidden
Done. Final exit code: 1
```

The step that precedes it, `Run Error Handling Tests`, passed 16/16. So the run never got as far as exercising anything this PR touches.

It looks like a registry blip confined to that evening. ITK ran at `2026-08-28T22:25Z` here and failed; on #691, also a fork PR, it ran at `07:03Z` the same day and passed, and #692 passed on `2026-09-01`. Every other open PR currently shows ITK green.

This PR's head has not moved since, so the failure just sits there. A re-run of that job should clear it.

The rest is current: the branch is even with `main` (`behind_by=0`) and the other 14 checks pass. The two bundle numbers the change argues from, 22,677 bytes for `compat/v0_3/index.js` against 57,830 for the new `compat/v0_3/translate` entry, were last measured on this exact head, which has not moved since, so the case for a separate subpath rather than widening the existing entry still stands.

@BennetKunze no rush, but this has been ready since you asked for it on #646 and I would rather it not look broken while it waits.

harshitwandhare:
Closing per the decision on #646: keeping the v0.3 to v1.0 translators internal is a deliberate semver call, not an oversight, so this subpath should not land. The compat README change here is dropped with it, since the README was accurate as written.

PR #671, opened 2026-08-25
refactor(server): small perfromance improvement when database task store is used

RonaldMarske:
# Avoid loading the Task twice from the TaskStore

## What changed

In `DefaultRequestHandler`, `_mapEventToStreamResponse` now reads the full Task
from the `ResultManager` (`getCurrentTask()`) instead of re-loading it from the
`TaskStore` for `task` events. It is now synchronous, and the output is
unchanged (the `|| event.data` fallback stays).

## Why

We run a database-backed store, so re-reading the Task is a round-trip per
streamed `task` event that adds up under load. It is pure overhead:
`processEvent`, run immediately before, has already built that Task. (Free on
the in-memory store.)


src/server/request_handler/default_request_handler.ts
```diff
@@ -980,17 +980,13 @@ export class DefaultRequestHandler implements A2ARequestHandler {
...
+  ): StreamResponse {
     switch (event.kind) {
       case 'task': {
-        const taskId = event.data.id;
-        const fullTask = await this.taskStore.load(taskId, context).catch((error): Task | null => {
-          console.warn('Failed to load full task from store, falling back to event data:', error);
-          return null;
-        });
+        const fullTask = resultManager.getCurrentTask();
         return { payload: { $case: 'task', value: fullTask || event.data } };
```
bartek-gralewicz:
```suggestion
        return { payload: { $case: 'task', value: fullTask ? structuredClone(fullTask) : event.data } };
```

PR #672, opened 2026-08-25
feat(itk): use shared scenarios

JakubWorek:
# Description

Making use of shared scenarios and run_itk script from itk repository


itk/run_itk.sh
```diff
@@ -1,211 +1,56 @@
...
-  rm -rf protos > /dev/null 2>&1 || true
-  rm -f instruction.proto > /dev/null 2>&1 || true
-  rm -f raw_results.json > /dev/null 2>&1 || true
-  echo "Done. Final exit code: $RESULT"
+# ITK harness for a2a-js — a thin shim over a2a-itk's shared driver.
+#
+# Everything that used to live here (clone, image build, container start,
+# readiness poll, POST /run, result reporting, nightly metrics) is now in
+# a2a-itk/scripts/run_itk_shared.sh, which all five SDK repos share. Only the
+# genuinely TS-specific part stays: generating the proto bindings.
```
bartek-gralewicz: This comment is very specific to this change and not to the file that will remain afterwards. I would suggest to remove/adapt. Any change related notes should be documented in the PR description.

itk/run_itk.sh
```diff
@@ -1,211 +1,56 @@
...
-# the ITK proto includes the newer `ResubscribeBehavior` / `hold_task` fields
-# the v0.3 reference baselines (go_v03, python_v03) already consume — required
-# for the resubscribe behavior scenarios below. a2a-itk is also the
-# repository used by the a2a-python / a2a-go ITK harnesses (see
-# `a2a-python/itk/run_itk.sh`), so all SDKs run against the same source of
-# truth.
+# --- bootstrap -------------------------------------------------------------
+# The shared driver lives in a2a-itk, so the checkout has to exist before it
+# can be sourced. CI has already placed it here via actions/checkout; locally
+# we clone it from a2aproject/a2a-itk.
```
bartek-gralewicz: nit: This comment seem very redundant. I'll suggest removing completely.

bartek-gralewicz (approved):
Nice unification.

A ballpark idea: since this change is motivated with unifying the ITK across other A2A SDKs, did you think about doing a reusable github action? Would be a very neat implementation of such functionality and could be called `with:` all relevant options needed.

PR #688, opened 2026-08-27
fix: stamp the resolved taskId onto sendMessage push configs

### LINKED ISSUE #682: [Bug]: sendMessage stores inline push configs without the resolved taskId
### What happened?

`sendMessage` and `sendMessageStream` save `structuredClone(params.configuration.taskPushNotificationConfig)` keyed by the resolved taskId, but never set `config.taskId`. Proto says input taskId should be empty on SendMessage. Later Get/List return `taskId: ""` for a config stored under a real task.

v0.3 `toCorePushNotificationConfig` always emits `taskId: ""`, so message/send with a push config has the same lie.

**Repro**
`sendMessage` a new message (no taskId) with `configuration.taskPushNotificationConfig = { url: "https://example.test/hook" }`. Then list configs for the created task.

Observed: stored `config.taskId` is `""`.
Expected: `config.taskId === task.id`.

### Relevant log output

n/a.

### Code of Conduct

anxkhn:
# Description

- [x] Follow the CONTRIBUTING Guide
- [x] Conventional Commits title
- [x] Tests and linter pass
- [x] Docs updated if necessary

Fixes #682

## Summary

sendMessage and sendMessageStream save a clone of `configuration.taskPushNotificationConfig` keyed by the resolved taskId, but never set `config.taskId`. Proto says input taskId should be empty on SendMessage. Get/List later return `taskId: ""` for a config stored under a real task.

Before save, the clone now gets `taskId` set to the resolved requestContext.taskId on both the blocking and stream paths.

## Testing

Extended the existing inline-config isolation test to assert `stored[0].taskId === task.id`.

```
npx --no-install vitest run test/server/request_handler/push_notification_config.spec.ts
```

5 passed.

### DIFF
--- a/test/server/request_handler/push_notification_config.spec.ts
+++ b/test/server/request_handler/push_notification_config.spec.ts
@@ -227,6 +227,68 @@ describe('DefaultRequestHandler.createTaskPushNotificationConfig (§3.1.7, §5.1
     expect(stored[0].token).toBe('tok');
   });
 
+  it('stamps the resolved taskId on sendMessageStream inline push config', async () => {
+    const executor = new MockAgentExecutor();
+    executor.execute.mockImplementation(fakeTaskExecute);
+    handler = new DefaultRequestHandler(
+      agentCard,
+      taskStore,
+      executor,
+      new DefaultExecutionEventBusManager(),
+      pushNotificationStore
+    );
+
+    const config: TaskPushNotificationConfig = {
+      tenant: '',
+      taskId: '',
+      id: 'c-stream',
+      url: 'https://example.test/hook-stream',
+      token: 'tok-stream',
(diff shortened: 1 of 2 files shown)


test/server/request_handler/push_notification_config.spec.ts
bartek-gralewicz, line 1: Since the change affected both `SendMessage` and `SendMessageStream`, can you please extend the change in tests to cover both as well?

anxkhn: Added a sendMessageStream case next to the sendMessage one. Both now assert the stored config.taskId matches the resolved task id.

bartek-gralewicz (approved):

PR #690, opened 2026-08-27
fix: reject missing required extensions before mutating task history

### LINKED ISSUE #684: [Bug]: sendMessage persists follow-up history before the required-extension check
### What happened?

`_createRequestContext` appends `incomingMessage` to history and `save()`s, then later throws `ExtensionSupportRequiredError` if a required extension is missing. Existing tests only cover new messages (`taskId: ""`), so the follow-up path is untested.

A client that forgets a required extension on a follow-up still permanently appends that message. Later getTask/listTasks/resubscribe show a message the server refused to execute.

**Repro**
1. Create a task.
2. Configure the agent card with a required extension.
3. sendMessage a follow-up with that taskId and no A2A-Extensions.

Observed: the call throws, but `taskStore.load(taskId).history` already contains the follow-up.
Expected: history unchanged.

### Relevant log output

n/a.

### Code of Conduct

anxkhn:
# Description

- [x] Follow the CONTRIBUTING Guide
- [x] Conventional Commits title
- [x] Tests and linter pass
- [x] Docs updated if necessary

Fixes #684

## Summary

`_createRequestContext` appended the incoming message to history and saved, then later threw ExtensionSupportRequiredError if a required extension was missing. Existing tests only covered new messages (`taskId: ""`).

A follow-up that forgot a required extension still permanently landed in the task. getTask, listTasks, and resubscribe then showed a message the server refused to execute.

The required-extension check now runs before any history append or save. New-task creation is unchanged except that a rejected request does not persist.

## Testing

Added a follow-up case that asserts `save` is not called after the extension error.

```
npx --no-install vitest run test/server/default_request_handler.spec.ts
```

95 passed.

### DIFF
--- a/src/server/request_handler/default_request_handler.ts
+++ b/src/server/request_handler/default_request_handler.ts
@@ -165,6 +165,21 @@ export class DefaultRequestHandler implements A2ARequestHandler {
     let task: Task | undefined;
     let referenceTasks: Task[] | undefined;
 
+    const agentCard = await this.getAgentCard();
+    const agentExtensions = agentCard.capabilities?.extensions ?? [];
+
+    // The client MUST declare support for every required extension.
+    const requestedSet = new Set(context.requestedExtensions ?? []);
+    const missingRequired = agentExtensions
+      .filter((ext) => ext.required && !requestedSet.has(ext.uri))
+      .map((ext) => ext.uri);
+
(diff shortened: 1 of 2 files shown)


bartek-gralewicz (approved):

bartek-gralewicz:

PR #692, opened 2026-09-01
feat(server): Introduce SettleByTaskId method

### LINKED ISSUE #620: [Clarification]: Evetbus is closed directly after executor has finished

bartek-gralewicz:
# Description

When the agent executor returns, `_settleBus` decides whether to tear down that task's event bus from the last state it saw published on the bus. That only works when the bus delivers synchronously, as `DefaultExecutionEventBus` does. A bus that batches or persists events before handing them to subscribers has published nothing observable at that moment, so `lastState` is `undefined`, the `keepBusAliveStates` guard is skipped, and the bus is released before its events arrive. The caller then hangs, or fails with `"Agent execution finished without a result"`.

This adds an optional `settleByTaskId` to `ExecutionEventBusManager`. When present, the handler delegates the decision entirely and performs no teardown of its own, letting a deferred-delivery bus settle from its own drain instead. `_settleBus` is the only caller of `finished()` and `cleanupByTaskId()` in `src/`, so the seam covers bus lifetime completely.

`keepBusAliveStates` is unchanged and remains the default policy — it is ignored only while `settleByTaskId` is present. `DefaultExecutionEventBusManager` deliberately does not implement the new method, so nothing changes for existing users.

Fixes #620 🦕

### DIFF
--- a/test/integration/support/delaying.ts
+++ b/test/integration/support/delaying.ts
@@ -1,0 +1,258 @@
+/**
+ * Test doubles that model a deployment which does NOT use the SDK's default
+ * in-process, zero-latency implementations.
+ */
+
+import {
+  AgentExecutionEvent,
+  DefaultExecutionEventBus,
+  EventListener,
+  ExecutionEventBus,
+  ExecutionEventName,
+  FinishedListener,
+} from '../../../src/server/events/execution_event_bus.js';
+import { ExecutionEventBusManager } from '../../../src/server/events/execution_event_bus_manager.js';
+import { ServerCallContext } from '../../../src/server/context.js';
+import { InMemoryTaskStore, TaskStore } from '../../../src/server/store.js';
+import { ListTasksRequest, ListTasksResponse, Task, TaskState } from '../../../src/types/pb/a2a.js';
+import { TERMINAL_STATE_LIST } from '../../../src/server/utils.js';
+
+export const sleep = (ms: number): Promise<void> =>
+  new Promise((resolve) => setTimeout(resolve, ms));
+
+/**
+ * Polls `predicate` until it returns true or `timeoutMs` elapses. Reports what
+ * was being waited for so a failure is legible instead of a bare timeout.
+ */
+export async function waitFor(
+  predicate: () => boolean | Promise<boolean>,
+  description: string,
+  timeoutMs = 5_000,
+  intervalMs = 10
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  for (;;) {
+    if (await predicate()) return;
+    if (Date.now() > deadline) {
+      throw new Error(`Timed out after ${timeoutMs}ms waiting for: ${description}`);
+    }
+    await sleep(intervalMs);
+  }
+}
+
+/**
+ * Resolves to `'pending'` if `promise` has not settled within `ms`. Used to
+ * assert that a call is still outstanding without hanging the test.
+ */
+export async function settlesWithin<T>(promise: Promise<T>, ms: number): Promise<T | 'pending'> {
+  return Promise.race([promise, sleep(ms).then(() => 'pending' as const)]);
+}
+
+/**
+ * An {@link ExecutionEventBus} that defers delivery. `publish()` and
+ * `finished()` append to a queue which is flushed to the delegate one batch at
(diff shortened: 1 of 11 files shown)


test/integration/support/delaying.ts
```diff
@@ -0,0 +1,258 @@
...
+  ): void {
+    // Deliberately no teardown: our reader loop below decides when the task is
+    // really finished. `lastObservedState` is always undefined here because
+    // nothing has been delivered yet, which is exactly why the handler's
+    // state-based policy cannot be used.
+    this.settleRequests.push({ taskId, lastObservedState });
+  }
+
+  protected onBusCreated(taskId: string, bus: DelayingExecutionEventBus): void {
+    bus.on('event', (event: AgentExecutionEvent) => {
```
RonaldMarske: Wouldn't it make sense to have this `bus.on` call in the settleByTaskId if the buss cannot be settled already?

When by the time `settleByTaskId` the task is in a resolved state - the bus can directly be closed.
Most likely in the delayed bus, the last state should be undefined, so you can subscribe on the events.

This could in theory also lead to another race condition.

test/integration/support/delaying.ts
```diff
@@ -0,0 +1,258 @@
...
+  settleByTaskId(
+    taskId: string,
+    _eventBus: ExecutionEventBus,
+    lastObservedState: TaskState | undefined
+  ): void {
+    // Deliberately no teardown: our reader loop below decides when the task is
+    // really finished. `lastObservedState` is always undefined here because
+    // nothing has been delivered yet, which is exactly why the handler's
+    // state-based policy cannot be used.
+    this.settleRequests.push({ taskId, lastObservedState });
```
RonaldMarske: `settleRequests` is technically unused and never get's cleaned.

I would have expected logic which determines if there is a last state and if the bus can be closed already.

Either: `settleRequests` returs a boolean saing "I Take care" or "You take care" - meaning if the task is resolved and the bus can be closed, `return true` and then let the request handler take care of the closing.

If the latest state is undefined or in a not terminated state -> Subscribing to the events and finishing as soon the task is in a terminated state.

The current logic in DefaultRequest handler delegates the closing always to the evenbus manager if the method is present.

Could also be done by a property, then as you show here, the method has not really a useful implementation at the moment. Using `onBusCreated` to take care of closing and a property saying `my responsibility` would be sufficient.  A method does not really bring benefit.

bartek-gralewicz: Applied the boolean return to the `settleByTaskId` method that directs the ownership of the bus settling.

ishymko (approved):

JakubWorek (approved):

PR #698, opened 2026-09-03
fix: validate JSON-RPC response version

### LINKED ISSUE #696: [Bug]: JSON-RPC client accepts responses with missing or invalid jsonrpc version
## What happened

The JavaScript client accepts an HTTP 200 JSON-RPC response even when its `jsonrpc` member is missing or is not `"2.0"`. The same behavior occurs for JSON-RPC envelopes delivered as SSE events.

For example, this successful response is currently decoded as a normal A2A result:

```json
{
  "result": {
    "message": {
      "messageId": "m-1",
      "role": "ROLE_AGENT",
      "parts": []
    }
  },
  "id": 1
}
```

A response with `"jsonrpc": "1.0"` is also accepted.

## What I expected

The client should reject malformed JSON-RPC response envelopes unless `jsonrpc` is exactly `"2.0"`. The validation should apply to both unary responses and each streaming SSE event.

## Steps to reproduce

1. Create a `JsonRpcTransport` with a custom `fetch` implementation.
2. Have the fetch implementation return HTTP 200 and the JSON response above (or change the version to `"1.0"`).
3. Call `sendMessage` with an otherwise valid `SendMessageRequest`.
4. Observe that the promise resolves with the decoded message instead of reporting a malformed JSON-RPC response.
5. Repeat with a `text/event-stream` response whose event data uses the same malformed envelope.

## Additional context

A2A section 6.11.2 requires the JSON-RPC Response `jsonrpc` member to be exactly `"2.0"`; the response must also contain either `result` or `error`. Strict envelope validation would improve cross-SDK interoperability and make malformed upstream or proxy responses visible to callers.

Suggested scope: add shared response-envelope validation for unary and streaming paths, with regression tests for missing and non-`2.0` versions.

russlan23:
# Description

Fixes #696.

## Summary

- reject JSON-RPC response envelopes unless `jsonrpc` is exactly `"2.0"`;
- apply the same validation to ordinary JSON responses and each streaming SSE event;
- cover missing and invalid protocol versions on both response paths.

## Testing

- `npm run lint:ci`
- `npm test -- --run test/client/transports/json_rpc_transport.spec.ts` — 17 passed
- `npm test` — 68 files and 1,529 tests passed

## Checklist

- [x] Use a Conventional Commits pull request title.
- [x] Update the appropriate documentation if necessary.

### DIFF
--- a/src/client/transports/json_rpc_transport.ts
+++ b/src/client/transports/json_rpc_transport.ts
@@ -33,6 +33,22 @@ import { pickMatchingInterface } from './pick_interface.js';
 
 const PROTOCOL_NAME: TransportProtocolName = 'JSONRPC';
 
+function assertJsonRpcResponseVersion(
+  response: unknown,
+  context: string
+): asserts response is Record<string, unknown> & { jsonrpc: '2.0' } {
+  if (
+    typeof response !== 'object' ||
+    response === null ||
+    !('jsonrpc' in response) ||
+    response.jsonrpc !== '2.0'
+  ) {
+    throw new Error(
+      `Invalid JSON-RPC response for ${context}: expected 'jsonrpc' to be exactly '2.0'.`
+    );
+  }
+}
+
 export interface JsonRpcTransportOptions {
   endpoint: string;
   fetchImpl?: typeof fetch;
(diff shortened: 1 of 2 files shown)


src/client/transports/json_rpc_transport.ts
```diff
@@ -33,6 +33,22 @@ import { pickMatchingInterface } from './pick_interface.js';
...
+  context: string
+): asserts response is Record<string, unknown> & { jsonrpc: '2.0' } {
+  if (
+    typeof response !== 'object' ||
+    response === null ||
+    !('jsonrpc' in response) ||
+    response.jsonrpc !== '2.0'
+  ) {
+    throw new Error(
+      `Invalid JSON-RPC response for ${context}: expected 'jsonrpc' to be exactly '2.0'.`
```
bartek-gralewicz, line 47: Please add what arrived in the error message.

```
`Invalid JSON-RPC response for ${context}: expected 'jsonrpc' to be exactly '2.0', got ${received}.`
```

russlan23: Updated in 52c5c85. The error now includes the received version for both unary responses and SSE events. Tests cover missing values, "1.0", numeric 2, and null. All 21 focused tests, the full 1,533-test suite, and lint/type checks pass.

bartek-gralewicz (approved):
Thank you for contributing!

PR #699, opened 2026-09-04
fix: stamp the dispatched wire version onto the A2A-Version header

### LINKED ISSUE #647: [Clarification]: DefaultPushNotificationSender does not set the a2a-version header on outgoing push notifications
Hi, thanks for the great work on the SDK.

When using DefaultPushNotificationSender, outgoing push notifications do not include the a2a-version header. This applies both for v1.0 and when using the legacyCompat serializer with v0.3. The receiver has no way to know which protocol version the body conforms to without inspecting the body itself.

OsamaAnsar:
## What

`DefaultPushNotificationSender` posts every push notification with only
`Content-Type` and auth headers — no `A2A-Version`. A receiver has no way
to tell a v1.0 body apart from a `legacyCompat` v0.3 one without inspecting
the payload shape.

The sender already resolves the exact wire version per dispatch
(`storedConfig.wireVersion`, used to pick the serializer) — this reuses
that same value for the `A2A-Version` header, matching what the client
side already sends on agent-card discovery requests
(`DefaultAgentCardResolver`).

## Testing

- Added a regression test asserting the header value for both a v1.0-tagged
  and a legacy v0.3-tagged dispatch to the same sender instance.
- `npm test` — 68 files / 1526 tests pass.
- `npm run lint:ci` (tsc x2 + eslint) clean.
- `npm run build` clean.

Closes #647

### DIFF
--- a/src/server/push_notification/default_push_notification_sender.ts
+++ b/src/server/push_notification/default_push_notification_sender.ts
@@ -2,6 +2,7 @@ import { TaskPushNotificationConfig, StreamResponse, Task } from '../../index.js
 import {
   A2A_LEGACY_PROTOCOL_VERSION,
   A2A_PROTOCOL_VERSION,
+  A2A_VERSION_HEADER,
   ProtocolVersion,
 } from '../../constants.js';
 import { ServerCallContext } from '../context.js';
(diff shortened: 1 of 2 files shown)


OsamaAnsar:
Closing this — I hadn't seen #659 when I opened this; it fixes the same issue (#647) far more thoroughly (per-config wire version snapshotting, mixed-version tasks on one task, custom-store compatibility). No need to carry two fixes for the same bug.

PR #707, opened 2026-09-08
fix: scope ExecutionEventBusManager by tenant and owner to match TaskStore

### LINKED ISSUE #704: Scope ExecutionEventBusManager by tenant and owner to match task store
### Summary
`DefaultExecutionEventBusManager` keys live event buses by `taskId` alone, while the rest of the multi-tenancy machinery - `InMemoryTaskStore`, `InMemoryPushNotificationStore`, `ResultManager` - scopes by
`(tenant, owner, taskId)`. The task store permits identical task IDs across scopes; the bus manager just uses task ids.

Buses stored in `Map<taskId, bus>`: [`execution_event_bus_manager.ts#L10`](https://github.com/a2aproject/a2a-js/blob/32c148cccadf4b5192306026bf3c2c783b83aed9/src/server/events/execution_event_bus_manager.ts#L10)

Task IDs are server-generated and the SDK hardcodes [`crypto.randomUUID()`](https://github.com/a2aproject/a2a-js/blob/32c148cccadf4b5192306026bf3c2c783b83aed9/src/server/request_handler/default_request_handler.ts#L193) with no override, and an unknown `taskId` on `message/send` is rejected with [`TaskNotFoundError`](https://github.com/a2aproject/a2a-js/blob/32c148cccadf4b5192306026bf3c2c783b83aed9/src/server/request_handler/default_request_handler.ts#L172).

A collision requires a host seeding the store with non-random IDs out of band.

### Credit
Raised via private security report by @Hem1700, assessed non-exploitable given the reasons above, tracked here as future-proof hardening (i.e. when id generation becomes an extension point).

JakubWorek:
# Description

`DefaultExecutionEventBusManager` keyed live buses on taskId alone while `TaskStore` is partitioned by `(tenant, owner, taskId)`. Where a host assigns deterministic task ids, a caller in one scope could attach to another scope's live execution through `resubscribe` or `cancelTask`, observe its events and metadata, and - via cancelTask's drain - persist that data into its own scope.

The handler now passes `ServerCallContext` on every `ExecutionEventBusManager` call, and the default manager partitions buses through `ScopedStore`, the same structure `InMemoryTaskStore` uses.

`context` is optional, existing one-argument callers and custom implementations still works.

A custom manager that accepts `context` and ignores it stays vulnerable. TypeScript does not flag that, so it has to be updated by hand.

Fixes #704  🦕


bartek-gralewicz (approved):

src/server/events/execution_event_bus_manager.ts
```diff
@@ -1,31 +1,49 @@
...
+ * Manages the live {@link ExecutionEventBus} for each active task.
+ */
 export interface ExecutionEventBusManager {
-  createOrGetByTaskId(taskId: string): ExecutionEventBus;
-  getByTaskId(taskId: string): ExecutionEventBus | undefined;
-  cleanupByTaskId(taskId: string): void;
+  createOrGetByTaskId(taskId: string, context: ServerCallContext): ExecutionEventBus;
+  getByTaskId(taskId: string, context: ServerCallContext): ExecutionEventBus | undefined;
+  cleanupByTaskId(taskId: string, context: ServerCallContext): void;
 }
```
ishymko, line 19: Let's schedule the breaking change for the next major release and use `context?` now.

ishymko (changes requested):

src/server/events/execution_event_bus_manager.ts
JakubWorek: Done, context is now optional, callers may not pass it and will still work as before this fix

ishymko (approved):

PR #708, opened 2026-09-08
fix: preserve call context for JSON-RPC tenant requests

### LINKED ISSUE #706: [Bug]: JSON-RPC tenant requests discard custom call context and extension activation
## What happened

When a JSON-RPC request includes `tenant` in its params, the request handler receives a replacement call context instead of the context produced by the configured context builder. Custom state, such as authorization or tracing data, and custom context subclasses are lost. Extensions activated through the replacement context are also not reflected in the `A2A-Extensions` response header.

The same request without a tenant preserves the custom context.

## What I expected

Adding a tenant to a JSON-RPC request should preserve the context builder's state and type, while making the tenant available on that context. Extension activations made during request handling should remain visible to the response layer.

## Steps to reproduce

1. Configure a JSON-RPC server with a custom call context builder that stores a marker such as `authz=allowed` in the context state.
2. Send `ListTasks` with JSON-RPC params containing `tenant: "tenant-a"`.
3. In the request handler, inspect the context state and activate a test extension.
4. Observe that the marker is missing and the response does not include the activated extension header.
5. Repeat without `tenant`; the custom state and extension activation remain visible.

## Additional context

This appears specific to JSON-RPC tenant handling. HTTP+JSON and gRPC resolve the tenant before building the call context. Losing custom context state can break authorization, tracing, and other request-scoped integrations on tenant-scoped JSON-RPC calls.

russlan23:
Fixes #706.

JSON-RPC tenant extraction replaced the supplied `ServerCallContext`, losing custom state and subclasses. Extension activations then landed on a different instance from the one Express reads for its response header.

This sets the tenant on the existing context, following the in-place approach already used for requested extensions. Following review, `setTenant()` now throws if the context already has a tenant, whether set by the constructor or a previous call. JSON-RPC delegates that guard to the setter and returns an error without dispatching when a request supplies a tenant for an already-populated context.

Regression coverage includes context identity, custom state, existing and newly activated extensions, the no-tenant path, the Express `A2A-Extensions` response header, and rejected tenant reassignment. The original two tenant-extraction regressions fail before the initial fix; five guard regressions fail before the review revision.

Validation: 67 focused tests and the full 68-file/1,546-test suite pass, as do `npm run lint:ci`, `npm run build`, `npm run test-build`, and `git diff --check`.

- [x] Follow the CONTRIBUTING guide.
- [x] Use a Conventional Commits title.
- [x] Tests and linter pass.
- [x] Document the new context method in its API comment.


src/server/transports/jsonrpc/jsonrpc_transport_handler.ts
```diff
@@ -96,12 +96,7 @@ export class JsonRpcTransportHandler {
         | string
         | undefined;
       if (paramsTenant && !context.tenant) {
```
bartek-gralewicz: Let's move the `!context.tenant` check to the `setTenant` method. If tenant is not set, proceed with setting, otherwise, a throw should me made not to overwrite an existing in-store tenant value since `setTenant` becomes a public method.

russlan23: Moved the guard into `setTenant()` in 640311b. It now throws before mutation if a tenant was supplied by the constructor or an earlier setter call, including attempts to set the same value again. JSON-RPC delegates the check to the setter, returns the error, and does not dispatch the request. Added direct setter and transport regressions; all 1,546 tests, lint, build, and bundle checks pass.

bartek-gralewicz (approved):
Thank you for contributing! Nice fix

PR #711, opened 2026-09-09
fix(client): cache the extended Agent Card only after signature verification

### LINKED ISSUE #705: [Bug]: Client caches an extended Agent Card before signature verification succeeds
## What happened

When `Client.getAgentCard` fetches an authenticated extended Agent Card with a signature verifier, the client replaces its cached Agent Card before verification completes. If the verifier rejects the card, the call throws, but the unverified card remains active inside the client.

This can change later capability decisions. For example, an unverified extended card can set `pushNotifications: true` and `extendedAgentCard: false`; after the verification error is caught, a later `getAgentCard()` returns that unverified card without fetching or verifying again.

## What I expected

A fetched Agent Card should replace the client's cached card only after the supplied signature verifier succeeds. If verification fails, the previously trusted/public card should remain active, so subsequent calls can retry and capability checks cannot be influenced by rejected content.

## Steps to reproduce

1. Create a `Client` with a public Agent Card that has `extendedAgentCard: true` and `pushNotifications: false`.
2. Configure its transport to return an extended Agent Card named `"UNVERIFIED"` with `extendedAgentCard: false` and `pushNotifications: true`.
3. Call `getAgentCard` with a verifier that throws `Error("invalid signature")`, and catch the error.
4. Call `getAgentCard()` again without a verifier.
5. Observe that the transport was called only once and the second call returns `"UNVERIFIED"` with `pushNotifications: true`, even though that card failed verification.

## Additional context

A2A section 8.4.3 says clients should verify at least one signature before trusting an Agent Card. Section 13.3 recommends replacing the cached public card with the extended version, but the replacement should happen only after successful verification.

Suggested behavior: fetch into a temporary value, run the supplied verifier, then commit it to the client's cached Agent Card. On verification failure, retain the previous card.

Spec: https://github.com/a2aproject/A2A/blob/main/docs/specification.md#843-signature-verification

Ishkirat-Singh:
# Description

`Client.getAgentCard` replaced the cached Agent Card with the fetched extended card *before* running the supplied signature verifier. When verification failed the call threw, but the rejected card stayed active inside the client, so later capability checks (`pushNotifications`, `extendedAgentCard`, `streaming`) were driven by content the caller had refused to trust, and a later `getAgentCard()` served it without fetching or verifying again.

This fetches into a local, runs the verifier, and only then commits the card to the cache. On verification failure the previously trusted card remains active and the next call fetches again, per A2A §8.4.3 / §13.3.

- [x] Ensure the tests and linter pass — `npx vitest run test/client/multitransport-client.spec.ts`: 50 passing; `eslint`/`prettier` clean on the changed files. The new test (`should keep the current agent card when signature verification of the extended card fails`) fails against `main` without the fix.
- [x] Appropriate docs were updated (if necessary) — no doc changes needed.

Fixes #705 🦕


bartek-gralewicz (approved):

PR #715, opened 2026-09-10
ci: implement ACTS SUT behaviors for the ITK agent with YAML config and CI workflows

JakubWorek:
# Description

SUT for ACTS conformance tests, mirrors [python](https://github.com/a2aproject/a2a-python/pull/1230)
Ci workflows, mirrors [python](https://github.com/a2aproject/a2a-python/pull/1243)

Based on spec: https://github.com/a2aproject/a2a/pull/1882


itk/acts_behaviors.ts
```diff
@@ -0,0 +1,392 @@
...
+      return;
+    }
+
+    default: {
+      const terminal: Record<string, TaskState> = {
+        'tck-complete-task': TaskState.TASK_STATE_COMPLETED,
+        'tck-task-failure': TaskState.TASK_STATE_FAILED,
+        'tck-reject-task': TaskState.TASK_STATE_REJECTED,
+        'tck-input-required': TaskState.TASK_STATE_INPUT_REQUIRED,
+        'tck-auth-required': TaskState.TASK_STATE_AUTH_REQUIRED,
```
bartek-gralewicz, line 377: nit: `..._INPUT_REQUIRED` and `..._AUTH_REQUIRED` are not terminal states, they are interrupted states. Perhaps consider renaming the dict to something more generic to avoid confusion in future refactors?

bartek-gralewicz (approved):

ishymko (commented):
Let's not use `feat` for this as it's not shipped for the end users so having such entries in changelog can be confusing. `ci` or `test` maybe?

itk/acts_behaviors.ts
JakubWorek: renamed

PR #717, opened 2026-09-11
feat(server/database): add the push notification store

JakubRybak:
Adds the push notification store, built on Kysely: DatabasePushNotificationStore and the ./server/database entry point. 
Includes tests on SQLite, PostgreSQL and MySQL.
For now no tests on edge environment.


src/server/database/push_notification/migrations.ts
```diff
@@ -3,7 +3,7 @@ import type { StoreMigrations } from '../migrator.js';
 import * as createPushNotificationConfigs from './migrations/0001_create_push_notification_configs.js';
 
 export const PUSH_NOTIFICATION_STORE_MIGRATIONS: StoreMigrations = {
-  id: 'push-notification',
+  id: 'push-notification-configs',
```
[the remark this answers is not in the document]
JakubRybak: The `a2a-db` CLI hasn't shipped, so there is no need to worry about compatibility.

src/server/database/push_notification/store.ts
```diff
@@ -0,0 +1,177 @@
...
+   */
+  async initialize(): Promise<void> {
+    // Only success is cached, so migrating a database this store already
+    // rejected does not need a restart.
+    this.schemaCheck ??= this.checkSchema().catch((error: unknown) => {
+      this.schemaCheck = undefined;
+      throw error;
+    });
+    await this.schemaCheck;
+  }
```
ishymko, line 61: IMO it's overcomplication, it's safe to assume that the database is provisioned, we shouldn't mix serving path and configuration. Also `checkSchema` does not validate against column types for instance, so it won't cover all possible discrepancies. I'd remove this for now.

ishymko (approved):

src/server/database/push_notification/store.ts
JakubRybak: Done in d926abf.

PR #721, opened 2026-09-14
fix: preserve HeadersInit in authenticating fetch

### LINKED ISSUE #716: [Bug]: Authentication fetch wrapper drops Headers entries and corrupts tuple-array headers
## What happened

`createAuthenticatingFetchWithRetry` preserves request headers supplied as a plain object, but loses entries supplied through a `Headers` instance and corrupts a tuple-array `HeadersInit`. This affects both the initial request and the authentication retry.

With `A2A-Version: 1.0` and `Content-Type: application/json`:

- Plain object: both headers are preserved.
- `new Headers(...)`: both headers disappear; only the authentication handler's Authorization header remains.
- `[['A2A-Version', '1.0'], ['Content-Type', 'application/json']]`: the request instead contains headers named `0` and `1`, with comma-joined pair values.

## What I expected

The fetch wrapper should preserve caller headers for all supported `HeadersInit` representations, on both the initial request and retry.

## Steps to reproduce

1. Wrap a mock fetch using `createAuthenticatingFetchWithRetry`, with `headers()` returning an Authorization header and `shouldRetryWithHeaders()` returning a refreshed Authorization header.
2. Have the mock capture `new Request(input, init).headers`, return 401 on its first invocation and 200 on its second.
3. Call the wrapped fetch with `headers: new Headers({ 'A2A-Version': '1.0', 'Content-Type': 'application/json' })`.
4. Observe that both captured requests lack the supplied version and content-type headers.
5. Repeat using a plain object and then a tuple array; observe the comparison above.

## Additional context

This can affect version negotiation, content-type handling, and custom extension headers when composing the authentication helper with other fetch wrappers. The helper is typed as `typeof fetch` and accepts `RequestInit`, whose headers support all these representations.

Normalizing the caller's headers through `Headers` before merging would support those representations while allowing the intended precedence to be preserved.

Fetch standard: https://fetch.spec.whatwg.org/#headers-class

sunruize93-cmyk:
# Description

Fixes #716.

Normalize authentication and caller headers through the native `Headers` API instead of spreading `RequestInit.headers` as a plain object. This preserves plain-object, `Headers`, and tuple-array inputs on both the initial request and the authentication retry.

Caller headers retain precedence over authentication-handler headers, with case-insensitive names. A fresh header collection is created for each request; neither the caller's headers nor the handler's header objects are mutated. The public API and retry flow are unchanged.

## Tests

- Added nine regression cases covering all three header representations, 401/403 challenges, refreshed authentication headers, POST body/method preservation, case-insensitive caller precedence, and input preservation.
- Updated existing authentication tests to inspect header values through `Headers`, rather than assuming a plain-object representation.
- Against upstream `ce10234b3a68e34c5b605eb95aaaf1474b9397fe`, the new tests produce **7 failures and 2 passes**. With this fix, all nine pass.
- `npm test`: **1,555 tests pass across 69 files**.
- `npm run lint:ci`: passes.
- `npm run build`: passes (ESM, CJS, and declarations).
- Additional local native-fetch/HTTP-server checks pass for raw `Headers` and tuple-array inputs, each with a 401 followed by 200, retaining caller headers, POST body, and refreshed authentication.

Validation used Node.js 22.22.3. The separately configured integration and edge suites were not run.

## Checklist

- [x] Follow the CONTRIBUTING guide.
- [x] Use a Conventional Commits PR title.
- [x] Ensure the default tests and linter pass.
- [x] Update relevant documentation (helper JSDoc documents header precedence).

AI assistance: this patch and its tests were developed and checked with Codex. The test results above are local executions, not maintainer approval or independent human review.


bartek-gralewicz (approved):

PR #723, opened 2026-09-14
feat(compat/v0_3): expose convertLegacyPushBody boundary function

BennetKunze:
## Summary

Adds a single `convertLegacyPushBody(body: unknown)` function exported from `@a2a-js/sdk/compat/v0_3`.

This is the narrow boundary-function shape discussed and approved in #670 - a stable input/output contract that converts a raw v0.3 push notification body to its v1.0 equivalent, while leaving the internal translators free to change without it being a breaking change.

**Use case:** receiving a v0.3 push notification on your own webhook route when you do not control the sending agent and cannot enforce v1.0 on the sender side.

```ts
import { convertLegacyPushBody } from '@a2a-js/sdk/compat/v0_3';

app.post('/callback', (req, res) => {
  const v1 = convertLegacyPushBody(req.body);
  // v1 is V1Task | V1Message | V1TaskStatusUpdateEvent | V1TaskArtifactUpdateEvent
});
```

## Changes

- `src/compat/v0_3/convert_legacy_push_body.ts` - implementation: dispatches on the `kind` field to the existing internal `toCore*` translators, throws `A2AError.invalidParams` for unrecognised payloads
- `src/compat/v0_3/index.ts` - one line re-exporting the function from the existing public subpath
- `test/compat/v0_3/convert_legacy_push_body.spec.ts` - 8 tests covering all four `kind` values plus error cases

## Related

Closes #670


BennetKunze:
Closing in favour of #726 which covers the same use case with a cleaner return type and more thorough validation.

PR #725, opened 2026-09-15
fix: avoid mutating caller message configuration

### LINKED ISSUE #724: [Bug]: Client send methods mutate reusable request configuration
## What happened

`Client.sendMessage()` and `Client.sendMessageStream()` mutate caller-owned `configuration` while applying defaults. Reusing a request with a second client can retain the first client's push notification URL. A frozen configuration throws `TypeError: Cannot assign to read only property 'acceptedOutputModes'` before reaching the transport, even when the configuration is complete.

## What I expected

Applying defaults should leave the caller's configuration unchanged. Each call should use its own client's defaults for missing values, while preserving explicit request values.

## Steps to reproduce

1. Install `@a2a-js/sdk@1.1.0` and run the following as an `.mjs` file with Node 22.
2. The two recorded callback URLs are both `https://first.example/callback`; the second should be `https://second.example/callback`.
3. The original request acquires the first callback configuration, and the frozen configuration throws.

<details>
<summary>Self-contained reproduction (no network calls)</summary>

```js
import { Client } from '@a2a-js/sdk/client';

const sent = [];
const transport = {
  protocolName: 'repro',
  protocolVersion: '1.0',
  async sendMessage(params) {
    sent.push(structuredClone(params.configuration));
    return params.message;
  },
};
const card = { capabilities: { streaming: false } };
const params = {
  tenant: '',
  message: {
    messageId: 'request', role: 1, parts: [], contextId: '', taskId: '',
    extensions: [], metadata: {}, referenceTaskIds: [],
  },
  configuration: {
    acceptedOutputModes: ['text/plain'], returnImmediately: false,
    historyLength: 0, taskPushNotificationConfig: undefined,
  },
  metadata: {},
};
const push = {
  tenant: '', taskId: '', id: 'first', token: '', authentication: undefined,
  url: 'https://first.example/callback',
};
await new Client(transport, card, { pushNotificationConfig: push }).sendMessage(params);
await new Client(transport, card, {
  pushNotificationConfig: { ...push, id: 'second', url: 'https://second.example/callback' },
}).sendMessage(params);
console.log(sent.map(c => c.taskPushNotificationConfig.url));
console.log('Caller configuration:', params.configuration);

try {
  await new Client(transport, card).sendMessage({
    ...params,
    configuration: Object.freeze({
      acceptedOutputModes: ['text/plain'], returnImmediately: false,
      taskPushNotificationConfig: undefined,
    }),
  });
  console.log('Frozen configuration accepted');
} catch (error) {
  console.log(error.name, error.message);
}
```

</details>

## Additional context

Reproduced on `@a2a-js/sdk@1.1.0` and main `ce10234b3a68e34c5b605eb95aaaf1474b9397fe`, Node 22.22.3. The reproduction uses a JavaScript transport stub; the configuration includes all required fields. Regression tests also cover consuming `sendMessageStream()` on a streaming-capable agent.

A focused fix and regression tests are prepared. AI-assisted investigation and implementation.

sunruize93-cmyk:
## Description

Fixes #724.

Copy the caller's message configuration before applying client defaults. This prevents frozen configurations from throwing and stops one client's default push callback from leaking into a request reused with another client. Explicit request values keep their existing precedence.

Regression tests cover frozen configurations and reuse across clients through both `sendMessage()` and `sendMessageStream()`. All four new cases fail on the baseline and pass with the fix.

## Validation

- `npm test`: 1,550 tests passed across 68 files.
- `npm run lint:ci`: passed (source/test TypeScript and ESLint).
- `npm run build` and `npm run test-build`: passed.
- No public API or wire-format changes; no documentation changes needed.

## Checklist

- [x] Follow the contributing guide
- [x] Conventional Commit PR title
- [x] Tests and linter pass
- [x] Appropriate docs updated if necessary (not needed)

AI-assisted implementation and regression testing.


sunruize93-cmyk:
The [Node 22 TCK failure](https://github.com/a2aproject/a2a-js/actions/runs/34927833829/job/104249554951) is a false positive in the pinned `0.3.0.beta5` test, rather than a configuration-copy regression.

`test_tasks_resubscribe_nonexistent` receives HTTP **404**, code **-32001**, with `Task not found: non-existent-task-id-61501690-1086-477f-a55d-5f70f0d5e179`. The generated UUID contains `501`. The [exception handler](https://github.com/a2aproject/a2a-tck/blob/0.3.0.beta5/tests/optional/capabilities/test_streaming_methods.py#L517) checks `"501" in error_msg` before checking `404` / `not found`, so it misclassifies the task ID as an HTTP 501 status.

Local verification:
- Replayed the actual pinned test function against the captured REST error: reproduced the same failure.
- Changed only `61501690` to `61401690`: the identical HTTP 404 error passes. An HTTP 501 control still fails.
- Sent real requests to this branch's compat SUT with both IDs: both correctly return HTTP 404 / -32001. This server-side resubscribe path does not use the modified client configuration helper.

The Node 24 job's logs also show all three capability transport suites passing before matrix cancellation. No SDK changes are needed for this particular failure.

AI-assisted diagnosis and local validation; the full TCK suite was not rerun locally.

I attempted to rerun the failed jobs, but GitHub rejected the request with `Must have admin rights to Repository`. Could a maintainer rerun the failed/cancelled TCK jobs, please?

bartek-gralewicz (approved):

PR #726, opened 2026-09-15
feat(compat): convert a v0.3 push notification body to a v1.0 StreamResponse

### LINKED ISSUE #646: Expose v0.3 to v1.0 conversion helpers as public API

harshitwandhare:
# Description

#646 asked for the v0.3 to v1.0 translators to be exported. That was declined in the thread, because publishing them makes the translation logic a semver contract and it is expected to keep changing as compat bugs come in. @bartek-gralewicz approved a narrower shape instead, and this is it: one boundary function, with `translate/` still private.

The case it serves, from @BennetKunze on the issue: they receive a v0.3 push notification on their own Express callback route and do not control the sending agent, so they cannot register the webhook as v1.0 and let the SDK translate on the way out.

```ts
import { legacyPushNotificationToV1StreamResponse } from '@a2a-js/sdk/compat/v0_3';

app.post('/webhook', express.json(), (req, res) => {
  const event = legacyPushNotificationToV1StreamResponse(req.body);
  switch (event.payload?.$case) {
    case 'task': ...
    case 'statusUpdate': ...
  }
  res.sendStatus(200);
});
```

## Why `StreamResponse` is the return type

`V03PushNotificationSerializer` already pins down what a v0.3 webhook body can be: the bare v0.3 `Task`, `Message`, `TaskStatusUpdateEvent` or `TaskArtifactUpdateEvent`, with no `StreamResponse` wrapper and no JSON-RPC envelope. Those four are exactly the arms of the v1.0 `StreamResponse` payload union, so this function is the inverse of that serializer and its return type is closed by construction. A receiver switches on `payload.$case` the same way it already does for a streaming response.

It goes on the `compat/v0_3` root entry, which holds only constants and pure functions, so it adds no peer dependency and stays Workers-safe. Verified below rather than assumed.

## Validation, and the calls I made

The body arrives from an agent the receiver does not control, so required fields are checked at the boundary instead of being passed through into a result that would not match its own declared type.

Checked per arm, all of them non-optional in `src/compat/v0_3/types/types.ts`: `task.id`, `task.contextId`, `task.status`; `status-update.taskId`, `.contextId`, `.status`; `artifact-update.taskId`, `.contextId`, `.artifact`. The `message` arm delegates, since `toCoreMessage` already validates `messageId`, `role` and `parts`. Anything else throws `A2AError.invalidParams` naming the field.

Two things I would rather you decide than have me guess, and one that is now settled:

- **`final` is not checked**, even though the v0.3 schema marks it required on `status-update`. `toCoreTaskStatusUpdateEvent` discards it, so rejecting a body over a field we throw away seemed wrong. @BennetKunze pointed out on the issue that this is not just an SDK choice: `docs/migration-guide.md` lists `TaskStatusUpdateEvent.final` under the changes defined by the spec, removed in v1.0. Checked, and it does. So I have left this as-is rather than as an open question.
- **The local `requireObject` is stricter than the one in `translate/_validate.ts`**, which accepts an array as an object. I did not change the shared one because it is used on paths I am not touching in this PR. Happy to unify either way in a follow-up.
- **The name.** It matches the `legacyJsonRpcToV1Method` style already on that entry, but it is a mouthful. Easy to change.

Unknown task states are left alone: `toCoreTaskState` maps them to `TASK_STATE_UNSPECIFIED`, and there is a test pinning that this does not throw.

This was tested against the SDK's own serializer output and against hand-written v0.3 bodies. It has not been run against a live third-party v0.3 agent.

## Verification

Four of the twenty tests feed a v1.0 event through `V03PushNotificationSerializer` and back through this function, one per payload case, and assert the result equals the original.

- **New tests:** 20, all passing.
- **Full suite:** rebased onto `main` at `ad59828` and re-measured, since the branch was 14 commits behind. `main` gives 72 files / 1589 passing, this branch gives 73 / 1609. Compared node-id by node-id rather than by count: exactly 20 added, 0 removed, 0 changed status. The two earlier baselines quoted here (1546, then 1571) were correct when they were taken; every gate below was re-run at the rebased head.
- **Non-vacuity:** the tests cannot fail on `main` because the module does not exist there, so two sabotages were run instead. Dropping the `task.contextId` guard fails exactly 1 test; mislabeling the `status-update` arm as `$case: 'task'` fails exactly 1. Reverted, 20/20 again.
- **`npm run lint:ci`:** exit 0 (`tsc --noEmit`, `tsc --noEmit -p tsconfig.test.json`, `eslint .`).
- **`npm run build`:** exit 0.
- **`build-tests.yml`, all three stages in CI's order:** `test-build:workers-safe` with `@grpc/grpc-js`, `@bufbuild/protobuf` and `express` uninstalled, then `test-build:grpc`, then `test-build:express`. All pass. In the emitted `dist/tmp-checks/compat/v0_3/index.js` the new symbol is present and there are 0 references to any of the three peers.

Fixes #646


src/compat/v0_3/push_notification.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (typeof value !== 'string' || value === '') {
+    throw A2AError.invalidParams(`${path} is required and must be a non-empty string`);
+  }
+}
+
+function requireObject(value: unknown, path: string): void {
+  if (value === null || typeof value !== 'object' || Array.isArray(value)) {
+    throw A2AError.invalidParams(`${path} is required and must be an object`);
+  }
+}
```
bartek-gralewicz, line 36: Hey, please move these methods to `src/compat/v0_3/translate/_validate.ts`. `requireObject` is already defined there and I believe that it could be reused instead of defining a second, very similar method.

harshitwandhare: Done in 11fc9ce. `requireString` now lives in `_validate.ts` next to the other two, with the same `(value, path)` shape and tests in `_validate.spec.ts`, and the receiver uses the existing `requireObject`.

One behaviour change comes with it, since the shared `requireObject` accepts arrays. A task or status-update with `status: []` now converts to `TASK_STATE_UNSPECIFIED` instead of throwing. `status: {}` already did that before this change, so I don't think the array check was protecting anything. An array as the whole body is still rejected, by the unknown-`kind` branch. The error text for a non-object body is now the shared "... is required".

I also merged `main` again, since #748 landed after your update. Full suite is 74 files / 1664 passing against 73 / 1636 on `main` (28 added, nothing removed or changed), and `lint:ci`, `build` and the three build-test stages pass.

bartek-gralewicz (approved):

PR #728, opened 2026-09-15
feat(server/database): add the task store

JakubRybak:
Adds `DatabaseTaskStore`, the Kysely-backed `TaskStore`, against the `tasks` table.
## Summary
- `save()` upserts, `load()` reads one task, `list()` filters by context, status and status timestamp;
- listing pages with a keyset cursor over `(status_last_updated, id)`;
- an unreadable row is skipped and logged rather than failing the page, and the cursor still advances past it;
- no `initialize()` or `checkSchema()`, matching the push notification store: a missing table surfaces as the driver's own error on first use;
- covered on SQLite, PostgreSQL and MySQL.


src/server/database/task/store.ts
```diff
@@ -0,0 +1,189 @@
...
+      .selectFrom(TASK_TABLE)
+      .select((eb) => eb.fn.countAll().as('total'))
+      .where(filter)
+      .executeTakeFirstOrThrow();
+    // The engines disagree on what type a count comes back as.
+    const totalSize = Number(counted.total);
+
+    let query = this.db
+      .selectFrom(TASK_TABLE)
+      .select([...(includeArtifacts ? TASK_TABLE_COLUMNS : TASK_TABLE_COLUMNS_WITHOUT_ARTIFACTS)])
```
ishymko: nit: We can do the same for history to skip `history` column for an explicit zero: https://a2a-protocol.org/v1.0.0/specification/#324-history-length-semantics.

Maybe it can be one filter statement where we can write this logic.

src/server/database/task/store.ts
```diff
@@ -0,0 +1,189 @@
...
+    await (
+      this.dialect === 'mysql'
+        ? insert.onDuplicateKeyUpdate(replaceable)
+        : insert.onConflict((clause) =>
+            clause.columns([...TASK_TABLE_KEY_COLUMNS]).doUpdateSet(replaceable)
+          )
+    ).execute();
+  }
+
+  async load(taskId: string, context: ServerCallContext): Promise<Task | undefined> {
```
ishymko, line 95: Hm, now I realize we do not propagate history length and include artifacts. We should think about propagating this info to allow task store implementation to be more efficient and not to query something which will be discarded.

Not in scope of this PR as it's an interface change (can be done non-breaking but still it's a scope change), but worth adding an issue to explore it.

ishymko (approved):

src/server/database/task/store.ts
JakubRybak: Done in 293f8b9

PR #730, opened 2026-09-16
feat(server/database): support node 20

JakubRybak:
## Summary
- `engines.node` returns to `>=20`, and the kysely peer range widens to `^0.28.5 || ^0.29.0`.
- kysely 0.28 exports the `Migrator` from the package root and 0.29 moved it to `kysely/migration`, leaving a compile-time error at the root — no single specifier resolves on both. `migrator.ts` now loads the class and the `NO_MIGRATIONS` sentinel at runtime from whichever path answers, and takes their types from `kysely/migration`, which is erased at compile time;
- `migrateStoreTo` translates the CLI's `base` revision into that sentinel itself, so nothing about kysely's migration API is re-declared here;
- cover `downgrade base` in both migrations specs — the only path that uses the sentinel, and previously untested;
- drop the `better-sqlite3` devDependency to `^12.4.1`.


.github/workflows/unit-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        node-version: [22, 24]
+        node-version: ['20.0.0', 22, 24]
```
ishymko:
```suggestion
        node-version: [20, 22, 24]
```

src/cli/migrator.ts
ishymko, line 1: Is this file used only in the CLI for migrations? Maybe we can move it outside of the `server` and keep under `cli`?

.github/workflows/unit-tests.yml
JakubRybak: Done in 4baf0e9

src/cli/migrator.ts
JakubRybak: Mostly — the engine is now `src/cli/migrator.ts`.

`StoreMigrations` stayed under `src/server`: both stores declare against it, so moving it would make `src/server` import from `src/cli`.

 The two migration specs moved to `test/cli/ too`.

JakubRybak:
A few changes beyond your comments, so they don't appear from nowhere:

Peer floor raised to `^0.28.17` (was `^0.28.5`). `0.28.17` is the last release of that line - so it's what any `^0.28` install actually resolves to.

The matrix now pairs a kysely version with each Node version — 20 with `0.28.17`, 22 and 24 on the lockfile (`0.29.5`). (Kysely `0.28.17` is the latest that supports Node 20).

New spec `test/cli/migrator_loader.spec.ts` covering the loader that bridges kysely's two layouts

ishymko (approved):

